Issue #4227, opened 2014-03-04
Building with clang


LordAro:
Building with clang (3.4) on Arch Linux is broken, primarily with errors relating to __float128

The following patch fixes this: http://paste.openttdcoop.org/show/3156/

Would you like me to make a pull request out of it?

Note that the follow warnings are still produced http://paste.openttdcoop.org/show/3157/

PR #4228, opened 2014-03-04
Fix compilation with clang

LordAro:
At least, clang 3.4, on Arch Linux

Otherwise it errors with __float128 errors, and there's also a "register keyword is deprecated" error fix in there too

"Fixes" #4227


jorgicio:
My 20 cents:
I found a workaround on LLVM bugtrack, then created a patch and applied into the CMakelists.txt in the src root. Using Clang 3.5.0 and Clementine 1.3.1 and working.

https://bugs.gentoo.org/attachment.cgi?id=433122&action=diff

Issue #853, opened 2013-12-06
Improve large album cover preview


Clementine-Issue-Importer:
_From [mail@icesheep.de](https://code.google.com/u/108326814684193951398/) on October 04, 2010 20:31:47_

I think there are two things that could be improved about the large album cover preview (which appears while playing a song).

1.) Remove the gradient from the bottom and display the text below the cover
2.) Add the ability to horizontally scale the cover preview

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=853_

PR #4334, opened 2014-05-07
Removed gradient & text from large album cover (Issue #853)

FletcherD:
Hello, this is my attempt to address issue #853 and make the large album art widget better. This simply removes the black gradient and text over the large album cover to give a cleaner interface and show off the album art better. If the user wants the song information next to the album art, they can still switch to the small album art. Thank you!
tv6yjgl


FletcherD:
I created a better patch which is optional instead: https://github.com/clementine-player/Clementine/pull/4727

Issue #2685, opened 2013-12-11
Make your player 20&#37; cooler with Rainbow Dash


Clementine-Issue-Importer:
_From [faust....@gmail.com](https://code.google.com/u/114894426186294597406/) on February 01, 2012 17:06:09_

You guys do really great player and not forget about funny things.

Could you add rainbow dash analyzer? like nyancat but with rainbow dash?

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=2685_

paperbagcorner:
This feature has been implemented in revision 663b7430d8f9b2dca088a150ecaf112cd5ff9543 (pull request #4367) so I believe this issue can be closed.

Also, I think the priority of this issue should be higher.

g4borg:
even this is the bestest feature ever, I find it sad, the colors are not exactly upside down to the nyan cat one, to meet the coloring of the trail to Dashs' tail.

PR #4367, opened 2014-05-22
Add new analyzer "Rainbow Dash"

TheUbuntuGuy:
As requested in SuperUltraCriticalShowstopper request #2685
I have built an analyzer based on Rainbow Dash
...oh god, what have I done...

This will either make some people happy, or is absolute garbage.
Totally your call.

screenshot from 2014-05-21 210616
screenshot from 2014-05-21 210631


Thrackerzod:
I used Clementine late last year on my previous Ubuntu install and the RD analyzer was there, but it doesn't seem to be anymore. Was it removed? D:

PR #4385, opened 2014-05-30
Live scanning

vranki:
The scanner now updates UI during the scan with changes for each directory. The scan can take really long time for a large/slow collection, and this change makes Clementine usable during the scan.

This is my first contribution to Clementine so someone who really understands how the scanner works should check if there are any errors in the code.

### DIFF
--- a/src/ui/mainwindow.ui
+++ b/src/ui/mainwindow.ui
@@ -477,11 +477,7 @@
      <x>0</x>
      <y>0</y>
      <width>1131</width>
-<<<<<<< HEAD
      <height>23</height>
-=======
-     <height>20</height>
->>>>>>> Support for aborting library scan gracefully.
     </rect>
    </property>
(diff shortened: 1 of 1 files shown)


alphapapa:
> There are so many dependencies that imho for developer a tool for installing all deps would be useful. No one has to use it.

The `apt-get build-dep` command does that, but of course only for packages in the distro archive.

A tool for that would definitely be useful, it should just be in a separate script.  Building a package shouldn't automatically install anything on the user's system.  The build script could recommend the dependency-installing script if it detects missing dependencies.

vranki:
I splitted the create_debian_package.sh to other script which installs the dependencies now.
I also rebased all the changes against up to date master branch.

alphapapa:
@vranki Thanks, I hope it gets merged!  :)

Issue #4419, opened 2014-06-25
Organizing Files after changing artist tag


StephanRichter:
Steps:
1. Load track to playlist
2. Change artist tag of tacks
3. Right-click tracks, choose "organize files" ("Dateien organisieren" on my german system)
4. use rename string containing %artistinitial

What is expected:
%artistinital is replaced with the uppercase version of the first letter of the NEW artist tag

What occurs:
%artistinitial is replaced with the uppercase version of the first letter of the PREVIOUS artist tag

examle:
loaded Beethoven's "Allegro Ma Non Troppo". Artist column says "Beethoven, Ludwig van". I manually altered this field to "Ludwig van Beethoven", which better suits my preferences.
Then, trying to organize the file with the replacement pattern "%artistinitial/{%artist/}{[%year] }%album{/CD %disc}/{%track - }%title.%extension", the preview says "/home/srichter/Musik/B/Ludwig van Beethoven/..."

Clementine 1.1.1 running on OpenSuse 13.1x64

Issue #221, opened 2013-12-06
sometimes clementine does not update tags (when they were edited with external tagging software for example)


Clementine-Issue-Importer:
_From [trickykid1982](https://code.google.com/u/trickykid1982@gmail.com/) on April 16, 2010 01:07:24_

What steps will reproduce the problem? 1.Editing metadata in .flac files, either using Clementine or external tag
editor/player. What is the expected output? What do you see instead? Tags do not update in library despite being written to file. What version of the product are you using? On what operating system? OS X 0.3 beta 2 Please provide any additional information below.

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=221_

jerobarraco:
i'm still having the issue, i'm trying to solve the problem but when i debug it it doesn't seem to show the behavior.

kevincox:
I'm using 1.2.3 on Arch Linux and tag editing now appears to be working most of the time.  There are some rare cases where it doesn't apply but I haven't been able to track down when that it (I suspect it has something to do with filtering a playlist then editing a track but am not sure).

tommycarstensen:
Same issue with 1.2.2 on Mac.

zoltan-sule:
I was happy when I found the Clementine, because I think this is a good alternative and what is more this is the best music player for kind of linux users like me.
But this caching (I just suppose it) problem gives me big troubles. This way I can't manage my albums. I have to spend a lot of time to remove the musics from the list, rename them and add them to the same list one after the other (remembering all of them) all the time if I change the BPM, the comment with bpm-tool or with puddletag.
I imagine a simple button: switch of the pre-caching. Or a button that clean the caches, but don't remove the lists and the ratings. I don't mind if the loading of it isn't so fast, because the process above takes hours.

I am in Ubuntu 14.04 + Clementine 1.2

Hepo:
I'm running Linux Mint 17.1 and Clementine 1.2.3 from PPA. Still encountering this bug; updating the tags, rescanning the library, removing/re-adding the library does not fix the issue. The artist is constantly put underneath its Genre as "Various Artists".

rodrigodiez:
Is there any special feature around the 'various' keywork in a tag? As far as I remember, most of the times that I have experimented issues with tagging they were while trying to update a 'various' ish tag

spock:
This issue looks like what I am looking for...

My setup:
- around 70 GB of music, 20.5k files (including .mood);
[...]

PR #4431, opened 2014-07-03
Update database after tags are edited

Narfinger:
For some reason nobody updated the database when tags where edited.
Problems:
- When editing many songs the Tagedit dialog stays open longer (which I think is ok).
- If there are calls to TagReaderClient::SaveFile (non blocking) which expect to return really fast we might not return as fast. There are some places where this is called and I could only test Inline Editing which is fine with this change.
- I didn't test CDRip.

This might fix: #4419, #221 but I only looked at the description if this might be what the reports mean.
P.S.: Initially I tried to use NewClosure to call the new function but I couldn't make it work. If somebody wants to take a stab at it, be my guest.

### DIFF
--- a/src/core/application.cpp
+++ b/src/core/application.cpp
@@ -75,12 +75,12 @@ Application::Application(QObject* parent)
       network_remote_(nullptr),
       network_remote_helper_(nullptr),
       scrobbler_(nullptr) {
-  tag_reader_client_ = new TagReaderClient(this);
-  MoveToNewThread(tag_reader_client_);
-  tag_reader_client_->Start();
-
   database_ = new Database(this, this);
   MoveToNewThread(database_);
+
+  tag_reader_client_ = new TagReaderClient(this, this);
+  MoveToNewThread(tag_reader_client_);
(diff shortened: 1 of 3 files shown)


thenewnano:
There is need for this function as I see issue #5946 having some help from this patch or the simpler approach of touching the folder (see my patch) and kicking a incremental scan (needed to be done by someone better than me at cpp ) . @Narfinger  could you please take a look at this again, I spend at least ten hours to make myself a work around, still I am thinking about moving away from Clementine, even if my own-compiled version with my patch behaves acceptable on OSX it is not optimal and I don't see any chances of it being merged as it only goes half the way.

Narfinger:
I think forcing a change date on the folder is really a workaround and not a good solution.

As far as I understood my problem, it had something to do with the custom implementation of signals for closures which are standard in qt5, so as soon as clementine is ported to qt5, a modification of my patch should work. I, however, moved (for various reason but this being one of them) away from clementine, so I doubt I will have a look at this ever again.

Issue #4415, opened 2014-06-22
Add "Remove dead items" option for playlist


choffm:
It would be quite useful to have an option which removes all (local) tracks from the playlist which do not exist.

Issue #3397, opened 2013-12-12
Remove dead items


Clementine-Issue-Importer:
_From [a...@devfine.net](https://code.google.com/u/116660575812332823427/) on January 01, 2013 19:52:41_

Would be useful to have a feature to remove dead items (items which were moved) in addition to scan for duplicates

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=3396_

PR #4499, opened 2014-08-24
Remove unavailable items from playlist

ppkt:
Hi,
This fix should resolve problems #3397 and #4415.

I'm only not sure if `song.url().isLocalFile()` is enough to determine if file is on local machine - I've seen additional checks for song type: `!song.is_stream() && !song.is_cdda()`

### DIFF
--- a/src/playlist/playlist.cpp
+++ b/src/playlist/playlist.cpp
@@ -2109,6 +2109,21 @@ void Playlist::RemoveDuplicateSongs() {
   removeRows(rows_to_remove);
 }
 
+void Playlist::RemoveUnavailableSongs() {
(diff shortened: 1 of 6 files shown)


Komzar:
This will be very helpful especially for DB. Can we have this feature to remove from DB all entries pointed to not existed files ?
 This will reduce a DB a lot. For example I have plenty of entries in DB for the same song (check in SQLite), different folder names (probably old, not existing) and in the Clementine I can see only one, the existing file. 
Because of this my DB is 3 times bigger then needed.
I understand that this feature should be not automatic, but run when user like to do it for presently connected drives/folders.
 This is kind of DB clean up.

PR #4551, opened 2014-09-29
Use residents notifications for better gnome shell/plasma compatibiliy

ghost:
Here a patch adding residents notifications support for clementine.

This make the desktop shell not open a new notification when a previous one is open.


Kwaadpepper:
Just wanted to say, clementine is fully integrated with last plasma version for notification and player tray (the plasma one)

Yanpas:
Can't clone it and test it, so I wonder how will it work under XFCE. Will the behavior remain the same?

Personally I wanted to add button "Next" to notification (buttons are supported across all DEs, this button exists in Banshee player), but now I 'm not sure whether I have to do it

Ferroin:
I can probably verify the behavior in XFCE, but I won't have the time to do so until next week.

PR #4596, opened 2014-10-27
Sorting by disc in advanced grouping

Homulvas:
Useful for last tier sorting when albums have multiple CDs so that you could easily queue just one of the discs.


futwick:
Using this grouping, when disc number information is missing, it is displayed as "-1".

Would it be possible to have group-by-disc only for multi-disc albums and be ignored when disc number information not available?

PR #4599, opened 2014-10-29
Improve handling of Spotify Top Tracks and compilations

cincodenada:
Currently, Spotify results are problematic for two reasons, which are both demonstrated in this screenshot of the results for a single New Pornographers album:
2014-10-28-205030_331x546_scrot

Firstly, they are not assigned an album artist, so albums with different artists per track that should have a single album artist and therefore show up as one album when sorting by Album Artist are split up into two.  In this case, track 8 (Born With a Sound) features Amber Webber, so even though I am grouping by Album Artist, the album is split up between two artists.  When you add the album under The New Pornographers to your playlist, it is missing track 8, since that is filed under "The New Pornographers, Amber Webber".  My proposed solution here is to choose a "majority artist" by tallying up which artist shows up most often, and assigning that as the album artist.  This should be the "correct" solution in most cases, and even in edge cases (Broadway cast recordings, etc) will at least group the album into a single entry, albeit perhaps with an odd Album Artist entry.

The second issue is that the individual "top tracks" that Spotify provides retain their album title, so when sorted create one or more extra album entries, each of which only have a few songs in them, which is confusing, and not how Spotify presents them.  In the example above, all the top tracks are from one album, so we just end up with two copies of the album, one of which only has the 5 Top Tracks in it.

:Here's an example of the 5 Top Tracks being spread across multiple albums, demonstrating how things can get extra messy:

2014-10-28-204628_357x793_scrot

My proposed solution to this is to group these tracks into an artifical "Top Tracks" album, so that they show up in one place, and it is clear what they are.

This pull request implements both of those solutions, and the result looks like this:

2014-10-28-204950_332x491_scrot

The duplicate album entries are gone, and the Top Tracks are clearly marked as such.

There are some other ways this could be done, but I implemented the most conservative changes as I saw them.  An option that is more invasive but perhaps more widely helpful would be instead of assigning the Majority Artist to the Album Artist, simply set each track's Artist to the Majority Artist, since the default sort options don't involve Album Artist.  This would be straightforward to implement, and show better results to more people, at the cost of losing accuracy of metadata.  The name for "Top Tracks" is also arbitrary, it could be "[Spotify] Top Tracks" to make clear it's a placeholder album, or whatever makes the most sense.

A note on the patch itself: I have successfully compiled against the 1.2.3 release, and these changes with that base are in the spotify-improvements-1.2 branch.  I have been unable to compile master starting at 56c9498, due to some apparent issues in my build environment with Gstreamer 1.x, but it's a small patch that doesn't touch any GStreamer things, so I can't imagine this patch against master will cause any problems.

### DIFF
--- a/src/globalsearch/spotifysearchprovider.cpp
+++ b/src/globalsearch/spotifysearchprovider.cpp
@@ -89,6 +89,19 @@ void SpotifySearchProvider::SearchFinishedSlot(
   PendingState state = it.value();
   queries_.erase(it);
 
+  /* Here we clean up Spotify's results for our purposes
+   *
+   * Since Spotify doesn't give us an album artist,
+   * we pick one, so there's a single album artist
+   * per album to use for sorting.
+   *
+   * We also drop any of the single tracks returned
+   * if they are already represented in an album
+   *
+   * This eliminates frequent duplicates from the
+   * "Top Tracks" results that Spotify sometimes
+   * returns
+   */
   QMap<std::string, std::string> album_dedup;
   ResultList ret;
 
@@ -96,13 +109,20 @@ void SpotifySearchProvider::SearchFinishedSlot(
     const pb::spotify::Album& album = response.album(i);
 
     QHash<QString, int> artist_count;
-    QString artist, majority_artist;
+    QString majority_artist;
     int majority_count = 0;
 
-    // Choose an album artist
+    /* We go through and find the artist that is
+     * represented most frequently in the artist
+     *
+     * For most albums this will just be one artist,
+     * but this ensures we have a consistend album artist for
+     * soundtracks, compilations, contributing artists, etc
+     */
     for (int j = 0; j < album.track_size(); ++j) {
+      // Each track can have multiple artists attributed, check them all
       for (int k = 0; k < album.track(j).artist_size(); ++k) {
-        artist = QStringFromStdString(album.track(j).artist(k));
+        QString artist = QStringFromStdString(album.track(j).artist(k));
         if (artist_count.contains(artist)) {
           artist_count[artist]++;
         } else {
@@ -116,6 +136,8 @@ void SpotifySearchProvider::SearchFinishedSlot(
     }
 
     for (int j = 0; j < album.track_size(); ++j) {
+      // Insert the album/track title into the dedup map
+      // so we can check tracks against it below
       album_dedup.insertMulti(album.track(j).album(), album.track(j).title());
 
       Result result(this);
@@ -132,6 +154,8 @@ void SpotifySearchProvider::SearchFinishedSlot(
   for (int i = 0; i < response.result_size(); ++i) {
     const pb::spotify::Track& track = response.result(i);
 
+    // Check this track/album against tracks we've already seen
+    // in the album results, and skip if it's a duplicate
     if (album_dedup.contains(track.album()) &&
         album_dedup.values(track.album()).contains(track.title())) {
       continue;
--- a/src/globalsearch/spotifysearchprovider.cpp
+++ b/src/globalsearch/spotifysearchprovider.cpp
@@ -45,8 +45,7 @@ SpotifyServer* SpotifySearchProvider::server() {
   if (service_->login_state() != SpotifyService::LoginState_LoggedIn)
     return nullptr;
 
-  if (!service_->IsBlobInstalled())
-    return nullptr;
+  if (!service_->IsBlobInstalled()) return nullptr;
 
   server_ = service_->server();
   connect(server_, SIGNAL(SearchResults(pb::spotify::SearchResponse)),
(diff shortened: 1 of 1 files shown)


cincodenada:
@ArnaudBienner, I arrived at a similar conclusion, and have implemented it as such, at least for the Spotify search.  "Top Tracks" results are now simply removed if they already exist in the album results (simple track/album comparison), which will take care of most of the user-facing issues, I believe.  Here's what the searches above now look like:
2015-02-15-142829_346x357_scrot
2015-02-15-142913_314x579_scrot

This also addresses the issue that @Asfaloth pointed out and the album name issue, since single tracks will still show up if they're not in the album results, and the "Top Tracks" fake album is done away with entirely.

I think given the constraints of the unified search interface, and the lack of info (as far as I could tell) from the Spotify search provider about what section the results were in, that this is probably the best overall solution - it's the one I'm using :)

cincodenada:
Is there anything further to be done on this PR?  I appreciate the feedback thus far - you all helped track down some failure cases that I hadn't run into.  I think my latest commit addresses all the concerns raised fairly well, and is an improvement over the current handling.

I was reminded to check in here, in fact, because I ran into an instance today that had exceptionally undesirable results: 
2015-04-08-153511_404x656_scrot

This, of course, is just an extreme case of the album having many artists, without a unifying album artist, which is one of the two things this PR addresses.

src/globalsearch/spotifysearchprovider.cpp
```diff
@@ -1,4 +1,4 @@
-/* This file is part of Clementine.
+f/* This file is part of Clementine.
```
amuttsch: f?

amuttsch:
A wild `f` appread, please remove it, run `make format` and squash the commits. After this it can be merged.

cincodenada:
Indeed, will do.  Shall I squash the commits into one?

cincodenada:
Okay, possibly stupid question: how/from where do I run `make format`?  Is there a utility I need to install?  I remember running it before for other improvements, and I don't remember having to do anything fancy, but simply running `make format`, from the repo root, `bin`, or `bin/src` just gives me `*** No rule to make target 'format'.  Stop.`

I couldn't find any reference to it or instructions on the wiki either...I know I've done this before, but I can't figure out what I did!

amuttsch:
Normally from the bin dir. Have You merged Master recently? Since your pr is quite old, maybe make format wasn't added back then. 

On 15 April 2015 19:12:02 CEST, Joel Bradshaw <notifications@github.com> wrote:
>Okay, possibly stupid question: how/from where do I run `make format`? 
>Is there a utility I need to install?  I remember running it before for
>other improvements, and I don't remember having to do anything fancy,
>but simply running `make format`, from the repo root, `bin`, or
>`bin/src` just gives me `*** No rule to make target 'format'.  Stop.`
>
>I couldn't find any reference to it or instructions on the wiki
>either...I know I've done this before, but I can't figure out what I
>did!
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/clementine-player/Clementine/pull/4599#issuecomment-93489756

-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

cincodenada:
Okay, turns out there were some missing dependencies for the new stuff that was preventing cmake from finishing.  Got that all straightened out, ran `make format` and squashed things down.  Should be good to go, I think.

src/globalsearch/spotifysearchprovider.cpp
```diff
@@ -90,30 +89,60 @@ void SpotifySearchProvider::SearchFinishedSlot(
   PendingState state = it.value();
   queries_.erase(it);
 
+  QMap<std::string, std::string> album_dedup;
```
ArnaudBienner: I think it would be better to add a comment to explain the logic we now apply here i.e. that we will check for duplicate in results.

src/globalsearch/spotifysearchprovider.cpp
```diff
@@ -90,30 +89,60 @@ void SpotifySearchProvider::SearchFinishedSlot(
...
-    Result result(this);
-    SpotifyService::SongFromProtobuf(track, &result.metadata_);
-
-    ret << result;
-  }
 
   for (int i = 0; i < response.album_size(); ++i) {
     const pb::spotify::Album& album = response.album(i);
 
+    QHash<QString, int> artist_count;
```
ArnaudBienner: Maybe this could be in another helper function, to keep things clearer?

src/globalsearch/spotifysearchprovider.cpp
```diff
@@ -90,30 +89,60 @@ void SpotifySearchProvider::SearchFinishedSlot(
...
+    QString artist, majority_artist;
```
ArnaudBienner: I think artist can be defined later...

src/globalsearch/spotifysearchprovider.cpp
```diff
@@ -90,30 +89,60 @@ void SpotifySearchProvider::SearchFinishedSlot(
...
+    int majority_count = 0;
+
+    // Choose an album artist
+    for (int j = 0; j < album.track_size(); ++j) {
+      for (int k = 0; k < album.track(j).artist_size(); ++k) {
+        artist = QStringFromStdString(album.track(j).artist(k));
```
ArnaudBienner: ... here, because you don't need it before, and don't reuse it outer the loop.
Usually, I prefer to keep the closest scope for object.
I don't think there is an overhead to do so, and even if there is one, it is probably very small.

src/globalsearch/spotifysearchprovider.cpp
```diff
@@ -90,30 +89,60 @@ void SpotifySearchProvider::SearchFinishedSlot(
...
+        if (artist_count.contains(artist)) {
+          artist_count[artist]++;
+        } else {
+          artist_count[artist] = 1;
```
ArnaudBienner: I don't think you need the else if/else, because IIRC trying to access an element that doesn't exist will create it with default value (0). That will make the code smaller.
That being said, maybe this behavior is a bit confusing and it's better to keep a the code as it is.

ArnaudBienner:
Thanks a lot for your work on this :)
The patch looks good (haven't tested it yet though). My comments are not mandatory things: just things which I think can make your patch even better :) The only thing I would like to have is at least small comment to briefly explain the logic we apply here.

cincodenada:
Thanks for the feedback!  I've added some clarifying comments (perhaps too much!), and moved `artist` down in the scope.  I left it in a separate commit so you can see the changes, but I'll squash things back down once it's ready to go.

src/globalsearch/spotifysearchprovider.cpp
```diff
@@ -90,30 +89,84 @@ void SpotifySearchProvider::SearchFinishedSlot(
...
 
+    QHash<QString, int> artist_count;
+    QString majority_artist;
+    int majority_count = 0;
+
+    /* We go through and find the artist that is
+     * represented most frequently in the artist
+     *
+     * For most albums this will just be one artist,
+     * but this ensures we have a consistend album artist for
```
ArnaudBienner: typo
s/consistend/consistent

PR #4617, opened 2014-11-17
Added option to hide cover art in now playing widget (closes #4574)

### LINKED ISSUE #4574: Allow to do not display cover in panel at the bottom-left
Currently at the bottom-left there is a panel with displayed cover art. It can be changed to be large or small but it is impossible to completely disable it.

asiviero:
@mkoniecz suggested an option to hide the cover art in the now playing widget in #4574, so I've added that option which draws only the text with a black background.

### DIFF
--- a/src/widgets/nowplayingwidget.cpp
+++ b/src/widgets/nowplayingwidget.cpp
@@ -99,6 +99,8 @@ NowPlayingWidget::NowPlayingWidget(QWidget* parent)
   CreateModeAction(LargeSongDetailsBelow,
                    tr("Large album cover (details below)"), mode_group,
                    mode_mapper);
+  CreateModeAction(NoCoverDetails, tr("No album cover (details only)"),
+                   mode_group, mode_mapper);
 
   menu_->addActions(mode_group->actions());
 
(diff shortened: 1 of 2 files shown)


matkoniecz:
@asiviero Thanks for your work on this! 

@clementine-player It would be nice to see some kind of feedback, completely ignored pull requests (since November 2014) are not encouraging.

I considered making one for #5315 but I see that it would not be a good use of my time.

src/widgets/nowplayingwidget.cpp
```diff
@@ -492,6 +488,32 @@ void NowPlayingWidget::DrawContents(QPainter* p) {
...
+                               gradient_mid + kGradientTail);
+      gradient.setColorAt(0, QColor(0, 0, 0, 0));
+      gradient.setColorAt(1, QColor(0, 0, 0, 255));
+
+      p->fillRect(0, gradient_mid - kGradientHead, width(),
+                  height() - (gradient_mid - kGradientHead), gradient);
+
+      // Draw the text on top
+      p->translate(x_offset, height() - text_height);
+      details_->drawContents(p);
```
vincele, line 510: stray space

vincele (approved):
Looks sensible

PR #4682, opened 2014-12-29
Adding some fixer scripts in dist, formating is src/dbus and fix for mtime in podcasts

sobkas:
This fixer scripts are really ugly, but I use them and it's pain to move the commits out of the way for each pr
Formatting in src/dbus, I just couldn't help myself
mtime was broken after change in PodcastDeleter that set an invalid listened date after deleting episode.

### DIFF
--- a/src/internet/podcasts/podcastepisode.cpp
+++ b/src/internet/podcasts/podcastepisode.cpp
@@ -146,7 +146,17 @@ void PodcastEpisode::InitFromQuery(const QSqlQuery& query) {
   d->duration_secs_ = query.value(6).toInt();
   d->url_ = QUrl::fromEncoded(query.value(7).toByteArray());
   d->listened_ = query.value(8).toBool();
-  d->listened_date_ = QDateTime::fromTime_t(query.value(9).toUInt());
+
+  // After setting QDateTime to invalid state, it's saved into database as time_t,
+  // when this number std::numeric_limits<unsigned int>::max() (4294967295) is read back
+  // from database, it creates a valid QDateTime.
+  // So to make it behave consistently, this change is needed.
+  if (query.value(9).toUInt() == std::numeric_limits<unsigned int>::max()) {
+    d->listened_date_ = QDateTime();
+  } else {
+    d->listened_date_ = QDateTime::fromTime_t(query.value(9).toUInt());
+  }
+
   d->downloaded_ = query.value(10).toBool();
   d->local_url_ = QUrl::fromEncoded(query.value(11).toByteArray());
 
--- a/src/dbus/metatypes.h
+++ b/src/dbus/metatypes.h
@@ -1,8 +1,25 @@
-#ifndef DBUSMETATYPES_H
-#define DBUSMETATYPES_H
+/* This file is part of Clementine.
+   Copyright 2011, John Maguire <john.maguire@gmail.com>
+
+   Clementine is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation, either version 3 of the License, or
+   (at your option) any later version.
+
+   Clementine is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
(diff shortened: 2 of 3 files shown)


sobkas:
Uh, hello?

davidsansome:
Please check this Python against the style guide: https://google-styleguide.googlecode.com/svn/trunk/pyguide.html
Particularly indentation (spaces vs tabs), imports (import * is bad) and whitespace around operators (foo = bar instead of foo=bar)

dist/copyright_force.py
```diff
@@ -0,0 +1,159 @@
...
+	else:
+		return 0
+
+def gen_copyrights(f):
+
+	commits = []
+	data = {}
+        copyrights = []
+        
+	for ln in Popen(["git", "blame", "--incremental", f], stdout=PIPE).stdout:
```
davidsansome: subprocess.check_output seems better?

dist/copyright_force.py
```diff
@@ -0,0 +1,159 @@
...
+		except KeyError:
+			n = (c["author"], c["author-mail"], set())
+			by_author[c["author"]] = n
+
+		# FIXME: Handle time zones properly
+		year = datetime.fromtimestamp(int(c["author-time"])).year
+
+		n[2].add(year)
+
+	for an, a in list(by_author.iteritems()):
```
davidsansome: I'm finding it hard to follow this, but can you avoid the O(n^2) somehow?

dist/copyright_force.py
```diff
@@ -0,0 +1,159 @@
...
+
+def change_file(filename):
+        content=[]
+        out=[]
+        extended=0
+        ends=0
+        with open(filename, "r") as fi:
+                content=fi.readlines()
+        with open("dist/copyright.tmpl", "r") as fi:
+                templet=fi.readlines()
```
davidsansome: s/templet/template/?

dist/copyright_force.py
```diff
@@ -0,0 +1,159 @@
...
+		if x == prev + 1:
+			prev = x
+			continue
+		
+		if prev == start:
+			r.append("%i" % prev)
+		else:
+			r.append("%i-%i" % (start, prev))
+
+		start = x
```
davidsansome: It's not really "start" then is it...

dist/copyright_force.py
```diff
@@ -0,0 +1,159 @@
...
+		prev = x
+
+	if not prev is None:
```
davidsansome: minor: I think if x is not None: is preferred.

dist/copyright_force.py
```diff
@@ -0,0 +1,159 @@
+#!/usr/bin/python
+
+from subprocess import *
+from sys import *
+from os import rename, remove
+from datetime import *
+
+def pretty_years(s):
```
davidsansome: At a glance I have no idea what this function, or any other function in this file does.  Please add high-level docstrings, or rewrite the code to be more self-explanatory.

src/internet/podcasts/podcastepisode.cpp
```diff
@@ -146,7 +146,16 @@ void PodcastEpisode::InitFromQuery(const QSqlQuery& query) {
   d->duration_secs_ = query.value(6).toInt();
   d->url_ = QUrl::fromEncoded(query.value(7).toByteArray());
   d->listened_ = query.value(8).toBool();
-  d->listened_date_ = QDateTime::fromTime_t(query.value(9).toUInt());
+
+  // After setting QDateTime to invalid state, it's saved into database as time_t,
+  // when this number(4294967295) is read back from database, it creates a valid QDateTime.
+  // So to make it behave consistently, this change is needed.
+  if (query.value(9).toUInt() == 4294967295) {
```
davidsansome: Maybe this would be better as a separate PR?  And also use std::numeric_limits<uint32_t>::max() - 1 or something.

sobkas:
Python stuff shouldn't be in this pr, will remove it now(as it's not my code, only got it from http://0pointer.de/public/copyright.py), so if it's only used by me I will keep it private.
But if there is a need for it, I will merge it with dist/copyright.py and fix all the bugs
Also I will use ULONG_MAX

sobkas:
So now it should be fixed

sobkas:
Instead of ULONG_MAX(or UINT_MAX), I have used std::numeric_limits<unsigned int>::max(), more in C++ style

PR #4692, opened 2015-01-06
Added iPod-like behavior to previous button (closes #4638)

### LINKED ISSUE #4638: Previous track to restart current song
I would love to have an option for the "Previous track" command to restart the current song if more than 2 seconds of the current song played.

This is the behaviour of my trustworthy iPod and best reflect my need. I mostly use previous after a phone call to start the current song from the beginning. Playing previous songs is possible by quickly pressing the previous button twice or more.

asiviero:
Hey guys, @corco requested in #4638 that the previous button restarted the song if it were pressed, like the iPod behavior, so I added that option in Clementine's behavior page. I kept the 2 seconds mentioned by him as I don't own an iPod myself.

### DIFF
--- a/src/core/player.cpp
+++ b/src/core/player.cpp
@@ -29,10 +29,10 @@
 
 #include <memory>
 
+#include <QSettings>
 #include <QSortFilterProxyModel>
 #include <QtDebug>
 #include <QtConcurrentRun>
-#include <QSettings>
 
 #include "config.h"
 #include "core/application.h"
@@ -62,7 +62,8 @@ Player::Player(Application* app, QObject* parent)
       last_state_(Engine::Empty),
       nb_errors_received_(0),
       volume_before_mute_(50),
-      last_pressed_previous_(QDateTime::currentDateTime()) {
+      last_pressed_previous_(QDateTime::currentDateTime()),
+      menu_previousmode_(PreviousBehaviour_DontRestart) {
   settings_.beginGroup("Player");
 
   SetVolume(settings_.value("volume", 50).toInt());
@@ -88,6 +89,8 @@ void Player::Init() {
           SLOT(EngineMetadataReceived(Engine::SimpleMetaBundle)));
 
   engine_->SetVolume(settings_.value("volume", 50).toInt());
+
+  ReloadSettings();
 
 #ifdef HAVE_LIBLASTFM
   lastfm_ = app_->scrobbler();
--- a/src/core/player.cpp
+++ b/src/core/player.cpp
@@ -32,6 +32,7 @@
 #include <QSortFilterProxyModel>
 #include <QtDebug>
 #include <QtConcurrentRun>
+#include <QSettings>
 
 #include "config.h"
 #include "core/application.h"
@@ -50,6 +51,8 @@
 
 using std::shared_ptr;
 
+const char* Player::kSettingsGroup = "Player";
+
 Player::Player(Application* app, QObject* parent)
     : PlayerInterface(parent),
       app_(app),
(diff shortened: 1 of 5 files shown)


src/core/player.cpp
```diff
@@ -32,6 +32,7 @@
 #include <QSortFilterProxyModel>
 #include <QtDebug>
 #include <QtConcurrentRun>
+#include <QSettings>
```
ArnaudBienner: Please keep includes alphabetically ordered

src/core/player.h
```diff
@@ -31,6 +31,7 @@
 
 #include <QObject>
 #include <QSettings>
+#include <QDateTime>
```
ArnaudBienner: ditto

src/ui/behavioursettingspage.ui
```diff
@@ -178,6 +178,32 @@
...
+         <number>0</number>
+        </property>
+        <item>
+         <property name="text">
+          <string>Jump to previous song right away</string>
+         </property>
+        </item>
+        <item>
+         <property name="text">
+          <string>Restart song, then jump to previous if pressed again (iPod behavior)</string>
```
ArnaudBienner: I guess other players behave like this, so I'm not sure it's worth to have "(iPod behavior)" at the end.
Also, I think we use UK English everywhere, so this should probably be "(iPod behaviour)" instead.

hatstand: Actually, we should use US English and then translate it back to UK in transifex.

src/core/player.cpp
```diff
@@ -58,7 +61,8 @@ Player::Player(Application* app, QObject* parent)
       stream_change_type_(Engine::First),
       last_state_(Engine::Empty),
       nb_errors_received_(0),
-      volume_before_mute_(50) {
+      volume_before_mute_(50),
+      last_pressed_previous_(QDateTime::currentDateTime()) {
```
ArnaudBienner: FWIW, init menu_previousmode_ here, probably to PreviousBehaviour_DontRestart.

src/ui/mainwindow.cpp
```diff
@@ -216,6 +216,7 @@ MainWindow::MainWindow(Application* app, SystemTrayIcon* tray_icon, OSD* osd,
   int volume = app_->player()->GetVolume();
   ui_->volume->setValue(volume);
   VolumeChanged(volume);
+  app_->player()->ReloadSettings();
```
ArnaudBienner: Why doing this here?
Probably to init menu_previousmode_ value?
If so, it might be better to init it in Player's constructor (maybe by calling ReloadSettings? Not sure if there are other places where we do something similar).

src/ui/behavioursettingspage.ui
ArnaudBienner: OK, my bad. Thanks for the clarification :)

src/ui/mainwindow.cpp
asiviero: this is the one thing I couldn't quite get my head around while coding it. I hadn't used `QSettings` before so I tried to mimic `MainWindow` settings behavior. I agree that moving to Player's constructor would make more sense. `MainWindow` constructor also calls a series of `ReloadSettings` (including its own) so its components can init their user defined settings. 

Would it be ok moving to `Player` constructor?

asiviero:
Adjustments done :D

src/core/player.h
```diff
@@ -119,6 +120,14 @@ class Player : public PlayerInterface {
   explicit Player(Application* app, QObject* parent = nullptr);
   ~Player();
 
+  static const char* kSettingsGroup;
+
+  // Don't change the values
```
ArnaudBienner: because they are stored in the settings, right?
Maybe better to write this down, to make things really clear and meaningful FWIW

ArnaudBienner:
Sorry for taking so long. Sounds good :) (just one small comment)
I didn't take the time to test it for real (I would like to do this before merging)
Will try asap

corco:
I've been using it this week and it's working great!

Thanks @asiviero for the implementation. Please merge!

src/core/player.h
asiviero: I believe it was written like this somewhere else in the code, but that is indeed the reason. I will clarify it

ArnaudBienner:
Thanks for the patch :)

PR #4715, opened 2015-01-16
Fix album artist for FLAC/OGG files (with vorbis comments)

### LINKED ISSUE #3194: Album Artist cannot be edited in flac in windows
_From [Zade...@gmail.com](https://code.google.com/u/113851922626882860131/) on October 02, 2012 23:50:28_

What steps will reproduce the problem? 1.Select edit track information on a flac file
2.Change the album artist
3.Save What is the expected output? What do you see instead? Expected: the album artist to be changed
Instead: no changes What version of the product are you using? On what operating system? Problem appears in Windows 7 and Windows 8, Clementine Version 1.0.1-670-gbb710ed Please provide any additional information below.

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=3193_

Chocobozzz:
Fixes #3194

### DIFF
--- a/ext/libclementine-tagreader/tagreader.cpp
+++ b/ext/libclementine-tagreader/tagreader.cpp
@@ -502,10 +502,9 @@ void TagReader::ParseOggTag(const TagLib::Ogg::FieldListMap& map,
                     100);
 }
 
-void TagReader::SetVorbisComments(TagLib::Ogg::XiphComment* vorbis_comments,
-                                  const pb::tagreader::SongMetadata& song)
-    const {
-
+void TagReader::SetVorbisComments(
+    TagLib::Ogg::XiphComment* vorbis_comments,
+    const pb::tagreader::SongMetadata& song) const {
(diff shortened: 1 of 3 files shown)


ArnaudBienner:
Would be nice to also read back this in ParseOggTag.
Also, is this standard? (I don't remember if other fields we are setting in comment are standards, and if it is standard to do it this way. I'm just wondering).

Also, it might be worth to add some unit test for this (in song_test.cpp?)
Maybe it is a bit overkill. But this can also help you to test this is working well, and it's probably not long to do. As you prefer.

Chocobozzz:
Yes good idea @ArnaudBienner for the tests I'll do it when I'll have time :)

ParseOggTag already implemented the ALBUM ARTIST field (line 470).

Infortunaly it doesn't seem that album artist field is a standard, but I always saw people use "ALBUMARTIST" or "ALBUM ARTIST" field name so I think it will be portable.

Chocobozzz:
It's done :)

PR #4731, opened 2015-01-31
Add support for custom lastfm:// URLs

### LINKED ISSUE #1: scrobbling support for more providers than just last.fm
_From [mattl-lists@fsf.org](https://code.google.com/u/100106606012998798867/) on February 22, 2010 22:57:45_

Can you offer people the chance to enter any host name they wish for
scrobbling? So, if people wish to use turtle.libre.fm instead of
post.audioscrobbler.com they can?

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=1_

Chocobozzz:
I removed liblastfm as dependecy (see the other pull request of the Dependencies repo) in the purpose that Clementine builds itself custom liblastfm in which we can override the host (see this commit of my fork to view the changes from the official library https://github.com/Chocobozzz/liblastfm/commit/e54e539547930ec142f2f077b722620db116b214 to facilitate the review).

So now we are sure user have liblastfm1, I removed the `HAVE_LIBLASTFM1` defines.

If the server field is not set the behaviour is the same as before.

Closes #1


hatstand:
We don't really want to ship custom versions of libraries if we can possibly avoid it. Fedora in particular dislikes it, better to submit patches to liblastfm first.

Chocobozzz:
I understand. I chose this solution because I hadn't any protests of my comment in https://github.com/clementine-player/Clementine/issues/1

Submitted a PR yesterday, just hope they'll accept.

PR #4753, opened 2015-02-19
Reorganize the CD ripper

paperbagcorner:
The code for the CD ripper has become somewhat unwieldy. I have therefore split the ripper into two classes, one RipCDDialog class for the user interface and one Ripper class that handles the ripping.

The Rip audio CD menu item is moved to the Tools menu since I believe it is a more natural place.

The commit f7185101cea9d672a2fc4afc16d389e8afbd832f fixes a bug were the wrong file was tagged after the transcoder had changed the output filename.


src/ripper/ripcddialog.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+  setWindowTitle(tr("Rip CD"));
+  AddDestinationDirectory(QDir::homePath());
+
+  // Get presets
+  QList<TranscoderPreset> presets = Transcoder::GetAllPresets();
+  qSort(presets.begin(), presets.end(), ComparePresetsByName);
+  for (const TranscoderPreset& preset : presets) {
+    ui_->format->addItem(
+        QString("%1 (.%2)").arg(preset.name_, preset.extension_),
```
hatstand: `.arg(a).arg(b)` is preferred as it avoids future bugs when formatting numbers due to the multiple overloads of `arg()`

src/ripper/ripcddialog.cpp
```diff
@@ -0,0 +1,299 @@
...
+}
+
+void RipCDDialog::InvertSelection() {
+  for (QCheckBox* checkbox : checkboxes_) {
+    if (checkbox->isChecked()) {
+      checkbox->setCheckState(Qt::Unchecked);
+    } else {
+      checkbox->setCheckState(Qt::Checked);
+    }
+  }
```
hatstand: `checkbox->setCheckState(checkbox->isChecked() ? Qt::Unchecked : Qt:Checked);`

src/ripper/ripper.cpp
```diff
@@ -0,0 +1,305 @@
...
+    cancel_requested_ = true;
+  }
+  transcoder_->Cancel();
+  RemoveTemporaryDirectory();
+  emit(Cancelled());
+}
+
+void Ripper::TranscodingJobComplete(const QString& input, const QString& output,
+                                    bool success) {
+  (*(success ? &finished_success_ : &finished_failed_))++;
```
hatstand: I think a simple `if` statement is more readable.

src/ripper/ripper.cpp
```diff
@@ -0,0 +1,305 @@
...
+  UpdateProgress();
+
+  // The the transcoder does not overwrite files. Instead, it changes
+  // the name of the output file. We need to update the transcoded
+  // filename for the corresponding track so that we tag the correct
+  // file later on.
+  for (TrackInformation& track : tracks_) {
```
hatstand: Don't use non-const references

src/ripper/ripper.cpp
paperbagcorner: It is and you've commented on in [before](https://github.com/clementine-player/Clementine/pull/4113/files#r8858034). I will change it.

hatstand: I thought I recognised it :-)

PR #4758, opened 2015-02-21
Try to make Seafile implementation more robust

Chocobozzz:
Just added informations via the taskmanager because the build of the Seafile index could be pretty long if the library have many directories

Now Clementine waits the previous index build if the user wants to change the library, it's cleaner.

Finally ReadTagsFinished ignore replies of a previous indexing which was cancelled. It's not the best solution since the tagreaders still download informations about the tracks but I didn't find a better one.

### DIFF
--- a/src/internet/seafile/seafileservice.cpp
+++ b/src/internet/seafile/seafileservice.cpp
@@ -26,6 +26,7 @@
 #include <QTimer>
 
 #include "core/application.h"
+#include "core/taskmanager.h"
 #include "core/player.h"
 #include "core/waitforsignal.h"
 #include "internet/seafile/seafileurlhandler.h"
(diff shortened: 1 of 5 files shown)


src/internet/seafile/seafileservice.cpp
```diff
@@ -196,6 +217,15 @@ void SeafileService::Connect() {
 }
 
 void SeafileService::UpdateLibraries() {
+  if (indexing_task_id_ == -1) {
+    indexing_task_id_ =
+        app_->task_manager()->StartTask(tr("Build index of Seafile"));
```
hatstand: "Indexing Seafile..."

src/internet/seafile/seafileservice.cpp
```diff
@@ -196,6 +217,15 @@ void SeafileService::Connect() {
...
+  }
+  // Quit if we are already updating the libraries
+  else {
```
hatstand: Do this at the start of the function, i.e.:
```c++
if (indexing_task_id_ != -1) {
  return;
}
```

hatstand:
Please run `make format` again too

src/internet/seafile/seafileservice.cpp
Chocobozzz: Are you sure ? Because this task is different of the CloudFileService (MaybeAddFileToDatabase) one. The user would have two tasks named "Indexing Seafile..." : 
 * one when the Seafile service try to build the index (by checking which directory of the server was updated with the local tree)
 * one (which could be executed in parallel) to index songs of the updated directories (MaybeAddFileToDatabase)

Anyway I think the current correct form would be "Building Seafile index..." no ?

hatstand: Sounds good

Chocobozzz:
Done :)

PR #4776, opened 2015-03-11
Rewrite of PodcastParser to use tinyxml2 as parser(partial), add support...

### LINKED ISSUE #4771: Can not play Podcast - Atom RSS feed
Hi all,

Other Podcast is working, but these type is not. :(
I can not play Atom RSS feed Podcasts, like this:
https://tilos.hu/feed/show/dzsungel-konyve

The problem is 
unrecognized type: application/atom+xml

sobkas:
... for ISO8601 time format, support for application/atom+xml, closes #4771


hatstand:
I'd prefer not adding new dependencies as it's a pain on mac and windows. If you really want a DOM API, there's a perfectly good implementation in Qt already anyway.

sobkas:
Dropping tinyxml2 shouldn't be that problematic
But what do you think about dropping opmlcontainer and switching to PodcastList

PR #4779, opened 2015-03-16
Add "opening issue" to the README

Chocobozzz:
I think Clementine needs this section to help people (and us) when they want to open an issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,28 @@ Clementine is a modern music player and library organizer for Windows, Linux and
 - Buildbot: http://buildbot.clementine-player.org/grid
 - Latest developer builds: http://builds.clementine-player.org/
 
+Opening an issue
+----------------
+### Ask for a new feature
+
+Please:
+
+ * Check if the new feature is not already implemented (Changelog)
+ * Check if another person didn't already open an issue
+ * If there is already an opened issue there is no need to comment "+1", it won't help. Instead, you can subscribe to the issue to be notified of anything new about it
+
+### Report a bug
+
+Please:
+ 
+ * Try with the latest developer build (http://builds.clementine-player.org/) to see if the bug is still present (**Attention**, those builds aren't stable so they might not work well and could sometimes break things like user settings). If if works like a charm although you see an opened issue please comment it to explain the issue was fixed
+ * Check if another person didn't already open an issue
+ * If there is already an opened issue, you can try to comment it to add precisions about the problem or confirm it
+ * If there isn't, you can open a new issue with an explicit title and much information as possible (OS, Clementine version, what steps will reproduce the problem...)
+ * Please use http://pastebin.com/ for logs/debug
+ 
+If there is no answers, it doesn't mean we don't care about your feature request/bug, it just mean we can't reproduce the bug or haven't time to implement it :o)
+
 Compiling from source
 ---------------------
 


README.md
```diff
@@ -8,6 +8,28 @@ Clementine is a modern music player and library organizer for Windows, Linux and
 - Buildbot: http://buildbot.clementine-player.org/grid
 - Latest developer builds: http://builds.clementine-player.org/
 
+Opening an issue
+----------------
+### Ask a new feature
+
+Please :
```
ArnaudBienner: In English, no space before double-punctuation signs (unlike French for example).

README.md
```diff
@@ -8,6 +8,28 @@ Clementine is a modern music player and library organizer for Windows, Linux and
...
+
+ * Check if the new feature is not already implemented (Changelog)
+ * Check if another person didn't already open an issue
+ * If there is already an opened issue there is no need to comment "+1", it won't help :o)
```
ArnaudBienner: Maybe adding a "instead, you can subscribe to the issue to be notified of anything new about it" or something like this.

README.md
```diff
@@ -8,6 +8,28 @@ Clementine is a modern music player and library organizer for Windows, Linux and
...
+
+### Report a bug
+
+Please :
+ 
+ * Try with the latest developer build to see if the bug is still present. If if works like a charm although you see an opened issue please comment it to explain the issue was fixed
```
ArnaudBienner: Would be nice to add the link: http://builds.clementine-player.org/
Maybe warn users that those builds aren't stable release, and might not work well, and sometimes break things like user settings?
Unfortunately, we currently don't have "profiles" (like Firefox for examples) so users can try developers builds with a "Test" profile to be sure they will not break anything in their current profile.

README.md
```diff
@@ -8,6 +8,28 @@ Clementine is a modern music player and library organizer for Windows, Linux and
 - Buildbot: http://buildbot.clementine-player.org/grid
 - Latest developer builds: http://builds.clementine-player.org/
 
+Opening an issue
+----------------
+### Ask a new feature
```
ArnaudBienner: "Ask for"? But I'm not a native English speaker, @hatstand or @davidsansome would know better :)

ArnaudBienner:
Sounds like a good idea :)

README.md
Chocobozzz: You mean break the settings ? It's a good idea.

Chocobozzz:
Thanks :)

Changes done, just wait for the "ask for".

README.md
ArnaudBienner: Yes. Or the database where all info about songs are indexed. I think this happened once or twice in the past.
I don't think we ever had anything more serious, like deleting songs by accident.
But there is also the problem that, in the case of some db schema change for example, this is not backward compatible, so users cannot downgrade Clementine after IIRC.
But well, this doesn't happen very often, so just warn "this might break few things like users settings" might be enough.
Actually, the main problem I think users might have when trying developers builds is that Clementine wouldn't start because a dependency isn't indicated, a DLL is missing, etc. or crash somewhere else, so it's not a big deal: in this case they can safely reinstall the last stable one without any problem.

README.md
Asfaloth: What about `Want to ask for a new feature?` This should be grammatically correct and sound quite nice imho.

ArnaudBienner: I like "Ask for a new feature" because it's shorter, thus clearer IMHO. But I don't have a strong opinion about this.

Asfaloth: I'm fine with that too. But the `for` must be added definitely ;)

Chocobozzz:
So I added the "for", I agree with you it's better :)

hatstand:
Can you squash this into one commit?

Chocobozzz:
Yes, it was many commits just for a Readme :p

Issue #4783, opened 2015-03-22
Feature: Adding support to read from IDv2 tag - Patch included


M-Bab:
Support for tag reading is an often requested feature of the Clementine music player. For me it was the last feature missing to the ultimate music player. Although the ultimatelyricsreader is already pretty strong it sometimes cant find lyrics of local songs. Of course a tag reader is also faster and doesn't need an internet connection.
Therefore I developed a tag reader mechanism for which the patch is included at the bottom of the thread.

What works and how:
- ext tag reader checks for USLT or SYLT frames and stores them
- Database schema 48 added to have a lyrics field in the music database
- Song code modified to read/write lyrics stored in database
- Lyricsprovider written to show the lyrics in the songinfo tab

What doesn't work:
- The Lyricsprovider itself unfortunately :-(. If I replace some code in the Ultimatelyricsprovider the tag lyrics are displayed. But the taglyricsinfoprovider does not trigger the display of the CollapsibleInfoPane on the InfoResultReady-Signal. Although taglyricsinfoprovider as Q_OBJECT inherits from SongInfoProvider and is added to SongInfoView.fetcher_ . Any ideas/tips on this are appreciated. Anything else works.

Further possible improvements:
If someone is interested and the tagreader may join the Clementine code there are more ideas how I could extend it:
- Options if the tagreader shall be actually activated
- Edit lyrics in the tag editor
- Show ultimatelyrics only if no taglyrics are available
- Store ultimatelyrics in the tag lyrics if empty

```PATCH
4b84abf523da94b877cdd1b2af37b4fc26835719
 data/data.qrc                                      |   1 +
 data/schema/schema-48.sql                          |   4 +
 ext/libclementine-tagreader/tagreader.cpp          |  14 ++
 .../tagreadermessages.proto                        |   1 +
 src/CMakeLists.txt                                 |   2 +
 src/core/database.cpp                              |   2 +-
 src/core/mpris1.cpp                                |   1 +
 src/core/organiseformat.cpp                        |   5 +-
 src/core/song.cpp                                  |  15 +-
 src/core/song.h                                    |   2 +
 src/songinfo/songinfoview.cpp                      |   2 +
 src/songinfo/taglyricsinfoprovider.cpp             |  41 ++++
 src/songinfo/taglyricsinfoprovider.h               |  31 +++
 src/ui/mainwindow.cpp                              |   1 +
 src/ui/organisedialog.cpp                          |   1 +
[...]

PR #4784, opened 2015-03-23
Added support to read/display the ID tag lyrics in MP3 files:

M-Bab:
The requested pull request to issue 4783 (https://github.com/clementine-player/Clementine/issues/4783).

-Added schema to the database to store it
-Added readers/writers for ID tags
-Added readers/writers for the database to the song class
-Added the taglyricsinfoprovider to show the lyrics in songinfo

Remaining bug:
The taglyricsinfoprovider does not trigger the display of the CollapsibleInfoPane on the InfoResultReady-Signal.
Although taglyricsinfoprovider as Q_OBJECT inherits from SongInfoProvider and is added to SongInfoView.fetcher_ .
Any ideas/tips on this are appreciated. Anything else works.

### DIFF
--- a/src/core/metatypes.cpp
+++ b/src/core/metatypes.cpp
@@ -32,11 +32,12 @@
 #include "globalsearch/searchprovider.h"
 #include "internet/digitally/digitallyimportedclient.h"
 #include "internet/core/geolocator.h"
+#include "internet/podcasts/podcastepisode.h"
+#include "internet/podcasts/podcast.h"
 #include "internet/somafm/somafmservice.h"
 #include "library/directory.h"
 #include "playlist/playlist.h"
-#include "internet/podcasts/podcastepisode.h"
-#include "internet/podcasts/podcast.h"
+#include "songinfo/collapsibleinfopane.h"
 #include "ui/equalizer.h"
 
 #ifdef HAVE_VK
@@ -53,6 +54,7 @@ class GstEnginePipeline;
 class QNetworkReply;
 
 void RegisterMetaTypes() {
+  qRegisterMetaType<CollapsibleInfoPane::Data>("CollapsibleInfoPane::Data");
   qRegisterMetaType<ColumnAlignmentMap>("ColumnAlignmentMap");
   qRegisterMetaType<const char*>("const char*");
   qRegisterMetaType<CoverSearchResult>("CoverSearchResult");
@@ -74,16 +76,16 @@ void RegisterMetaTypes() {
   qRegisterMetaType<PlaylistItemPtr>("PlaylistItemPtr");
   qRegisterMetaType<PodcastEpisodeList>("PodcastEpisodeList");
   qRegisterMetaType<PodcastList>("PodcastList");
-  qRegisterMetaType<QList<CoverSearchResult> >("QList<CoverSearchResult>");
-  qRegisterMetaType<QList<PlaylistItemPtr> >("QList<PlaylistItemPtr>");
+  qRegisterMetaType<QList<CoverSearchResult>>("QList<CoverSearchResult>");
+  qRegisterMetaType<QList<PlaylistItemPtr>>("QList<PlaylistItemPtr>");
   qRegisterMetaType<PlaylistSequence::RepeatMode>(
       "PlaylistSequence::RepeatMode");
   qRegisterMetaType<PlaylistSequence::ShuffleMode>(
       "PlaylistSequence::ShuffleMode");
-  qRegisterMetaType<QList<PodcastEpisode> >("QList<PodcastEpisode>");
-  qRegisterMetaType<QList<Podcast> >("QList<Podcast>");
-  qRegisterMetaType<QList<QNetworkCookie> >("QList<QNetworkCookie>");
-  qRegisterMetaType<QList<Song> >("QList<Song>");
+  qRegisterMetaType<QList<PodcastEpisode>>("QList<PodcastEpisode>");
+  qRegisterMetaType<QList<Podcast>>("QList<Podcast>");
+  qRegisterMetaType<QList<QNetworkCookie>>("QList<QNetworkCookie>");
+  qRegisterMetaType<QList<Song>>("QList<Song>");
   qRegisterMetaType<QNetworkCookie>("QNetworkCookie");
   qRegisterMetaType<QNetworkReply*>("QNetworkReply*");
   qRegisterMetaType<QNetworkReply**>("QNetworkReply**");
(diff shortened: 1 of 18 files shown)


src/core/organiseformat.cpp
```diff
@@ -50,7 +50,8 @@ const QStringList OrganiseFormat::kKnownTags = QStringList() << "title"
                                                              << "samplerate"
                                                              << "extension"
                                                              << "performer"
-                                                             << "grouping";
+                                                             << "grouping"
+                                                             << "lyircs";
```
hatstand: Typo

src/core/song.cpp
```diff
@@ -513,6 +518,8 @@ void Song::InitFromProtobuf(const pb::tagreader::SongMetadata& pb) {
   d->filetype_ = static_cast<FileType>(pb.type());
   d->etag_ = QStringFromStdString(pb.etag());
 
+  std::string tmp_lyrics = pb.lyrics();
```
hatstand: Is this used anywhere?

src/songinfo/songinfoview.cpp
```diff
@@ -20,6 +20,7 @@
 #include "songinfoview.h"
 #include "ultimatelyricsprovider.h"
 #include "ultimatelyricsreader.h"
+#include "taglyricsinfoprovider.h"
```
hatstand: alpha-sort

hatstand:
Please run `make format` too

hatstand:
https://github.com/M-Bab/Clementine/blob/master/src/songinfo/songinfoview.cpp#L130 might be why it's not working.

M-Bab:
Thanks for reviewing the code!
1. Yeah thats a typo...
2. You might be right and its not used anymore. Have to check. Maybe a remnant
from debugging.
3. alpha-sort, ok
4. What happens by running "make format" at Clementine?
5. I also checked
https://github.com/M-Bab/Clementine/blob/master/src/songinfo/songinfoview.cpp#L130
on my bug search - but I doubt its the cause. Because it should skip my
lyricsprovider when it can not cast it to an ultimatelyricsprovider object.
Disabling lyrics sources like this seams to be an exclusive feature of the
ultimatelyricsprovider.

Asfaloth:
`make format` formats the code to match the google code style guide. Executing this makes sure your code is formatted correctly.

For your emitting problem: Add debug outputs to backtrack the source of the problem. Is the `emit' executed? Does your code even reach the `emit`? Otherwise you can compile Clementine with debug information and start debugging with `gdb`.

M-Bab:
Patched the mentioned issues by hatstand and ran the make format before a fixup commit.

Maybe some more details on the remaining bug I extracted from my debugging:
1. Like all lyricsproviders the signal InfoResultReady is emitted at code line songinfofetcher.cpp:72
2. Like all lyricsproviders the corresponding slot in songinfoview.cpp:80 is called
3. UNLIKE to the other lyricsproviders it does not react on "AddSection(new CollapsibleInfoPane(data, this));" while the ultimatelyricsprovider/lastfmprovider jumps to songinfobase.cpp:88.

I dont get the different behavior since all providers generate a CollapsibleInfoPane.

amuttsch:
Okay, I found the issue. In `TagLyricsInfoProvider` you aren't doing any async operations. The problem is that both objects (`TagLyricsInfoProvider` and `SongInfoFetcher`) are in the same thread and due the default `connect` behaviour, the slot is executed immediatly (see http://doc.qt.io/qt-4.8/qt.html#ConnectionType-enum).

This causes the `SongInfoView` to not know `current_request_id_`, which is set to the result of `FetchInfo`, see `SongInfoBase::Update`.

So when `SongInfoView::InfoResultReady` is called, `id != current_request_id_` is true and the result isn't added to the titles.

amuttsch:
Okay, to fix this add the following stuff:

`core/metatypes.cpp`:
Register `CollapsibleInfoPane::Data`.

`SongInfoFetcher::AddProvider`
Add `Qt::QueuedConnection` to the three connects.

M-Bab:
Awesome! As far as I tested this trick fixes the problem and the lyrics are displayed. I will update my pull request soon so it can be implemented. Direct async exchange between the threads - tricky bug.

Do these changes may have any negative impact on the other Clementine modules?

amuttsch:
I don't think so. The slots might be executed a bit later than now, but this won't be an issue.

Please update your fork to master and squash your commits into one, so we can merge it. And run `make format` once more for good luck ;-)

M-Bab:
All right. Merged the current master and squashed my commits. "make format" is now in my QtCreator build chain anyway so I can't forget it anymore ;-)

src/core/song.cpp
```diff
@@ -1058,7 +1065,7 @@ bool Song::IsMetadataEqual(const Song& other) const {
          d->samplerate_ == other.d->samplerate_ &&
          d->art_automatic_ == other.d->art_automatic_ &&
          d->art_manual_ == other.d->art_manual_ &&
-         d->cue_path_ == other.d->cue_path_;
+         d->cue_path_ == other.d->cue_path_ && d->lyrics_ == other.d->lyrics_;
```
amuttsch: Newline after `&&`

amuttsch:
One small code style issue, otherwise we can merge this.

M-Bab:
All right, included!

M-Bab:
Wow this schema number change by the amazon cloud confused me a bit. Now it should be up-to-date and correct everywhere.

src/core/metatypes.cpp
```diff
@@ -25,6 +25,7 @@
 #include <QMetaType>
 #include <QNetworkCookie>
 
+#include "songinfo/collapsibleinfopane.h"
```
amuttsch: Alpha sort.

amuttsch:
One tiny code style issue, then we are good to go :)

M-Bab:
Finally ... hopefully.

amuttsch:
Looks good. Thanks for your contribution :-)

Chocobozzz:
I think I have an issue with this PR because Jamendo has not lyrics column and thus there is a problem in `LibraryBackend::AddOrUpdateSongs`: the method try to insert lyrics.

M-Bab:
Interesting point: Should the schema actually be applied to %allsongstables or just the songs table? I used the %allsongstables because its recommended causing in general LESS problems (and maybe other modules want to store lyrics in the DB as well). If other Clementine modules are affected by the new column maybe the schema should only apply to the songs table?

Chocobozzz:
Even if jamendo do not use some columns, it has to have the same scheme than songs tables (see comment in jamendo.sql). If it's not the case this causes issues in `LibraryBackend::AddOrUpdateSongs`. 

It seems that %allsongtables doesn't include Jamendo songs table.

amuttsch:
I added the lyrics field in the jamendo db. Should work now.

stifi:
Thanks for implementing this feature! I have been waiting for this a very long time now.

I see decoding errors of non-ASCII letters in the lyrics. The root cause can might be found in the line <code>song->set_lyrics(lyrics.toStdString()); </code> I think  the <code>TagReader::Decode</code>method can solve this issue. The decoding errors appear in Clementine's Title Info only. The debug output into the shell seems fine.

Sorry, for not creating a patch. I have very little programming skills only.

M-Bab:
Yeah thats correct. I fixed it successfully (at least for german special characters) but wanted to collect some more features before the new pull request:
1. Edit lyrics in tag editor
2. Update lyrics with web lyrics if empty (and the option is selected in clementine config)

M-Bab:
@stifi Your issue will be fixed with the new pull request https://github.com/clementine-player/Clementine/pull/4890

stifi:
Encodings are now fine thanks, thanks!

However, I encountered an unwanted behavior when modifying the lyrics with the tag editor. If there are more than one lyrics stored in file (according to the ID3v2 specification this is possible and I sometimes use this feature) all lyrics are deleted except the ones shown in the tag editor. Additionally, the *Content descriptor* is cleared and the *language*  is set to XXX.

@M-Bab Is it a good practice to file issues here or should I make use of the Issues functionality at github?

M-Bab:
@stifi You should file issues either in the issues tab or at least in the corresponding pull request https://github.com/clementine-player/Clementine/pull/4890 . Because your issue is not really linked to the pull request where you commented.

PR #4796, opened 2015-04-08
Re-read FMPS_Rating from file when changed

alphapapa:
When the FMPS_Rating changes in a file, Clementine will now re-read it
and change the rating in the library database.  This lets you modify
song ratings outside of Clementine and then load the new ratings into
Clementine.  Previously Clementine would ignore changed ratings in a
file's metadata after a song had been scanned into the library.

This is consistent with the handling of other standardized tags (e.g. artist, album, title).  It's especially useful if a user has a shared music library which he accesses from multiple systems.


ArnaudBienner:
Thanks for the patch.
I've merged it manually in commit 44cd2f916c3f816681a05ac9719650231813dbe7 since some modifications happened in the meantime in song.cpp that would prevent your patch to apply.

alphapapa:
Thanks.

PR #4802, opened 2015-04-11
[WIP] Port to Qt5

Chocobozzz:
This is the first shot for a port to Qt5. ~~There isn't the last commits of the master branch~~. It compiles and I can play musics on Archlinux (if I install Qt5 versions of some libs as qca for example).

**What has been done:**
  * Desactived all optional components.
  * Updated the CMakeList but this is ugly: we need to choose the cmake minimum version and then adapt all the CMakeLists
  * Updated the sqlite plugin
  * ~~Replaced QtSingleApplication by this Qt5 alternative https://github.com/itay-grudev/SingleApplication~~ Updated QtSingleApplication
  * Updated libqxt
  * Rewrited toAscii() by toLatin1
  * Rewrited Qt::Escape by toHtmlEscaped()
  * Rewrited setResizeMode by setSectionResizeMode
  * Rewrited QSqlQuery prepare
  * Rewrited the qjson calls with the official QJson of Qt5
  * Rewrited url paths to be absolute
  * Rewrited QCleanLookStyle and QWindowsStyle by QCommonStyle (PlaylistProxyStyle, FancyTabWidget)
  * There was a use of QHttp in jamendo to avoid the keep alive connection. Is this is still needed ? I replaced with a QNetworkManager (temporarily).

**Known issues:**
  * ~~"QFSFileEngine::open: No file name specified" message from albumcoverloader.cpp:201 when filename == "" (reproduce : change current song)~~
  * ~~"QIODevice::write: device not open" message but I don't know what is the cause (reproduce : change current song)~~
  * Sometimes the tray icon does't appear. ~~I don't know why (for now).~~ It seems to be a qt5 regression (https://bugreports.qt.io/browse/QTBUG-31762). This should be fixed with 5.5. Wait & see.
  * Certainly others because I have not tested many things.

**Removed features (to re implement later of course):**
  * ~~Passing parameters to existing Clementine application (due to the replacement of QtSingleApplication). I have not implemented it yet because I don't know what is the best way to replace QtSingleApplication (I just use SingleApplication for now).~~
  * ~~All the disabled things in CMakeList~~

**What's next:**
 * Adapt correctly CMakeLists
 * ~~Better use of SingleApplication~~
 * ~~Reactive all optionals components and (and so adapt the 3rdparty libraries)~~
 * Try to compile and test on Mac/Windows
 * Test and fix potentials bugs
 * ~~Merge latest commit from master branch~~
 * Add qt5 libraries which can not be in all supported distributions (qca for example)
 * Certainly many other things I forgot :)

Of course I don't want merge it in the master branch. It's just for people who want to participate, give some advices, reviews...

I tried to separate the updated files in different commits and didn't make format for more clarity.


ArnaudBienner:
Why did you get rid of QtSingleApplication? I'm pretty sure it should work with Qt5 (maybe with some minor modifications though).

Chocobozzz:
Because it uses deprecated classes so I thought we could use a more elegant option. But yeah I think I can try with this https://qt.gitorious.org/qt-solutions/lgeyers-qt-solutions/source/c50d0f3f142606b823f9106c7cd8b9a7d066e6ff:qtsingleapplication/src

ArnaudBienner:
OK. Why not then, if we don't miss anything big in the alternative we would like to use.
BTW, great work :) Didn't have a deep look yet, but that sounds promising :)

hatstand:
Happy to merge this as a WIP on a different branch if you can make it merge cleanly to master again.

Chocobozzz:
Good idea. I'll make it in this days.

Chocobozzz:
Okay so it's done :)

hatstand:
Merged into qt5 branch.

PR #4810, opened 2015-04-18
Prevent suspend mode

miku84:
I have autosuspend configured after 10minutes in Plasma5 and it goes suspended when nothing happens, only song playing. I think Clementine should not go to suspend as VLC and qbittorrent does not go to suspend when active. 

https://bugs.kde.org/show_bug.cgi?id=346323
The application needs to call Inhibit on the org.freedesktop.PowerManagement.Inhibit DBus interface.


ArnaudBienner:
This should be an issue, not a pull request. Not sure how you managed to open this.

PR #4814, opened 2015-04-19
Adding cover art retrieval for Subsonic.

davidgfnet:
You guys probably want some sort of mechanism to generalize this hack :D

### DIFF
--- a/src/translations/af.po
+++ b/src/translations/af.po
@@ -400,7 +400,7 @@ msgstr "AAC 64k"
 msgid "AIFF"
 msgstr "AIFF"
 
-#: widgets/nowplayingwidget.cpp:149
+#: widgets/nowplayingwidget.cpp:151
 msgid "ALL GLORY TO THE HYPNOTOAD"
 msgstr "Alle glorie aan die HYPNOTOAD"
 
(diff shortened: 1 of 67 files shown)


src/covers/albumcoverloader.h
```diff
@@ -104,11 +105,12 @@ class AlbumCoverLoader : public QObject {
   QQueue<Task> tasks_;
   QMap<QNetworkReply*, Task> remote_tasks_;
   QMap<QString, Task> remote_spotify_tasks_;
+  QMap<QString, Task> remote_subsonic_tasks_;
   quint64 next_id_;
 
   NetworkAccessManager* network_;
 
-  bool connected_spotify_;
+  bool connected_spotify_, connected_subsonic_;
```
hatstand: One variable per line

src/internet/subsonic/subsonicservice.cpp
```diff
@@ -494,6 +495,15 @@ void SubsonicLibraryScanner::OnGetAlbumFinished(QNetworkReply* reply) {
   }
 }
 
+void SubsonicLibraryScanner::OnGetAlbumCoverFinished(QNetworkReply* reply, QString albumid) {
+  reply->deleteLater();
+  pending_requests_.remove(reply);
+
+  QByteArray image = reply->readAll();
+
+  service_->emitImageLoaded(albumid, QImage::fromData(image));
```
hatstand: Don't do this. Instead add a signal to `SubsonicLibraryScanner` and connect that signal directly to the `ImageLoaded` signal in `SubsonicService`.

davidgfnet: Yeah I knew you'd say that, but again, hack over hack.
Do you guys plan to rework the cover art links? We could create URI handlers so services can hook them, more generic and abstract. What do you think?

hatstand: Yeah, that would be much nicer.

PR #4817, opened 2015-04-20
Clementine Menu bar hide by default (user modifiable preference - ALT key press unhide)

narunlifescience:
Clementine menubar is rarely used by users as most of the options in the menu bar are layed out on the application main window itself.  this modification hides menu bar by default (user modifiable preference via preferences/appearance menu). pressing alt key unhides the menubar. (similar to Mozilla firefox menubar)

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,13 +22,13 @@ Please:
 
 Please:
  
- * Try with the latest developer build (http://builds.clementine-player.org/) to see if the bug is still present (**Attention**, those builds aren't stable so they might not work well and could sometimes break things like user settings). If if works like a charm although you see an opened issue please comment it to explain the issue was fixed
- * Check if another person didn't already open an issue
- * If there is already an opened issue, you can try to comment it to add precisions about the problem or confirm it
- * If there isn't, you can open a new issue with an explicit title and much information as possible (OS, Clementine version, what steps will reproduce the problem...)
+ * Try the latest developer build (http://builds.clementine-player.org/) to see if the bug is still present (**Attention**, those builds aren't stable so they might not work well and could sometimes break things like user settings). If it works like a charm even though you see an open issue, please comment on it and explain that the issue has been fixed
+ * Check if another person has already opened the same issue to avoid duplicates
+ * If there already is an open issue you could comment on it to add precisions about the problem or confirm it
+ * In case there isn't, you can open a new issue with an explicit title and as much information as possible (OS, Clementine version, how to reproduce the problem...)
  * Please use http://pastebin.com/ for logs/debug
  
-If there is no answers, it doesn't mean we don't care about your feature request/bug, it just mean we can't reproduce the bug or haven't time to implement it :o)
+If there are no answers, it doesn't mean we don't care about your feature request/bug. It just means we can't reproduce the bug or haven't had time to implement it :o)
 
 Compiling from source
 ---------------------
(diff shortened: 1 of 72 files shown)


amuttsch:
Hm, I don't like the trend with the hidden menu bar. With this PR, users don't have an indication that the menu bar can be shown with the `alt` key. They might think it's gone completely.

Tatsh:
No thanks. You can definitely add the feature to hide the menu bar as all KDE apps have that built-in with Ctrl+M but it should not be hidden by default. This is not a Windows Vista+ app (like Explorer or WMP).

narunlifescience:
Modifying UI is always controversial! This code can also be modified so that user can enable or disable this feature by using an entry in preferences and settings (some thing like a checkbox). @Tatsh > I am very much aware that this is a cross platform app (I use arch + xfce). This is not a KDE app either (There are tons of desktop environments out there).

Tatsh:
I can yell too.

You are correct! You know why?! Clementine is based on Amarok 1.4! Stop trying to make it into Amarok 2!

Ferroin:
I agree with narunlifescience, this should be a user modifiable preference.

narunlifescience:
modified the code as a user selectable preference.

Tatsh:
I have a sort of feeling that Linux should not necessarily get the Alt key treatment. It makes sense on Windows because Windows has this behaviour in applications like Explorer, IE, and WMP.

In KDE Ctrl+M toggles the menu in most applications. I do not really want to see detection of DE in Clementine code really but Firefox does do this. They use Alt key on KDE and F10 on GNOME. Not sure why F10. Firefox probably does a check with environment, `XDG_CURRENT_DESKTOP` (a Freedesktop standard). No real consistency anywhere (like I said most KDE applications use Ctrl+M), so I do not care that much.

One big thing: have you considered what is going to happen in OS X? Perhaps you should completely disable this feature there (`#ifndef Q_OS_OSX`).

narunlifescience:
Thank you for pointing out the mac issue. may have to find a work around; some thing similar to what you pointed out.
Don't you think that removing menu bar would give clementine a cleaner look and little more screen-space? especially in small form factor gadgets (laptops,netbooks etc)? (Well, it is debatable I guess)

Although I am not perticular about using alt key, I don’t see any better choice.  Also, I don’t want to make some weired key combination. I thought that ALT key would be easy for users as most of the users irrespective of the operating system would be familiar with ALT key usage in firefox. 

Even in KDE, this code should not make any issues (although i did not test it in KDE). both Ctrl+M and ALT should work (even menu bar hidden by Ctrl+M should unhide with alt key press and vise versa).

ArnaudBienner:
Thanks for the effort you put in this pull request.
However, I don't really like the idea.
In Firefox, options are still available through the menu in the right, so even if the menu bar isn't visible, all the options are still accessible to the users. Good point it will be an option and activated by default, otherwise I would have been definitely against this (but still, some user might turn on this option accidentally, and don't figure out how set turn if off).
Also, in Firefox, this makes sense: to make web browsing more comfortable, it makes sense to save as much space as possible. In Clementine, IMHO, the advantage of this is less obvious: Clementine is used to listen to music, not to watch some content, so saving space isn't as critical. All the contrary actually: I believe it's better to most of the options instantly visible to the user, so he can easily access them.

narunlifescience:
Hmm.. I don’t want to trouble users by adding a feature. If most of you guys feel that it should not be added then may be you are right.

PR #4821, opened 2015-04-22
Modify icon name of .protocol and .desktop too

Chocobozzz:
Sorry I forgot theses.

Unfortunately I can't test but now it's should be fine.


ArnaudBienner:
Thanks for fixing this.
Forgot to reply to your message: I would have fixed it myself otherwise (was just too busy yesterday)

Chocobozzz:
No problem :) I'll try to install a virtual machine with Ubuntu to avoid this type of problem in the future

PR #4822, opened 2015-04-22
Move queue manager 

Aceler:
Move queue manager from the separate window to docked window on the sidebar.


src/playlist/queuemanager.cpp
```diff
@@ -48,12 +48,9 @@ QueueManager::QueueManager(QWidget* parent)
   connect(ui_->move_up, SIGNAL(clicked()), SLOT(MoveUp()));
   connect(ui_->remove, SIGNAL(clicked()), SLOT(Remove()));
   connect(ui_->clear, SIGNAL(clicked()), SLOT(Clear()));
-
-  QShortcut* close = new QShortcut(QKeySequence::Close, this);
-  connect(close, SIGNAL(activated()), SLOT(close()));
 }
 
-QueueManager::~QueueManager() { delete ui_; }
+QueueManager::~QueueManager() { }
```
hatstand: This adds a memory leak

src/ui/mainwindow.h
```diff
@@ -305,6 +304,7 @@ class MainWindow : public QMainWindow, public PlatformInterface {
   DeviceView* device_view_;
   SongInfoView* song_info_view_;
   ArtistInfoView* artist_info_view_;
+  QueueManager* queue_manager_;
```
hatstand: This also adds a memory leak

src/playlist/queuemanager.cpp
Aceler: We never destroy Queuemanager anymore.

hatstand: It doesn't matter, it should still be cleaned up by MainWindow's destructor.

Aceler: Ah, OK.

src/ui/mainwindow.h
Aceler: Is this still a memory leak, if we have a correct destructor?

hatstand: Yes, `unique_ptr` was deleting this automatically in `MainWindow::~MainWindow()`

Aceler: But other tabs does not have such a destructor. There is "delete ui" in ~QueueManager(), I thought it is enough.

hatstand: Actually, you're right. It's using the QObject tree now instead.

TheUbuntuGuy:
I love this! Can I suggest that you change the label on the tab to read just "Queue" rather than "Queue Manager". As it stands, in both the small and large sidebar modes, it displays as "Queue..." or "Queue...nager".

Aceler:
Sure I can do this, but it will need a new i18n string.

othalan:
I have tried out this code and love the new location of the queue manager!  It makes the queue manager significantly easier to use.

accumulator:
I also prefer to name the tab just 'Queue'. The contents of the tab is in *technical* terms a queue manager UI, but as a user, you're looking at a queue conceptually (which you can manage).

TheUbuntuGuy:
@Aceler do you have plans to move forward on this change? If not, I'd be happy to take over the branch and bring it up to mergable condition.

Aceler:
@TheUbuntuGuy I made this patch as a proof-of-concept to decide if we really need that change. Actually, I also wanted to make queue as a separate playlist, because today the queue is tied with current playlist, but I never finished this.

So, yes, feel free to merge the code.

PR #4830, opened 2015-04-27
GIF white pixel around 

narunlifescience:
Default spinner gif image shows white pixels around the image. this is clearly visible when using a dark theme. I included a new gif image which fixes this problem.


narunlifescience:
OH! my mistake, wrong modifications.

PR #4850, opened 2015-05-12
Add stop after each song repeat mode

kiselev-dv:
It's very useful to have such mode for concerts or performances when you have ordered list of tracks and starts them as performance go.

### DIFF
--- a/src/core/player.cpp
+++ b/src/core/player.cpp
@@ -292,7 +292,7 @@ void Player::RestartOrPrevious() {
 
 void Player::Stop(bool stop_after) {
   engine_->Stop(stop_after);
-  app_->playlist_manager()->active()->set_current_row(-1);
+  //app_->playlist_manager()->active()->set_current_row(-1);
   current_item_.reset();
 }
 
--- a/src/core/player.cpp
+++ b/src/core/player.cpp
@@ -292,7 +292,6 @@ void Player::RestartOrPrevious() {
 
 void Player::Stop(bool stop_after) {
   engine_->Stop(stop_after);
-  //app_->playlist_manager()->active()->set_current_row(-1);
   current_item_.reset();
 }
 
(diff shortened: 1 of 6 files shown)


kiselev-dv:
There is a strange behavior, which I didn't get.
In player.cpp at line 295 we call:
```
app_->playlist_manager()->active()->set_current_row(-1);
```
Why? This call confuses autoselection of next row in playlist after stop_after have been called.

src/widgets/osd.cpp
```diff
@@ -314,6 +314,9 @@ void OSD::RepeatModeChanged(PlaylistSequence::RepeatMode mode) {
       case PlaylistSequence::Repeat_Playlist:
         current_mode = tr("Repeat playlist");
         break;
+      case PlaylistSequence::Repeat_OneByOne:
+        current_mode = tr("Tracks one by one");
```
hatstand: I think "Stop after every track" is more descriptive

kiselev-dv: Mb. "Stop after each track" then?

src/core/player.cpp
```diff
@@ -292,7 +292,7 @@ void Player::RestartOrPrevious() {
 
 void Player::Stop(bool stop_after) {
   engine_->Stop(stop_after);
-  app_->playlist_manager()->active()->set_current_row(-1);
+  //app_->playlist_manager()->active()->set_current_row(-1);
```
hatstand: If you're certain about this, just delete it, don't leave commented out code lying around.

kiselev-dv:
Sorry, I've used comments for testing and forget to delete them.

hatstand:
Please run `make format`

kiselev-dv:
done

hatstand:
This failed to build so I reverted it.

PR #4860, opened 2015-05-16
add google play music

BigBoot:


src/internet/playmusic/playmusicservice.cpp
```diff
@@ -0,0 +1,562 @@
...
+PlayMusicService::PlayMusicService(Application* app, InternetModel* parent)
+        : InternetService(kServiceName, app, parent, parent),
+          root_(nullptr),
+          search_(nullptr),
+          network_(new NetworkAccessManager(this)),
+          context_menu_(nullptr),
+          search_box_(new SearchBoxWidget(this)),
+          search_delay_(new QTimer(this)),
+          next_pending_search_id_(0) {
+  qsrand(QTime::currentTime().msec());
```
hatstand: Remove, this is seeded in `main()` already

src/internet/playmusic/playmusicservice.cpp
```diff
@@ -0,0 +1,562 @@
...
+
+void PlayMusicService::ShowConfig() {
+  app_->OpenSettingsDialogAtPage(SettingsDialog::Page_PlayMusic);
+}
+
+void PlayMusicService::Homepage() {
+  QDesktopServices::openUrl(QUrl("https://play.google.com/"));
+}
+
+void PlayMusicService::Login(const QString &username, const QString &password) {
```
hatstand: Not thrilled about this - presumably this access token does not work for other services?

And I guess this will fail if 2-step verification is enabled

BigBoot: Yes, it's using the (deprecated) ClientLogin, but we need it to get the device keys.
As a result, you'll need a app specific password if you're using 2-step verification.

hatstand: In general we try not to use deprecated or unofficial APIs so I don't think we can really merge this.

BigBoot: Hmm, then there's probably no chance, as google doesn't offer an official API for play music.

PR #4861, opened 2015-05-18
Fix Last.fm scrobbling after seek

### LINKED ISSUE #4836: Tracks not scrobbled after seeking
Last.fm scrobbling fails almost every single time if the track is being seeked into any direction. This isn't a simple user error regarding playing the track long enough, even backwards seeking breaks submissions. From console output it's pretty clear Clementine does not try to scrobble these tracks since there is no reply printed out either.

TheUbuntuGuy:
Fixes #4836
Last.fm defines a scrobble should be sent if:
-the track is longer than 30 seconds.
-the track has been played for at least half its duration, or for 4 minutes (whichever occurs earlier.)

Clementine has treated this as seconds from the start of the track, and if any seeking occurs, it nullifies the scrobble.
This IMO is incorrect. If I skip the first 10 seconds of a song, but listen to the rest (still meeting the time requirements),
I should still be able to scrobble the play. This change moves the scrobble point with every seek, requiring continuous playback
from any point that satisfies the time criteria.

### DIFF
--- a/src/core/player.cpp
+++ b/src/core/player.cpp
@@ -436,13 +436,9 @@ void Player::SeekTo(int seconds) {
       qBound(0ll, qint64(seconds) * kNsecPerSec, length_nanosec);
   engine_->Seek(nanosec);
 
-  // If we seek the track we don't want to submit it to last.fm
-  qLog(Info) << "Track seeked to" << nanosec << "ns - not scrobbling";
-  if (app_->playlist_manager()->active()->get_lastfm_status() ==
-      Playlist::LastFM_New) {
-    app_->playlist_manager()->active()->set_lastfm_status(
-        Playlist::LastFM_Seeked);
-  }
+  // If we seek the track we need to move the scrobble point
+  qLog(Info) << "Track seeked to" << nanosec << "ns - updating srobble point";
+  app_->playlist_manager()->active()->UpdateScrobblePoint(nanosec);
 
   emit Seeked(nanosec / 1000);
 }
--- a/src/playlist/playlist.cpp
+++ b/src/playlist/playlist.cpp
@@ -1705,14 +1705,25 @@ Song Playlist::current_item_metadata() const {
   return current_item()->Metadata();
 }
 
-void Playlist::UpdateScrobblePoint() {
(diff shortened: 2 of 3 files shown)


src/core/player.cpp
```diff
@@ -436,13 +436,9 @@ void Player::SeekTo(int seconds) {
...
 
-  // If we seek the track we don't want to submit it to last.fm
-  qLog(Info) << "Track seeked to" << nanosec << "ns - not scrobbling";
-  if (app_->playlist_manager()->active()->get_lastfm_status() ==
-      Playlist::LastFM_New) {
-    app_->playlist_manager()->active()->set_lastfm_status(
-        Playlist::LastFM_Seeked);
-  }
+  // If we seek the track we need to move the scrobble point
+  qLog(Info) << "Track seeked to" << nanosec << "ns - updating srobble point";
```
hatstand: typo

src/playlist/playlist.cpp
```diff
@@ -1705,14 +1705,25 @@ Song Playlist::current_item_metadata() const {
   return current_item()->Metadata();
 }
 
-void Playlist::UpdateScrobblePoint() {
+void Playlist::UpdateScrobblePoint(qint64 seek_point) {
```
hatstand: What units is this in?

src/playlist/playlist.cpp
```diff
@@ -1705,14 +1705,25 @@ Song Playlist::current_item_metadata() const {
...
   const qint64 length = current_item_metadata().length_nanosec();
 
-  if (length == 0) {
-    scrobble_point_ = 240ll * kNsecPerSec;  // 4 minutes
+  if (seek_point == 0) {
+    if (length == 0) {
+      scrobble_point_ = 240ll * kNsecPerSec;  // 4 minutes
```
hatstand: Can you add some more descriptive constants while you're here about what the 240 & 31 actually mean? Something like UNKNOWN_LENGTH_SCROBBLE_POINT_SECS

TheUbuntuGuy:
I've added the constants and nanosec units to the seek point.

PR #4867, opened 2015-05-23
Fix playlist not deselecting playing track when stopping

TheUbuntuGuy:
Regression from #4866 

I don't know why this line was removed. It is required to tell the playlist manager to stop the glow on the playing track once it has stopped. Without it, the track will always show as playing, and the moodbars stop working.

Adding it back in doesn't seem to break the stop after every track feature from my testing.

### DIFF
--- a/src/core/player.cpp
+++ b/src/core/player.cpp
@@ -292,6 +292,7 @@ void Player::RestartOrPrevious() {
 
 void Player::Stop(bool stop_after) {
   engine_->Stop(stop_after);
+  app_->playlist_manager()->active()->set_current_row(-1);
   current_item_.reset();
 }
 


kiselev-dv:
Hi I've removed this line. It breaks selecting of next track after current track stops.

TheUbuntuGuy:
I don't see how that breaks selection of the next track. In playlist.cpp, you changed ```stop_after_current() ``` to return ```true``` when stop after every track is enabled. When the track stops playing, ```HandleStopAfter()``` in player.cpp fires and selects the next row to play when play is pressed again.
Isn't that what you want?

Either way, you **cannot remove this call** entirely. It causes the playlist to fire several events updating the state of the player, which are used by all aspects of the UI including the play scrubber, NowPlayingWidget, etc.

kiselev-dv:
Yeah, sorry I see that I can't just remove this call.

I've removed this line because I've tried to avoid inconvenience for user: it's not obvious which song will play next time. But I've got now playing marker behavior in wrong way.

PR #4869, opened 2015-05-23
Sort discs  numerically when using Group by disc

paperbagcorner:
When grouping by disc in the library, the discs are currently sorted in lexical order, e.g  1 10 100 2 22 3, which is quite irritating for albums with a large number of discs.

This pr changes this so that the discs are sorted numerically.

### DIFF
--- a/src/library/librarymodel.cpp
+++ b/src/library/librarymodel.cpp
@@ -207,7 +207,7 @@ void LibraryModel::SongsDiscovered(const SongList& songs) {
             key = song.performer();
             break;
           case GroupBy_Disc:
-            key = song.disc();
+            key = QString::number(song.disc());
             break;
           case GroupBy_Grouping:
             key = song.grouping();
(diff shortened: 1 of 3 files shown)


src/library/librarymodel.cpp
```diff
@@ -1098,14 +1108,9 @@ QString LibraryModel::SortTextForArtist(QString artist) {
...
-  QString str = QString::number(year);
-  return QString("0").repeated(qMax(0, 4 - str.length())) + str;
-}
-
-QString LibraryModel::SortTextForBitrate(int bitrate) {
-  QString str = QString::number(bitrate);
-  return QString("0").repeated(qMax(0, 3 - str.length())) + str;
+QString LibraryModel::SortTextForNumber(int number) {
+  QString str = QString("%1").arg(number, 4, 10, QChar('0'));
+  return str;
```
hatstand: Just `return` directly

paperbagcorner: Oops, I missed that. It is fixed now.

PR #4896, opened 2015-06-11
[Qt5] Fix release compilation and set cmake min version to 2.8.11

Chocobozzz:
I don't know why but when we set the *QT_NO_DEBUG_OUTPUT* flag in release compilation mode `QDebug` (returned by `qLog()` ) breaks the compilation. So we need to return a `QNoDebug`.

Thus I modified the `qLog`: 

* Don't print info and debug messages when *QT_NO_DEBUG_OUTPUT* is set
* Don't print warning messages when *QT_NO_WARNING_OUTPUT* is set

I'm not an expert in macros so tell me if you think there is a better solution.

I added *QT_NO_WARNING_OUTPUT* flag in release mode.


ext/libclementine-common/core/logging.h
```diff
@@ -27,10 +27,44 @@
...
-#define qLog(level)                                                        \
+#define qLog(level) \
+  qLog##level()
+
+#define qCreateLogger(level) \
   logging::CreateLogger(logging::Level_##level,                            \
                         logging::ParsePrettyFunction(__PRETTY_FUNCTION__), \
                         __LINE__)
+
+#define qLogFatal() \
```
hatstand: These could probably be functions, making things a bit easier to debug.

ext/libclementine-common/core/logging.h
```diff
@@ -27,10 +27,44 @@
...
+#define qLogError() \
+  qCreateLogger(Error)
+
+#ifdef QT_NO_WARNING_OUTPUT
+#define qLogWarning() \
+  while (false) QNoDebug()
+#else
+#define qLogWarning() \
+  qCreateLogger(Warning)
+#endif
```
hatstand: Put comments after the `#endifs` saying what the original `#ifdef` was.

Chocobozzz:
Yes it looks better with this way.

hatstand:
Don't bother marking them as `inline` though

Chocobozzz:
Why not ?

hatstand:
Just let the compiler figure it out

Chocobozzz:
Okay, forgot the compiler was so intelligent :p

It's done, I modified the last commit.

PR #4903, opened 2015-06-14
Add track intro mode

TheUbuntuGuy:
This repeat mode plays the beginning of each track, good for quickly getting a feel for a new album for example. An inferior music player I used to use had this and I found it quite useful.


TheUbuntuGuy:
It seems ```make format``` broke outgoingdatacreator.cpp:384. Oops.

PR #4905, opened 2015-06-15
Add track intro mode

TheUbuntuGuy:
Fixed compile issue on some compilers from original PR #4903


Edistaeonas:
Could we have a setting to choose the length in seconds for the duration of the intro playback, for example if I want each track to play during exactly 30 seconds. The current duration seems a bit random and rather long (2 or 3 minutes). Thank you.

PR #4924, opened 2015-06-26
Update Boom and Turbine analyzers

TheUbuntuGuy:
Boom and Turbine remain small fixed-size analyzers with a few issues with regards to colours.
This update attempts to remedy these issues.
Changes:
-Both analyzers now resize to fit the size of their container, and scale to the lower frequencies if in a small window
-The colour now does not change as the main window gains/loses focus
-The peak bars now use a colour which is actually visible (I didn't even know they existed before!)

### DIFF
--- a/src/analyzers/boomanalyzer.h
+++ b/src/analyzers/boomanalyzer.h
@@ -1,7 +1,7 @@
 /* This file is part of Clementine.
    Copyright 2004, Max Howell <max.howell@methylblue.com>
    Copyright 2009-2010, David Sansome <davidsansome@gmail.com>
-   Copyright 2014, Mark Furneaux <mark@romaco.ca>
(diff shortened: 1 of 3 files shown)


src/analyzers/boomanalyzer.h
```diff
@@ -49,9 +44,15 @@ class BoomAnalyzer : public Analyzer::Base {
 
  protected:
   void resizeEvent(QResizeEvent* e);
+  void paletteChange(const QPalette&);
+
+  static const uint COLUMN_WIDTH;
```
hatstand: Can you update the style of all the constants/variables while you're here?
i.e. `COLUMN_WIDTH` -> `kColumnWidth`
`m_bands` -> `bands_`

TheUbuntuGuy:
Refactoring done. I have another PR coming soon, and I'll update the rest of the analyzer source in that PR.

PR #4929, opened 2015-07-01
Add "Psychedelic Colour" mode to all analyzers

TheUbuntuGuy:
(Well, except Nyanalyzer and Rainbow dash because they are already colourful enough.)

I have added functionality for any 2D analyzer to change any part of its colour palatte with the frequency content of the music, in the same way that Moodbars do.
I find this gives the analyzer a sort of "third dimension".
This is built into Analyzer::Base, so all analyzers can use it and override it as they please. I have thus added support for Block, Boom, Turbine, Sonogram, and Bar, however Boom and Block seem to look the best in my opinion. This is of course all optional and is toggled by a checkbox in the context menu for the analyzer, disabled by default. I have not been able to measure any increase in CPU activity with this enabled, even at 60fps. I am by no means an expert on colour theory, so there is probably someone out there who can tweak the output.

Here are some stills of what the various analyzers can look like:
psychedelicanalyzers

I've also refactored the remaining source to the common variable/constant format.

### DIFF
--- a/src/analyzers/analyzercontainer.cpp
+++ b/src/analyzers/analyzercontainer.cpp
@@ -3,7 +3,7 @@
    Copyright 2010, 2014, John Maguire <john.maguire@gmail.com>
    Copyright 2011-2012, Arnaud Bienner <arnaud.bienner@gmail.com>
    Copyright 2013, Vasily Fomin <vasili.fomin@gmail.com>
-   Copyright 2014, Mark Furneaux <mark@romaco.ca>
+   Copyright 2014-2015, Mark Furneaux <mark@furneaux.ca>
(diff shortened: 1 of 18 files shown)


src/analyzers/analyzercontainer.cpp
```diff
@@ -145,6 +151,12 @@ void AnalyzerContainer::DisableAnalyzer() {
   Save();
 }
 
+void AnalyzerContainer::TogglePsychedelicColors() {
+  psychedelic_colors_on_ ^= true;
```
hatstand: I think `psychedelic_colors_on_ = !psychedelic_colors_on_;` is more readable.

TheUbuntuGuy:
Done. I think I'm too used to doing this in C. :)

PR #4932, opened 2015-07-02
Make seeking with keys behaviour consistent.

eduardosm:
Seeking using arrow keys behaved differently depending on which widget received the keypress event.

It may be a good idea to define a constant with the number of seconds to seek instead of using a hardcoded 1.

### DIFF
--- a/src/core/player.cpp
+++ b/src/core/player.cpp
@@ -52,6 +52,7 @@
 using std::shared_ptr;
 
 const char* Player::kSettingsGroup = "Player";
+const int kSeekSec = 10;
 
 Player::Player(Application* app, QObject* parent)
     : PlayerInterface(parent),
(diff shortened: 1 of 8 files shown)


hatstand:
Yes, please make it a constant.

eduardosm:
It finally used Player::SeekForward() and Player::SeekBackward() to handle it. I'm thinking of adding an option to configure the number of seconds to seek.

src/core/player.cpp
```diff
@@ -52,6 +52,7 @@
 using std::shared_ptr;
 
 const char* Player::kSettingsGroup = "Player";
+const int kSeekSec = 10;
```
hatstand: Put this in an anonymous namespace

eduardosm:
Done. I also renamed the constant to kSeekStepSec, which I found more appropriate.

Issue #4354, opened 2014-05-18
Send album artist to Last.fm


diogocp:
Not a duplicate of #2901. This is about sending album artist as album artist, not album artist as artist.

Patch: https://gist.github.com/diogocp/db723bb55e269dd156ba

liftarn:
Does not appear to work in 1.3.1

PR #4934, opened 2015-07-02
Send album artist to Last.fm

diogocp:
This was reverted in 1f97406171eb82027fd2bec73514e497f3df7eee, but like I commented there, MutableTrack::setAlbumArtist is supported by the latest version of liblastfm, as well as all versions since at least 1.0.0, released Jun 11, 2012.

My guess is that @hatstand was using an old version of liblastfm.

Issue: #4354
Original PR: #4363

### DIFF
--- a/src/core/song.cpp
+++ b/src/core/song.cpp
@@ -998,6 +998,7 @@ void Song::ToLastFM(lastfm::Track* track, bool prefer_album_artist) const {
   } else {
     mtrack.setArtist(d->artist_);
   }
+  mtrack.setAlbumArtist(d->albumartist_);
   mtrack.setAlbum(d->album_);
   mtrack.setTitle(d->title_);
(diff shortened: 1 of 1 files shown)


ArnaudBienner:
This breaks the build

diogocp:
@ArnaudBienner on what version of liblastfm?

hatstand:
0.4.0~really0.3.3-0ubuntu1 on Ubuntu precise

diogocp:
@hatstand liblastfm's git history doesn't go that far back, but that's probably too old.

Is it really necessary to keep supporting such old versions of liblastfm?

hatstand:
It's nice to be able to support Ubuntu LTS versions.

diogocp:
@hatstand in that case, would it be ok to use a preprocessor conditional to only compile that line on liblastfm >= 1.0.0?

hatstand:
Sure

PR #4935, opened 2015-07-03
Add mono output support in the engine

TheUbuntuGuy:
Mono playback was removed from the pipeline in the transition to gstreamer 1.0.


ArnaudBienner:
IIRC @davidsansome reverts this when migrating to gstreamer 1.0 because it breaks Spotify playback.
I just tested the latest build on OS X, and it just completely breaks the playback for all types of tracks (local or spotify) when the option is activated.

TheUbuntuGuy:
I only have Linux machines and it works on them. I tested the win32 build at work and it is indeed broken across all playback sources, strange. I'll look into this when I can.

PR #4940, opened 2015-07-04
Fix mono playback on certain configurations

TheUbuntuGuy:
Relates to #4935.

Gstreamer was failing to link the pipeline if a 32bit pipe could not be established.
We should just let gst autonegotiate the bit depth of the pipeline, which it does with mono disabled anyways.
I've tested with a Windows system that failed before, and it now works correctly with this change.

### DIFF
--- a/src/engines/gstenginepipeline.cpp
+++ b/src/engines/gstenginepipeline.cpp
@@ -391,11 +391,10 @@ bool GstEnginePipeline::Init() {
 
   // add caps for mono, but only if requested
   if (mono_playback_) {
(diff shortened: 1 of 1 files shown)


ArnaudBienner:
Still doesn't work for me on Mac.
@hatstand do you also have the same problem or is it just me?

hatstand:
Doesn't work for me either on mac

TheUbuntuGuy:
I don't own or have access to a Mac, so I don't think I can pursue this any further. I've tested on 2 different Linux, and 2 different Windows machines, and I've seen no issues. I don't know what Macs do differently.

Issue #4748, opened 2015-02-14
Player randomly stuck in transition to next song


cschreib:
Running Linux, Clementine 1.2.3. This issue happens randomly, but quite frequently.

When a new track plays, sometimes the "progress bar" gets stuck at 0:01, and no sound is produced. I tried waiting for more than 20 minutes, and gave up. In this situation, if I press "pause", Clementine obeys. But if then I try to play the song again, by clicking the "play" button, Clementine refuses and the song remains paused. The only way out is to play another song (for example by clicking the "next" button), which starts immediately.

I found nothing particular in the debug output. The last entry is about the Moodbar, which says that the data has been successfully generated (that was the first time I played that song). This was printed a few instants after what was supposed to be the beginning of the next song.

So far, it seems that the issue only happens for a particular artist. I see no reason why, apart from the fact that the file name (and MP3 tags) contains accentuated characters. Note that playing one of these songs does not always trigger the problem, it still happens randomly.

cschreib:
Actually this is not an issue with accentuated characters. The problem just happened on a song without any such character, neither in the file name nor in the tags.

OskarW85:
I have exactly same problem on Windows with Clementine 1.2.3-978-g9dbd582. But it happens for a long time for me (at least 2 months). I am not using moodbar. Player just refuses to advance to next song. Sometimes after longer period of being stuck (couple of minutes) it begins playing next song by itself.

cschreib:
Good to know this is not Linux related. Does this happen for any song, or could you spot a pattern? It seems from my side that this only happens for songs that have never been played before, but I'm not 100% sure.

Ferroin:
I don't think that it's directly related to #4613, as I have a relatively small (only about 300 songs) music library and still see this myself.

I've noticed that on both Linux and Windows, it tends to happen more frequently when there are a lot of other things running.  I've also found that, sometimes, if you stop playback, and then restart, it will still get stuck trying to play the next track, although other times this seems to fix the issue.  I've also found that adjusting priorities (I usually run audio software at real-time priorities on Linux) doesn't seem to help at all, and that saving stats to the library makes it worse.

On Linux (and other unices), we might get better results if we do some magic with fadvise/readahead to tell the VFS layer to drop the cached data from the last track and preload the next track into cache.  I'd actually been considering writing a patch to do this for some time now just cause it should improve performance, but hadn't had the time to even begin looking at the code.

OskarW85:
[...]

Issue #4613, opened 2014-11-13
Works bad with big music library on windows


Yanpas:
Every 10th-15th next track loads for about minute or two. When I had about 800 tracks in my library everything was OK. Now I have 1500 and have this problem. When I listen music after some track playng pause, the timeline showa 0:00 and track doesn't play for a minute or two. If I press Previous track and then next track - playng continues.

yafp:
I cant reproduce that on Windows with a bigger library.

@Yanpas 
What version of Clementine are you using and does it still happen?

Yanpas:
1.2.3 
Yes, it still happens. Try suspend and resume your computer. 
I listen to music and after random track Clementine doesn't play for 10-20 minutes. And then automatically begins to play

Issue #4189, opened 2014-02-07
Playback sometimes paused when track ends


nuklea:
Playback paused at 00:01 on next track. After some time playback resumed.

I use clementine from git.

``` bash
$ LC_ALL=C yaourt -Qi clementine-git | grep Version
Version        : 1.2.1.305.gacfc7e6-1

$ uname -a
Linux footbaby 3.12.9-2-ARCH #1 SMP PREEMPT Fri Jan 31 10:22:54 CET 2014 x86_64 GNU/Linux
```

fortysixandtwo:
I can confirm what pernatiy reported:
Transition between songs with 44,1 andd 48kHz sample rate causes player to be stuck at 00:01.
Using Clementine 1.2.3 on Win7 64bit (but I think I have the same issue in Linux aswell).

Ferroin:
I'd been doing some experimentation myself, and had come across the same thing, with the caveat that it happened for any case of transitioning between sample rates (I tested with a mix of 22.05k, 44.1k, 48k, and 96k sample rates).

Ferroin:
Also, it happens more consistenly for me with larger sample sizes (eg, more often for 24-bit than 16-bit), and more frequently with FLAC than MP3 or Vorbis.

zluca:
Clementine 1.3.1. Bug is still here.

PR #4945, opened 2015-07-10
Add sample rate selection

TheUbuntuGuy:
This is meant to be a feature, but it also triages #4189, #4748, #4613?

This change provides the ability to set a fixed pipeline sample rate as an alternate to automatically negotiating it (the default). This can be useful on systems with sound cards that work at a fixed rate, as well as it can triage issues where changing tracks hangs due to a problem with gstreamer's caps negotiation.

I've noticed myself that when transitioning between tracks with different sample rates on Windows, the pipeline will always hang on starting the next song. This seems to be gstreamer not negotiating the downstream caps for sample rate. In fact if you disable "exclusive mode" on your sound card in Windows, the pipeline cannot start at all without a set sample rate. I don't want to say this is *the* solution, but it at least stops all the hanging issues. Hopefully someone with more gstreamer experience can find the underlying issue - that is, if this is indeed an issue in Clementine, and not a bug in gstreamer.

### DIFF
--- a/src/engines/gstengine.cpp
+++ b/src/engines/gstengine.cpp
@@ -103,6 +103,7 @@ GstEngine::GstEngine(TaskManager* task_manager)
       buffer_duration_nanosec_(1 * kNsecPerSec),  // 1s
       buffer_min_fill_(33),
       mono_playback_(false),
+      sample_rate_(kAutoSampleRate),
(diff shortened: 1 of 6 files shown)


hatstand:
Cue people setting it to 192kHz because it "sounds better".

PR #4951, opened 2015-07-14
Update changelog

Chocobozzz:
Once again, tell me if you don't agree with some things.

There are many things in the changelog, maybe we could prepare a new release when we'll have time?

I know it's a lot of work, so do no hesitate to tell us if you need something.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -2,9 +2,11 @@ Next release:
   Major features:
     * Vk.com support
     * Seafile support
+    * Amazon Cloud Drive support
     * Add Ampache compatibility (through Subsonic service)
     * Add new analyzer "Rainbow Dash"
     * Answer to the ultimate question of life, the universe and everything
+    * Add "Psychedelic Colour" mode to all analyzers
 
   Other features:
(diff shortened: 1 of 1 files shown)


Changelog
```diff
@@ -2,9 +2,12 @@ Next release:
   Major features:
     * Vk.com support
     * Seafile support
+    * Amazon Cloud Drive support
     * Add Ampache compatibility (through Subsonic service)
     * Add new analyzer "Rainbow Dash"
     * Answer to the ultimate question of life, the universe and everything
+    * Add "Psychedelic Colour" mode to all analyzers
+    * Add mono output support in the engine
```
hatstand: This was already there and just temporarily removed in the gstreamer 1.0 transition IIRC

Chocobozzz:
Oup's missed that. I modified the commit.

TheUbuntuGuy:
By a crude analysis, I think this makes the next release the largest by sheer number of changes ever (137). The largest so far was 1.0 (104).

PR #4959, opened 2015-07-22
Cache and restore further mp3-file metadata-frames instead of cruel deletion

M-Bab:
Softened the metadata-editor behavior to not deliberately delete additional, valid frames which are stored in mp3-files.

This fixes also a few bug reports because the Clementine metadata-editor could actually lead to losing valid data from an mp3-file(!) STILL the metadata-editor does not show further frames or allows to edit them but at least it also does not remove them anymore.

Stashed commit of the following:

01c71c4c67be44dd6889498abc323a6d2d20dd99
Added description "Clementine editor" to new created lyrics frames.

70dfa0d1f177523ff897985ae022979cd052a6ed
Updated the SetTextFrame method as well:
Existing frames are cached and only the first frame (which is actually shown in the metadata editor) will be updated.
Prevents the cruel deletion of all other existing frames in a well defined mp3-file.

dfddf76eda417442c8c168eb8868038b0675cb10
Updated behavior of metadata-editor concerning lyrics (might follow for other id3v2-tag properties):
Existing further frames in the tags are not deleted but cached so only the true changes by the metadata-editor are applied on the edited frame.
Further frames in the metadata are preserved.

### DIFF
--- a/ext/libclementine-tagreader/tagreader.cpp
+++ b/ext/libclementine-tagreader/tagreader.cpp
@@ -25,6 +25,7 @@
 #include <QNetworkAccessManager>
 #include <QTextCodec>
 #include <QUrl>
+#include <QVector>
 
 #include <aifffile.h>
 #include <asffile.h>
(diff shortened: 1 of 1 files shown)


ext/libclementine-tagreader/tagreader.cpp
```diff
@@ -893,36 +894,63 @@ void TagReader::SetTextFrame(const char* id, const QString& value,
...
   }
 
-  // Create and add a new frame
-  TagLib::ID3v2::TextIdentificationFrame* frame =
-      new TagLib::ID3v2::TextIdentificationFrame(id_vector,
-                                                 TagLib::String::UTF8);
-  frame->setText(StdStringToTaglibString(value));
-  tag->addFrame(frame);
+  // If no frames stored create empty frame
+  if (frames_buffer.size() == 0) {
```
hatstand: `.isEmpty()`

ext/libclementine-tagreader/tagreader.cpp
```diff
@@ -893,36 +894,63 @@ void TagReader::SetTextFrame(const char* id, const QString& value,
...
+    TagLib::ID3v2::TextIdentificationFrame* frame =
```
hatstand: This leaks. Do this instead:
```c++
TagLib::ID3v2::TextIdentificationFrame(id_vector, TagLib::String::UTF8);
frames_buffer.push_back(frame.render());
```

ext/libclementine-tagreader/tagreader.cpp
```diff
@@ -893,36 +894,63 @@ void TagReader::SetTextFrame(const char* id, const QString& value,
...
+    frames_buffer.push_back(frame->render());
+  }
+
+  // Update and add the frames
+  for (int lyrics_index = 0; lyrics_index < frames_buffer.size();
+       lyrics_index++) {
+    TagLib::ID3v2::TextIdentificationFrame* frame =
+        new TagLib::ID3v2::TextIdentificationFrame(
+            frames_buffer.at(lyrics_index));
+    if (lyrics_index == 0) frame->setText(StdStringToTaglibString(value));
```
hatstand: Use `{}`

ext/libclementine-tagreader/tagreader.cpp
```diff
@@ -893,36 +894,63 @@ void TagReader::SetTextFrame(const char* id, const QString& value,
...
+    tag->addFrame(frame);
```
hatstand: Comment that this takes ownership of `frame`

ext/libclementine-tagreader/tagreader.cpp
```diff
@@ -893,36 +894,63 @@ void TagReader::SetTextFrame(const char* id, const QString& value,
...
     tag->removeFrame(tag->frameListMap()[id_vector].front());
   }
 
-  // Create and add a new frame
-  TagLib::ID3v2::UnsynchronizedLyricsFrame* frame =
-      new TagLib::ID3v2::UnsynchronizedLyricsFrame(TagLib::String::UTF8);
-  frame->setText(StdStringToTaglibString(value));
-  tag->addFrame(frame);
+  // If no frames stored create empty frame
+  if (frames_buffer.size() == 0) {
```
hatstand: `isEmpty()`

ext/libclementine-tagreader/tagreader.cpp
```diff
@@ -893,36 +894,63 @@ void TagReader::SetTextFrame(const char* id, const QString& value,
...
+    TagLib::ID3v2::UnsynchronizedLyricsFrame* frame =
```
hatstand: This is leaked

M-Bab:
Ok. Thanks for your review. I fixed the leaks and the cosmetic comments.

Issue #2317, opened 2013-12-10
Sorting by disc should do a secondary sort on track


Clementine-Issue-Importer:
_From [robl...@gmail.com](https://code.google.com/u/112867229612031305652/) on October 20, 2011 09:55:00_

1. get a 2+ disc album and load it into your playlist area
2. scramble it by sorting on title or something 
3. sort by disc#

At this point observe that your playlist is indeed sorted by disc#, but the tracks are still out of order.  This is undesirable in any circumstance.

4. Sort by track
5. Sort by disc# once more

Now, your tracks are in proper order, but this is quite an awkward workaround. Please see my related FR about the album field sorting doing this as well.

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=2317_

Issue #2315, opened 2013-12-10
Sorting by album should have a secondary and tertiary sort for disk and track


Clementine-Issue-Importer:
_From [robl...@gmail.com](https://code.google.com/u/112867229612031305652/) on October 20, 2011 09:36:56_

When you sort by album, if all of the ID3 tags for a set of tracks are the same for the "Album", it would be nice if the software would do a secondary sort on the disc #, and a tertiary sort on the track.  

From what I can tell, it SEEMS like this can be accomplished by sorting by disc #, but it would be nice to see it work for album too... Amarok 1.4 did this.

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=2315_

PR #4972, opened 2015-08-01
Playlist sort by album considers disc and track numbers

santigl:
When sorting the playlist by album, it also sorts by disc number and track number. That way multi-disk albums get their proper playing order.

This was requested in isues #2315 and #2317.


hatstand:
Please run `make format`

santigl:
Sorry about that, hatstand. There I've done it and pushed the result.

Issue #4877, opened 2015-05-30
'Show in file browser' stopped working with special characters


electric-indigo:
Few days ago I noticed the 'Show in file browser' feature had stopped working, showing 'File not found' error instead. It still works for paths which do not contain spaces, braces, etc. The file path in the error message uses URL-style encoding for all non-standard characters, including spaces.
I use the 1.2.3 version (32bit) from the official Arch Linux repositories.
The file manager (Thunar) does not seem to be the issue, as I tested several previous versions.

electric-indigo:
I've done some further investigation and it turns out that this problem occurs only when the said characters appear in directories’ names. The file name itself does not trigger the error. I also stumbled upon possibly related issue when trying to play files from the library (on PC) with the remote for Android. The only difference is the error message’s formatting — specifically, spaces aren't encoded as %20, while square brackets (for instance) are still encoded as %5B and %5D.

Rika-chan:
Can confirm, 1.2.3 from debian Stretch repos.

Same symptoms being experienced.

Rika-chan:
Compiled from the git, still doesn't appear to work.

21:04:22.101 DEBUG Utilities:406                    "/home/rika-chan/Bernkastel/01 11h30.mp3" 
21:04:23.577 DEBUG Utilities:406                    "/home/rika-chan/Bernkastel/Music/[Audio-4U] Foxtail-Grass Studio 1.0 (flac)/[2015.05.10] ゆきどけアゼリア/09. Foxtail-Grass Studio — ハルノウタ.flac"

The first one opened fine, the second did not.

electric-indigo:
This is probably due to Qt update; I encountered similar problems in few other Qt4 applications.

Jojo144:
It doesn't seem to solve the problem for me.
I have exactly the same symptoms with Clementine 1.2.3-1364-gd51a1ad
(from http://builds.clementine-player.org/ubuntu-trusty/ on Ubuntu Mate 15.04).

Rika-chan:
an update in either xdg-open or exo-open appeared to fix it for me, I'm on 1.1.0~rc3+git20150922-1 for xdg-utils, and 0.10.7-1 for exo-open.

Debian Stretch and clementine 1.2.3 still.

Jojo144:
Indeed!
It works fine with [xdg-open 1.1.0~rc3](https://launchpad.net/ubuntu/+source/xdg-utils).

Thanks.

[...]

PR #4984, opened 2015-08-07
Workaround for malfunctioning "Show in file browser" feature in Linux

electric-indigo:
(issue #4877).

### DIFF
--- a/src/core/utilities.cpp
+++ b/src/core/utilities.cpp
@@ -38,6 +38,7 @@
 #include <QIODevice>
 #include <QMetaEnum>
 #include <QMouseEvent>
+#include <QProcess>
 #include <QStringList>
 #include <QTcpServer>
 #include <QtDebug>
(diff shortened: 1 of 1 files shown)


src/core/utilities.cpp
```diff
@@ -411,7 +410,9 @@ void OpenInFileBrowser(const QList<QUrl>& urls) {
 #elif defined(Q_OS_WIN32)
     ShowFileInExplorer(path);
 #else
-    QDesktopServices::openUrl(QUrl::fromLocalFile(directory));
+    QStringList args;
+    args << directory;
+    QProcess::startDetached("xdg-open", args);
```
hatstand: This won't work on mac

electric-indigo: Isn't mac handled in the section with Q_OS_DARWIN?

hatstand: Oh right, the diff just skips that part out

Issue #4741, opened 2015-02-07
Make mood files hidden in NTFS


Yanpas:
Win explorer shows files starting with dot by default. Add to mood files attribute Hidden. System command: ATTRIB +H

santigl:
How about adding these lines to function `MoodbarLoader::RequestFinished` (defined in **src/moodbar/moodbarloader.cpp**) after the write to `mood_filename`?

```C++
#ifdef Q_OS_WIN32
        if (!SetFileAttributes((LPCTSTR)mood_filename.utf16(),
                               FILE_ATTRIBUTE_HIDDEN)) {
          qLog(Warning) << "Error changing hidden attribute for file"
                        << mood_filename;
        }
#endif
```
I can't test it on Windows, but if someone confirms it works I can send the patch.

Chocobozzz:
@santigl You can make a pull request :)

Yanpas:
I can test it on windows

narunlifescience:
@hatstand I think this issue should be closed..

PR #4986, opened 2015-08-08
 Save mood file as hidden on Windows

santigl:
In response to issue #4741.

This haven't been tested, but it may work.


src/moodbar/moodbarloader.cpp
```diff
@@ -167,6 +167,15 @@ void MoodbarLoader::RequestFinished(MoodbarPipeline* request, const QUrl& url) {
       QFile mood_file(mood_filename);
       if (mood_file.open(QIODevice::WriteOnly)) {
         mood_file.write(request->data());
+
+#ifdef Q_OS_WIN32
+        if (!SetFileAttributes((LPCTSTR)mood_filename.utf16(),
```
hatstand: You'll need to `#include <windows.h>` behind `Q_OS_WIN32` as well

santigl: Done. I mistakenly assumed it would already be included.

PR #5009, opened 2015-08-22
Add custom hosts scrobbling support with libscrobbler

Chocobozzz:
Liblastfm doesn't seem to want to merge my PR which add the ability to customize the host and thus use libre.fm. I emailed them but nothing changed. So I forked the library and renamed lastfm to scrobbler (https://github.com/Chocobozzz/libscrobbler). I added this library to the AUR. This is now the choice of packagers/users to compile with liblastfm or libscrobbler. 

This PR adds support of libscrobbler. If Clementine compiles with liblastfm it doesn't allow to modify the scrobbler host. If it compiles with libscrobbler the user can edit the host (by default it always use lastfm host).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,9 +81,18 @@ if (WIN32)
   find_library(QTSPARKLE_LIBRARIES qtsparkle)
 endif (WIN32)
 
-find_library(LASTFM_LIBRARIES lastfm)
-find_path(LASTFM_INCLUDE_DIRS lastfm/ws.h)
-find_path(LASTFM1_INCLUDE_DIRS lastfm/Track.h)
+find_library(SCROBBLER_LIBRARIES scrobbler)
+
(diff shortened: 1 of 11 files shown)


hatstand:
This is a pretty messy and intrusive change for something few people care about.

Chocobozzz:
Intrusive? This just adds support for a custom library, to let people choose what they want (for people using the custom library it just adds a checkbox for others nothing change). But yes, it complicates the code with a lot of defines. As you want. But this is the fourth most commented issue (https://github.com/clementine-player/Clementine/issues/1).

hatstand:
Itt makes the code much more difficult to read. I'm not against it in principle but you'll need to find a more modular way of doing this.

Chocobozzz:
Okay, I'll try to think about that when I'll have time. :)

hatstand:
Thanks for making the effort :-)

Reopen this when you want me to take another look.

PR #5015, opened 2015-08-31
clean up library view

narunlifescience:
library view sorts the library and display the contents in alphabetical order (bold letter/number followed by a line) . the line cannot be themed as it is hard-coded "QPen line_pen(opt.palette.color(QPalette::Dark))". Here i try to remove the line, align the letter to right and give the letter a highlight using QRect. this way it looks better and is completely themable ...

snapshot2
snapshot3

check the screenshots.. (should work for all platforms)

### DIFF
--- a/src/library/libraryview.cpp
+++ b/src/library/libraryview.cpp
@@ -62,6 +62,8 @@ void LibraryItemDelegate::paint(QPainter* painter,
     QString text(index.data().toString());
 
     painter->save();
+    painter->setRenderHint(QPainter::Antialiasing);
+    painter->setRenderHint(QPainter::HighQualityAntialiasing);
 
     QRect text_rect(opt.rect);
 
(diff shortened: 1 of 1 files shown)


narunlifescience:
snapshot2
snapshot3

here is a comparison..

hatstand:
Can you run `make format` and check that this works in RTL (switch the language to one of ar/fa/he).

Chocobozzz:
I don't find it very beautiful :/ 
clementine_cleaning_library_view

narunlifescience:
@Chocobozzz it looks like an antialiasing issue.. I will try to fix it. @hatstand I will try that too..

narunlifescience:
@Chocobozzz could you please check the modified code.. now it seems to be OK (checked in xfce).
@hatstand it seems to work fine with RTLs. check the screenshots below
xfce (ar)>
screenshot_2015-09-01_11-57-59
screenshot_2015-09-01_11-58-38

KDE plasma (fa)>
snapshot4
snapshot5

I hope this is what you asked me to check..

Chocobozzz:
Looks better but I'm not a fan of the the filling. Of course it's just a personnal opinion and I'm not very good in design.

hatstand:
Formatting is still wrong, please use `make format`

narunlifescience:
@hatstand I am unable to understand your comment. please explain with  little more details..

hatstand:
Your code is incorrectly formatted. Run the command `make format` to fix it.

narunlifescience:
@hatstand could you please point out the formatting errors ? it seems fine to me..

hatstand:
Please revert the `README.md` changes

narunlifescience:
I am sorry, that was my mistake..

hatstand:
This is kind of a mess of commits now, can you please squash them into one?

narunlifescience:
@hatstand Done!

hatstand:
Thanks

ArnaudBienner:
I agree with @Chocobozzz: I preferred the old one too :(
And the line is visible, even if it is not adapted to the theme.
Can't we just make the line themable? Or achieve a similar effect without redesigning this completely?
Things that makes me prefer the old ones:
- Letters were "nicer" (here on my mac they don't fit the the circle they are in)
screen shot 2015-09-14 at 19 21 14
but regarding your screenshots, this seems to be OS X specific
- No line behind the letter: this was cleaner IMO
- Letter on the right makes this difficult to see/understand. Looks better indeed for RTL languages, but then we should put them on the left for LTR languages.

narunlifescience:
@ArnaudBienner, @Chocobozzz  what would be a better solution for this ? from what i can see, it should 'nt be hard to fix this in OSX or use #ifdef Q_OS_DARWIN to force OSX to use old style. but that is not an ideal solution.
what about adding an option to configure this in appearance settings? some thing like choosing the alignment (left/right), letter/number style(previous clementine style/filling highlight style), draw line (yes or no). this way, it should be more modular and we could add more configurable options in future. what is your thought on this ?
OR should we revert back to previous style? the line can be easily themed..

TheUbuntuGuy:
I agree with @ArnaudBienner. I really don't like the way this is currently implemented. At a minimum they must be on the left for LTR languages as they look way out of place on the right.

I personally liked the line, and I'm not sure why we should remove it just because it didn't theme well in the old implementation.

I strongly disagree about adding an option for styling this. Adding options like that shows users how we as developers can't decide on what our UI should look and feel like, and it clutters up the settings. We should appeal to the majority of users, who won't care about settings like that. I agree that Clementine should be customizable to those who want to tinker (like myself), but there needs to be a point where that stops.

I'm not trying to come off as rude, this is just my opinion.

narunlifescience:
what you guys think about something like this ?
mod

ArnaudBienner:
I agree with @Chocobozzz 
IMHO we should revert to the old style, and make the line themable to fix the initial issue.

narunlifescience:
Hmm.. then let's revert back ..

ArnaudBienner:
I've reverted your change.
But thanks anyway for the effort you put on this, and on all the other patches you did recently :)
If you want, feel free to submit a pull request to make the line themable (or I might try to have a look at this later)

narunlifescience:
I already have a pull request for the volume slider glow animation (https://github.com/clementine-player/Clementine/pull/5042). once that get processed, i will work on it..

PR #5030, opened 2015-09-10
volume text color palette

narunlifescience:
This modification makes the volume percentage text color match general text color.
untitled


TheUbuntuGuy:
Using the default theme on Ubuntu, the text and outline colour are now really light and have very little contrast to the light background. It is much harder to read now IMO.

Old:
screenshot from 2015-09-19 173654

New:
screenshot from 2015-09-19 17 34 04

narunlifescience:
it is because volume % text use disabled text color from system theme.. yes, it is bit light in this case but hard coding the volume% color is even worse...

PR #5035, opened 2015-09-14
Volume pixmap render on the fly using system theme

narunlifescience:
This modification renders volume pixmap image using system forground theme color. 
untitled


src/widgets/sliderwidget.h
```diff
@@ -98,6 +98,7 @@ class VolumeSlider : public Slider {
 
  public:
   VolumeSlider(QWidget* parent, uint max = 0);
+  QPixmap volumePixmapDraw();
```
hatstand: `QPixmap DrawVolumePixmap() const;`

hatstand:
Run `make format` too

src/widgets/sliderwidget.h
narunlifescience: Done!

narunlifescience:
@hatstand Done!

src/widgets/sliderwidget.cpp
```diff
@@ -301,6 +301,8 @@ void Amarok::VolumeSlider::paintEvent(QPaintEvent*) {
   const int padding = 7;
   const int offset = int(double((width() - 2 * padding) * value()) / maximum());
 
+  m_pixmapInset = volumePixmapDraw();
```
ArnaudBienner: Do you really need to recompute it for every paint event?
m_pixmapInset is already set in the constructor, and I don't expect it to change. If I'm right, the only reason for this to happen is in case the theme changes, am I right?
If so, it would be better to avoid calling the function everytime (e.g. detect if the theme changed first? Or just don't consider this corner case).
In the past, we had some performance troubles because we were doing too much work on paintEvent functions.
That being said, maybe the drawing is simple enough so it wouldn't impact the performance (worth to check).

narunlifescience: yes, the function call inside the paint event is only needed when the theme changes. as you mentioned, I thought that it wont be that expensive to put it inside the paint event. However, it cannot be omitted as clementine would need a restart for the theme to take effect otherwise. if the foreground color get changed manually from clementine (appearance settings), it wont even detect the change. 
I modified the code such that the function get called only in case of a theme change.

narunlifescience:
Done!

PR #5036, opened 2015-09-15
Use "title" field for name if "text" is empty in OPML when importing podcast list.

jbroadus:
Some applications (example: gpodder) omit the "text" attribute if "title" exists when exporting OPML. If the "text" attribute does not exist, then try "title".

### DIFF
--- a/src/internet/podcasts/podcastparser.cpp
+++ b/src/internet/podcasts/podcastparser.cpp
@@ -291,7 +291,13 @@ void PodcastParser::ParseOutline(QXmlStreamReader* reader,
           // Parse the feed and add it to this container
           Podcast podcast;
           podcast.set_description(attributes.value("description").toString());
-          podcast.set_title(attributes.value("text").toString());
+          QString title = attributes.value("text").toString();
+          // The OPML 2.0 specification requires a "text" attribute. Some
+          // applications, however, will leave this empty if "title" exists.
+          if (title.isEmpty()) {
+             title = attributes.value("title").toString();
+          }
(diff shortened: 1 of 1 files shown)


src/internet/podcasts/podcastparser.cpp
```diff
@@ -291,7 +291,13 @@ void PodcastParser::ParseOutline(QXmlStreamReader* reader,
           // Parse the feed and add it to this container
           Podcast podcast;
           podcast.set_description(attributes.value("description").toString());
-          podcast.set_title(attributes.value("text").toString());
+          QString title = attributes.value("text").toString();
```
narunlifescience: do we need to declare a variable to store this ? what about 
if (attributes.value("text").toString().isEmpty())
   podcast.set_title(attributes.value("title").toString());
else
   podcast.set_title(attributes.value("text").toString());

jbroadus: It just saves one call to toString() in one case. Negligible, so I'm happy to change it if that's more in line with the Clementine coding standard.

narunlifescience: but you declare a variable instead. any way that was a personal comment..

jbroadus: True. It's an extra copy construction.

hatstand:
Can you squash the commits?

jbroadus:
I was trying to rebase, but was getting a lot of conflicts. It might be easier if I start clean.

PR #5042, opened 2015-09-17
volume slider handle glow effect using system theme

narunlifescience:
with this modification volume slider glow effect can use system theme highlight color. If at some point, there comes a need to change the volume slider handle icon, this code should work there as well. 

untitled


src/widgets/sliderwidget.cpp
```diff
@@ -366,3 +353,41 @@ QPixmap Amarok::VolumeSlider::volumePixmapDraw () const {
   // Return QPixmap
   return pixmap;
 }
+
+void Amarok::VolumeSlider::VolumeSliderHandle() {
```
ArnaudBienner: Please use a more meaningful name (e.g. drawVolumeSliderHandle)

src/widgets/sliderwidget.h
```diff
@@ -99,6 +99,7 @@ class VolumeSlider : public Slider {
  public:
   VolumeSlider(QWidget* parent, uint max = 0);
   QPixmap volumePixmapDraw() const;
+  void VolumeSliderHandle();
```
ArnaudBienner: Should be private

src/widgets/sliderwidget.h
```diff
@@ -99,6 +99,7 @@ class VolumeSlider : public Slider {
  public:
   VolumeSlider(QWidget* parent, uint max = 0);
   QPixmap volumePixmapDraw() const;
```
ArnaudBienner: This one should be private too

src/widgets/sliderwidget.cpp
narunlifescience: Ok

src/widgets/sliderwidget.h
narunlifescience: Oh yes, they are accessed only inside public member function of the class.. I can change that

src/widgets/sliderwidget.h
narunlifescience: I can change that too..

narunlifescience:
Done!

ArnaudBienner:
@hatstand do you want to review this as well?

narunlifescience:
Done!

PR #5050, opened 2015-09-21
make library view sort line themable

narunlifescience:
untitled

@ArnaudBienner could you please check if the alpha level is ok ? or should i tweak the value ?

### DIFF
--- a/src/library/libraryview.cpp
+++ b/src/library/libraryview.cpp
@@ -101,7 +101,9 @@ void LibraryItemDelegate::paint(QPainter* painter,
     painter->drawText(text_rect, text);
 
     // Draw the line under the item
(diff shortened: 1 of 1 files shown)


ArnaudBienner:
An alpha value of 100 looks more like what we previously had.
So if 100 is good for you, maybe we can use that? I feel like line looks too "strong" with a 200 alpha value, and we don't need to have it so visible. Or maybe something in between; as you prefer :)

narunlifescience:
alpha value of 100 looks fine

untitled

narunlifescience:
Done!

ArnaudBienner:
Thanks for the patch :)

PR #5051, opened 2015-09-24
check custom icon location, then system theme & then fallback theme

narunlifescience:
clementine looks for the icon first in the system theme folder and if it couldn't find, then the fallback icons are used. here I try to implement custom icon loading from a location ("clementine settings folder"/customiconsetfolder). with this patch, clementine will first look at the custom location for the icon, then system theme & finally the fallback icons. This should make clementine icons easily customable for users (just copy the icons to custom icon set location folder) without messing around with system themes & source code. Those users who don't care about customization wont feel a difference though. 

screenshot 1 (unmodified clementine running in plasma 5.4 with breeze dark icon set)
snapshot27

screenshot 2 (modified clementine running in plasma 5.4 with custom icon set)
snapshot26

here is the custom icon set I used >
http://www.filedropper.com/customiconset
just extract it to the settings folder (/home/$USER/.config/Clementine in case of linux)

at present form, this work is incomplete (need to add support for providers & lastfm custom icons). This pull request is just for a feedback.. 

need to test in macosx & windows


src/ui/iconloader.cpp
```diff
@@ -31,10 +33,20 @@ void IconLoader::Init() {
 QIcon IconLoader::Load(const QString& name) {
   QIcon ret;
 
+  // If the icon name is empty 
   if (name.isEmpty()) return ret;
 
+  // Try to load icons from the custom icon location initially
+  const QString locate(QDir::toNativeSeparators(Utilities::GetConfigPath(Utilities::Path_Icons) + "/icons/%1x%2/%3.png"));
```
hatstand: I don't think you need `toNativeSeparators`, just stick to `/` everywhere and it will just work with Qt functions.

narunlifescience: Ok, I will change that..

narunlifescience:
Patch is complete now. here is the screen shorts..
using custom icons>
snapshot29

using fallback icons>
snapshot28

if some one would like to test here is the updated custom icon set package
http://www.filedropper.com/customiconset_1

narunlifescience:
@hatstand could you please check this ? It should be working fine now.. although, I am not perfectly fine with the usage of 3 functions which basically do the same job of loading icons (Load, LoadProvider & LoadLastfm). But that should work fine..

a better way would be to use
  enum IconType {
    base,
    provider,
    lastfm
};
and then pass IconType as a second argument to IconLoader::Load function. this way only one function will be needed to load icons. I did not do that because there was too much calls to Load function through out.

src/ui/iconloader.cpp
```diff
@@ -50,3 +66,75 @@ QIcon IconLoader::Load(const QString& name) {
...
+
+// Last fm icons
+QIcon IconLoader::LoadLastfm(const QString& name) {
+  QIcon ret;
+
+  // If the icon name is empty 
+  if (name.isEmpty()) return ret;
+
+  const QString lastfm_location = Utilities::GetConfigPath(Utilities::Path_Icons) + "/last.fm";
+  if(QDir(lastfm_location).exists()) {
```
hatstand: This code is more or less duplicated from the other two Load functions

narunlifescience: Sorry! let me optimize..

narunlifescience:
@hatstand now it should be fine..

src/ui/iconloader.cpp
```diff
@@ -50,3 +66,71 @@ QIcon IconLoader::Load(const QString& name) {
...
+
+#if QT_VERSION >= 0x040600
+  // Then try to load it from the system theme
+  ret = QIcon::fromTheme(name);
+  if (!ret.isNull()) return ret;
+#endif
+
+  // Otherwise use our fallback theme
+  const QString path(":/providers/%1x%2/%3.png");
+  for (int size : sizes_) {
```
hatstand: This size search loop is still duplicated.

narunlifescience:
@hatstand could you check it now ?

hatstand:
Can you fix the conflicts and squash it?

narunlifescience:
Done!

src/ui/iconloader.cpp
```diff
@@ -17,30 +17,82 @@
...
 
 QList<int> IconLoader::sizes_;
+QString IconLoader::custom_icon_path_;
+QList<QString> IconLoader::icon_sub_path_;
 
 void IconLoader::Init() {
   sizes_.clear();
   sizes_ << 22 << 32 << 48;
+  custom_icon_path_ = Utilities::GetConfigPath(Utilities::Path_Icons);
+  icon_sub_path_.clear();
```
hatstand: This is pointless

src/ui/iconloader.cpp
```diff
@@ -17,30 +17,82 @@
...
+  icon_sub_path_ = {"/icons", "/providers", "/last.fm"};
```
hatstand: Hopefully this syntax works on all our platforms

src/ui/iconloader.h
```diff
@@ -22,13 +22,21 @@
 
 class IconLoader {
  public:
+  enum IconType {
+    base = 0,
```
hatstand: `base` -> `Base` and use `enum class` if possible

src/ui/iconloader.cpp
narunlifescience: yes indeed! let me change that

src/ui/iconloader.h
narunlifescience: Ok, i will try to use enum class. did you mean to change base to Base, provider to Provider and so on ?

hatstand: Exactly, yes

narunlifescience: OH! that should take ages... hahaha

hatstand: `find src -type f -print0 | xargs -0 sed -e 's/IconLoader::base/IconLoader::Base/'`

hatstand: oops, should be `sed -i`

narunlifescience: will try that

narunlifescience:
@hatstand I am not able to use enum class! The build fails with a similar error message and a warning like when you use enum class in a qt project without QMAKE_CXXFLAGS += -std=gnu++11. It seems like an issue with cmake files.

hatstand:
Ok, thanks for trying. Can you squash it all once more and then I think we're done

narunlifescience:
Done!

hatstand:
I think you need to merge with master again, there are some pretty weird diffs.

narunlifescience:
Looks like I have to start out fresh.. It seems pretty messed up.. I will close this request and make another ..

PR #5062, opened 2015-10-05
[Qt5] Fix the command line options when another instance is running

Chocobozzz:


src/main.cpp
```diff
@@ -298,7 +298,7 @@ int main(int argc, char* argv[]) {
         qLog(Info)
             << "Clementine is already running - activating existing window";
       }
-      if (a.sendMessage(options.Serialize(), 5000)) {
+      if (a.sendMessage(QString::fromLatin1(options.Serialize().toBase64()), 5000)) {
```
hatstand: Probably better to just change the serialization format to base64

src/ui/mainwindow.cpp
```diff
@@ -2047,13 +2047,13 @@ void MainWindow::PlaylistEditFinished(const QModelIndex& index) {
 
 void MainWindow::CommandlineOptionsReceived(
     const QString& serialized_options) {
-  if (serialized_options == "wake up!") {
+  if (QByteArray::fromBase64(serialized_options.toLatin1()) == "wake up!") {
```
Chocobozzz: Not sure if I keep this part ?

Chocobozzz:
Seems to be better like that

src/core/commandlineoptions.cpp
```diff
@@ -293,7 +293,7 @@ bool CommandlineOptions::is_empty() const {
          toggle_pretty_osd_ == false && urls_.isEmpty();
 }
 
-QByteArray CommandlineOptions::Serialize() const {
+QString CommandlineOptions::Serialize() const {
```
hatstand: You can still use `QByteArray` everywhere I think

src/ui/mainwindow.cpp
hatstand: just remove it I think

Chocobozzz:
Done, thanks for the review :)

PR #5074, opened 2015-10-14
check custom icon location, then system theme & then fallback theme

narunlifescience:
same as pull request #5051


narunlifescience:
@hatstand now it should be fine..

PR #5077, opened 2015-10-14
Lock/Unlock Ratings edit status

### LINKED ISSUE #5029: Add a 'const' rating column
Hi,
I'd like to suggest to add an option to show 'const' rating column in main window where it wouldn't be possible to modify any song rating.
This would prevent changing a song rating by unconscious mouse click.
Preferences->General->Behaviour: "Enable song metadata inline edition with click" does not affect rating column.

narunlifescience:
closes #5029
snapshot31

### DIFF
--- a/src/playlist/playlistheader.cpp
+++ b/src/playlist/playlistheader.cpp
@@ -21,6 +21,7 @@
 #include <QtDebug>
 #include <QContextMenuEvent>
 #include <QMenu>
+#include <QSettings>
 #include <QSignalMapper>
 
 PlaylistHeader::PlaylistHeader(Qt::Orientation orientation, PlaylistView* view,
@@ -32,6 +33,13 @@ PlaylistHeader::PlaylistHeader(Qt::Orientation orientation, PlaylistView* view,
   hide_action_ = menu_->addAction(tr("&Hide..."), this, SLOT(HideCurrent()));
   stretch_action_ = menu_->addAction(tr("&Stretch columns to fit window"), this,
                                      SLOT(ToggleStretchEnabled()));
+  rating_lock_ = menu_->addAction(tr("&Lock Rating"), this, 
+                                  SLOT(ToggleRatingEditStatus()));
+  rating_lock_->setCheckable(true);
+  QSettings s;
+  s.beginGroup("Playlist");
+  rating_lock_->setChecked(s.value("RatingLocked", false).toBool());
+  s.endGroup();
   menu_->addSeparator();
 
   QMenu* align_menu = new QMenu(tr("&Align text"), this);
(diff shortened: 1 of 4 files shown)


src/playlist/playlistheader.cpp
```diff
@@ -22,6 +22,7 @@
 #include <QContextMenuEvent>
 #include <QMenu>
 #include <QSignalMapper>
+#include <QSettings>
```
hatstand: alpha sort

src/playlist/playlistheader.cpp
```diff
@@ -71,6 +79,10 @@ void PlaylistHeader::contextMenuEvent(QContextMenuEvent* e) {
     QString title(
         model()->headerData(menu_section_, Qt::Horizontal).toString());
     hide_action_->setText(tr("&Hide %1").arg(title));
+    
+    // show rating_lock action only for ratings section
+    if (title == "Rating") rating_lock_->setVisible(true);
```
hatstand: Use `{}`

src/playlist/playlistview.h
```diff
@@ -244,6 +246,8 @@ class PlaylistView : public QTreeView {
   int drop_indicator_row_;
   bool drag_over_;
 
+  bool rating_lock_status_; // To store Ratings section lock status
```
hatstand: Can you rename this to something more obvious about which way round is locked or not? e.g. `ratings_locked_`

narunlifescience:
Done!

src/playlist/playlistheader.cpp
```diff
@@ -71,6 +79,13 @@ void PlaylistHeader::contextMenuEvent(QContextMenuEvent* e) {
     QString title(
         model()->headerData(menu_section_, Qt::Horizontal).toString());
     hide_action_->setText(tr("&Hide %1").arg(title));
+    
+    // show rating_lock action only for ratings section
+    if (title == "Rating") {
+      rating_lock_->setVisible(true);
+    } else { 
+      rating_lock_->setVisible(false);
+    }
```
hatstand: actually, this would be nicer as:
```c++
rating_lock_->setVisible(title == "Rating")
```

narunlifescience: that looks clean & less cluttered..

narunlifescience:
Done!

src/playlist/playlistheader.cpp
```diff
@@ -71,6 +79,9 @@ void PlaylistHeader::contextMenuEvent(QContextMenuEvent* e) {
     QString title(
         model()->headerData(menu_section_, Qt::Horizontal).toString());
     hide_action_->setText(tr("&Hide %1").arg(title));
+    
+    // show rating_lock action only for ratings section
+    rating_lock_->setVisible(title == "Rating")
```
hatstand: Does this work when the titles are translated?

narunlifescience: oops ;

narunlifescience:
now it should work for other languages as well..

src/playlist/playlistheader.cpp
hatstand: I think `menu_section_` corresponds to `Playlist::Column` so you should use that instead of the string title, e.g.
```c++
rating_lock_->setVisible(menu_section_ == Playlist::Column_Rating)
```
but make sure it actually works.

narunlifescience:
rating_lock_->setVisible(menu_section_ == Playlist::Column_Rating) works!

Chocobozzz:
This seems to break the left click on the playlist view: if the "lock rating" is not checked (which is the default case) the click in the playlist view does nothing.

narunlifescience:
OH no! Sorry guys, I missed it.. need to test more before pull requests.. I will be careful hereafter..

Issue #2461, opened 2013-12-10
Play beyond end of track in cue file


Clementine-Issue-Importer:
_From [leonel.m...@gmail.com](https://code.google.com/u/113340714934334129103/) on December 04, 2011 13:11:00_

What steps will reproduce the problem? 1. Play a track from a cue file with many tracks in a single FLAC/APE file (CD image, for instance).
2. Listen to the end of the track, it will play 2/3 more seconds beyond the end into the next track. What version of the product are you using? On what operating system? Issue present in 1.0rc in Windows binary, haven't tested other platforms. Please provide any additional information below. This problem occurs with every track in every cue file I tested.

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=2461_

WGH-:
It's still an issue as of Clementine 1.4 rc1. Strangely enough, when this happens, the progress bar may show something like 4:25 out of 4:18. The playback starts at the correct time, and 4:18 is the correct length. The question is why Clementine forgets to stop.

WGH-:
And I also have one song that tend to end 6 seconds earlier. I think most songs stop correctly, though.

PR #5085, opened 2015-10-23
Disable cheating with pipeline

### LINKED ISSUE #4751: single file flac album, track switching problem in shuffle mode
I'm having an interesting problem when playing a single file flac album associated with a cue file. This happens when **shuffle mode** is on. 

When currently playing song finishes, clementine doesn't immediately switch to the *next random song*, but plays the next track (by order) in the album for a few seconds then it switches to randomly selected song. That's a bit annoying, because each time a track finishes, I listen to the next numbered one for a few seconds then switch to another one.. Even the time moves out of its range, see:

clementine_timeline

Same problem occurs with queuing when shuffle is off. I listen to the next track a couple seconds then it switches to queued one. 

I split this file to separate files and didn't observe this problem so I don't think it is a problem with the cue file.

zluca:
This disabling cheating with gstreamer pipeline when single file album file used. Should fix #4751 and #2461

### DIFF
--- a/src/engines/gstenginepipeline.cpp
+++ b/src/engines/gstenginepipeline.cpp
@@ -850,24 +850,12 @@ GstPadProbeReturn GstEnginePipeline::HandoffCallback(GstPad*,
 
     if (end_time > instance->end_offset_nanosec_) {
       if (instance->has_next_valid_url()) {
-        if (instance->next_url_ == instance->url_ &&
-            instance->next_beginning_offset_nanosec_ ==
-                instance->end_offset_nanosec_) {
-          // The "next" song is actually the next segment of this file - so
-          // cheat and keep on playing, but just tell the Engine we've moved on.
           instance->end_offset_nanosec_ = instance->next_end_offset_nanosec_;
           instance->next_url_ = QUrl();
           instance->next_beginning_offset_nanosec_ = 0;
           instance->next_end_offset_nanosec_ = 0;
-
-          // GstEngine will try to seek to the start of the new section, but
-          // we're already there so ignore it.
-          instance->ignore_next_seek_ = true;
+          instance->ignore_next_seek_ = false;
           emit instance->EndOfStreamReached(instance->id(), true);
-        } else {
-          // We have a next song but we can't cheat, so move to it normally.
-          instance->TransitionToNext();
-        }
       } else {
         // There's no next song
         emit instance->EndOfStreamReached(instance->id(), false);


hatstand:
Pretty sure this will break things, gapless playback of cue files for one.

zluca:
Gapless playback requires removing gap between INDEX 00 and INDEX 01 wich is just ignored for now. And when i maked cueparser parse INDEX 00 i found that this bug is prevent it from normal work.
So "gapless" playback only work for a specific cue's without INDEX 00.

Anyway i use this for myself cos it's smaller evil, than 2 seconds gap in some cues.

zluca:
Ok. Im returned cheating. But with some changes. Now it should work fine. At least i can't hear any problem in several flacs. And bugs looks fixed for me.

WGH-:
Here's updated patch for current master (I ported without really understanding it, though):
```diff
diff --git a/src/engines/gstenginepipeline.cpp b/src/engines/gstenginepipeline.cpp
index 23b05ed3c..550db28bc 100644
--- a/src/engines/gstenginepipeline.cpp
+++ b/src/engines/gstenginepipeline.cpp
@@ -920,21 +920,18 @@ GstPadProbeReturn GstEnginePipeline::HandoffCallback(GstPad*,
         if (instance->next_.url_ == instance->current_.url_ &&
             instance->next_beginning_offset_nanosec_ ==
                 instance->end_offset_nanosec_) {
-          // The "next" song is actually the next segment of this file - so
-          // cheat and keep on playing, but just tell the Engine we've moved on.
-          instance->end_offset_nanosec_ = instance->next_end_offset_nanosec_;
-          instance->next_ = MediaPlaybackRequest();
-          instance->next_beginning_offset_nanosec_ = 0;
-          instance->next_end_offset_nanosec_ = 0;
-
           // GstEngine will try to seek to the start of the new section, but
           // we're already there so ignore it.
           instance->ignore_next_seek_ = true;
-          emit instance->EndOfStreamReached(instance->id(), true);
         } else {
           // We have a next song but we can't cheat, so move to it normally.
-          instance->TransitionToNext();
+          instance->ignore_next_seek_ = false;
         }
+        instance->end_offset_nanosec_ = instance->next_end_offset_nanosec_;
+        instance->next_ = MediaPlaybackRequest();
+        instance->next_beginning_offset_nanosec_ = 0;
+        instance->next_end_offset_nanosec_ = 0;
+        emit instance->EndOfStreamReached(instance->id(), true);
       } else {
         // There's no next song
         emit instance->EndOfStreamReached(instance->id(), false);
```

I tested the patch on my test cases (https://github.com/clementine-player/Clementine/issues/2461#issuecomment-626197453), and unfortunately it's far from ideal.

The song that was previously consistently overshoot (4:25 out of 4:18) now cuts _a little_ bit earlier: at 4:16 or 4:17, I think (unless it's followed by the next song of the raw file in the playlist, in which case the playback seems gapless). Not as annoying as that overshooting, but still.

I think most or all songs are now cut a little bit earlier, it's just many of them have silence at the end, so it's not really apparent.

The song that ends 4 seconds early is the last in the raw file, and this patch doesn't change anything about this song (this might be a different bug). The song is played correctly to the end only when it's the last entry in the playlist (no matter whether this patch applied or vanilla master).

PR #5089, opened 2015-10-26
fix rating half star issue

narunlifescience:
when rating a song with half stars (0.5, 1.5, 2.5 etc), becomes slightly overlapped by the next. I found this issue related to kStarSize / 2. changing  kStarSize value to an even number will completely solve this issue... 

before
snapshot32

after
snapshot33


ArnaudBienner:
Following up conversation in #5137 and comment by @TheUbuntuGuy I feel like the star icons where nicer previously.
I guess you needed to change the PNGs to give them a even size, but can't you just resize them without modifying them?

narunlifescience:
@ArnaudBienner I made a pull request with old PNGs re-sized

Issue #3355, opened 2013-12-12
Tag fetcher applies incorrect tags for songs without any results


Clementine-Issue-Importer:
_From [dhuli...@gmail.com](https://code.google.com/u/117638345113817948213/) on December 12, 2012 23:39:10_

What steps will reproduce the problem? 1. Select two tracks, one with MusicBrainz metadata, one without. Right Click > `Edit Track Information`.
2. Click `Complete Tags Automatically`
3. Click `Ok` What is the expected output? What do you see instead? Instead of leaving the tags of unfound track unchanged, Clementine's Tag Fetch with apply all the tags found from last found track to the unfound track. 

Example: The first track's artist tag was correctly fetched as `Michael Jackson`. No results were found for the second track. The notification, `Clementine was unable to find results for this file`, is displayed. If I click `Ok`, The second track's artist will now be changed to `Michael Jackson`. What version of the product are you using? On what operating system? Clementine 1.1.1, Windows 7 Ultimate 64-bit Please provide any additional information below.

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=3354_

PR #5094, opened 2015-10-30
Fix 3355: Tag fetcher now ignores songs without any results

terrorfisch:
This commit fixes issue #3355

Before this fix the fetched data of the first title was applied to all selected titles. If the other titles hat correctly fetched tags, this information was overridden by the correct data later on. So the error only occured in case of failure during fetching.

I'm however not sure if the UI-update check is efficient.


src/ui/edittagdialog.cpp
```diff
@@ -842,26 +842,32 @@ void EditTagDialog::FetchTagSongChosen(const Song& original_song,
...
-      ui_->track->setValue(new_metadata.track());
-      ui_->year->setValue(new_metadata.year());
-    } else {
-      data->current_.set_title(new_metadata.title());
-      data->current_.set_artist(new_metadata.artist());
-      data->current_.set_album(new_metadata.album());
-      data->current_.set_track(new_metadata.track());
-      data->current_.set_year(new_metadata.year());
+  int id;
+  for (id = 0; id < data_.count(); ++id)
```
hatstand: I think you can do something like:
```c++
auto it = std::find_if(data_.constBegin(), data_.constEnd(), [filename](const Data& d) {
  return d.original_.url().toLocalFile() == filename;
}
if (it != data_.constEnd()) {
  const Data* data = &*it;
...
```

src/ui/edittagdialog.cpp
```diff
@@ -842,26 +842,32 @@ void EditTagDialog::FetchTagSongChosen(const Song& original_song,
...
+    if (data_[id].original_.url().toLocalFile() == filename)
+      break;
+  
+  if( id == data_.count() ) {
+    qLog(Warning) << "Could not find song to filename: " << filename;
+    return;
+  }
+  
+  Data& data = data_[id];
```
hatstand: Please run `make format` and follow the [style guide](https://google-styleguide.googlecode.com/svn/trunk/cppguide.html)
i.e. `{}` for all `for` and `if` statements, no use of non-const references

PR #5098, opened 2015-11-02
Update about.cpp

DevynCJohnson:
If it is okay with you, I would like to be listed since I added some equalizers.


trougnouf:
Your contribution is rather insignificant. The guy above you (@TheUbuntuGuy ) has 81 commits (3,049 ++  1,812 --), @Chocobozzz  doesn't even seem to be in the list for some reason even though he is responsible for the Qt5 port, I think this commit is incomplete so long as it doesn't include all 100 people on https://github.com/clementine-player/Clementine/graphs/contributors .

PR #5107, opened 2015-11-08
Mouse wheel track slider Seeking

### LINKED ISSUE #4568: Seeking with mouse wheel broken
When I try to seek a track using the mouse wheel while pointer is on the track slider / moodbar the song is always rewinded by a few ms. If I do that to often in a row (~ 5-10 times) the track starts from the beginning.

I saw
  Amarok::Slider::wheelEvent(QWheelEvent* e)
in
  widgets/sliderwidget.cpp
It says vertical scroll events "Will be handled by the parent widget" but where exactly is that?

I'm using Version 1.2.3-782-g78804b1

narunlifescience:
Fixes #4568

### DIFF
--- a/src/widgets/tracksliderslider.cpp
+++ b/src/widgets/tracksliderslider.cpp
@@ -24,6 +24,7 @@
 #include <QStyle>
 #include <QStyleOptionSlider>
 #include <QtDebug>
+#include <QWheelEvent>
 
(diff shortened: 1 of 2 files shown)


hatstand:
Please run `make format`

narunlifescience:
Done!

PR #5123, opened 2015-11-21
combine nyancatanalyzer & rainbowdashanalyzer(eliminate duplicated code)

narunlifescience:


hatstand:
I think you can make this a lot cleaner by adding back a class each for RainbowDash and NyanCat that inherits from your new RainbowAnalyzer. Then you can avoid all the `newInstance` magic and you don't need to change the constructor signature.

narunlifescience:
Done!

src/analyzers/rainbowanalyzer.cpp
```diff
@@ -18,24 +21,33 @@
...
-const char* NyanCatAnalyzer::kName = "Nyanalyzer cat";
-const float NyanCatAnalyzer::kPixelScale = 0.02f;
+const int Rainbow::RainbowAnalyzer::kHeight[] = { 21, 33 };
+const int Rainbow::RainbowAnalyzer::kWidth[] = { 34, 53 };
+const int Rainbow::RainbowAnalyzer::kFrameCount[] = { 6, 16 };
+const int Rainbow::RainbowAnalyzer::kRainbowHeight[] = { 21, 16 };
+const int Rainbow::RainbowAnalyzer::kRainbowOverlap[] = { 13, 15 };
+const int Rainbow::RainbowAnalyzer::kSleepingHeight[] = { 24, 33 };
+
+const char* Rainbow::NyanCatAnalyzer::kName = "Nyan Cat Rainbow";
```
hatstand: Can you change the name back to Nyanalyzer cat?

narunlifescience: OK

narunlifescience:
Done!

narunlifescience:
I was trying to add just one RainbowAnalyzer to analyzercontainer.cpp AddAnalyzerType<RainbowAnalyzer>() 
and then use a second invokable function declared in RainbowAnalyzer class to toggle between RainbowType enum values (Nyancat & rainbow dash). But with current implimentation of analyzercontainer, I am unable to do that..

hatstand:
I don't think that optimisation is really worth it.

narunlifescience:
Yes, indeed.. but at times i get caught in some silly things and cant get it out of my head..

PR #5124, opened 2015-11-22
Respawn mainwindow on correct monitor

TheUbuntuGuy:
PR #5081 created a regression where the mainwindow will always start on the primary monitor regardless of which one it was closed on when maximized. This restores previous functionality.


narunlifescience:
@TheUbuntuGuy Thanks for the fix! I missed that scenario..

Issue #1246, opened 2013-12-07
Doesn't detect the track length of some files, so I can't seek through the track


Clementine-Issue-Importer:
_From [gmandx](https://code.google.com/u/gmandx/) on January 06, 2011 07:26:13_

What steps will reproduce the problem? 1. Load the problematic file
2. Notice that the length column doesn't have a value
3. Play the track What is the expected output? What do you see instead? It's supposed to show the track length and the track slider should move as the song advances. Instead, the track duration is not detected, the slider is disabled, and all numbers aside the slider are set to "00:00". What version of the product are you using? On what operating system? Ubuntu 10.04 (Lucid), Clementine 0.6.90~ r2562 ~lucid (from development PPA) Please provide any additional information below. I managed to "fix" the issue thanks in part to MP3 Diags [1]. When I load the files in MP3 Diags, I notices that they had several ID3 tags (each one in a diferent ID3 version). So, with MP3 Diags, I only apply one transformation, the one that only leaves the last ID3v1 tag encountered and the first ID3v2 tag encountered (and ended up being the ID3v2.4 version) [2]. Then, in Clementine, the "Update Library" menu doesn't appear to update the loaded tracks, so I just opened the track info dialog and clicked "Ok", without actually change anything. The track length then got updated, and the slider got working again.

So, I guess that Clementine has problems when loading files with multiple ID3 tags, that messes with the track length detection.

[1] http://mp3diags.sourceforge.net/index.html [2] http://mp3diags.sourceforge.net/145_transformations_list.html

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=1246_

malexandre:
And do we know when will be the next Clementine release including this correction? A lot of the podcasts I managed with Clementine does not have the track length. It's kind of annoying to continue playing the episode on another device.

Chocobozzz:
I don't know for the next release but you can install the latest build http://builds.clementine-player.org/.

If you use GNU/Linux, you'll have to wait a new release of Taglib (or compile the latest sources).

malexandre:
It works better. I built the last Taglib and installed Clementine from the last build you gave. The track slider, the current time and the total time displayed besides the track slider works fine. But the "length" column is still empty and the total playlist time does not take into account the tracks with the problem. Here's a screenshot:
![](http://i.imgur.com/wF41AnL.png)

supine:
[...]

PR #5129, opened 2015-11-24
Update taglib to 1.10.0

Chocobozzz:
Taglib 1.10 is out (https://github.com/taglib/taglib/releases) :). I updated the 3rdparty directory inside Clementine.

I think it would be useful to update https://github.com/clementine-player/Dependencies with this new release because I think that some Clementine issues would be fixed (maybe https://github.com/clementine-player/Clementine/issues/221 or https://github.com/clementine-player/Clementine/issues/1246 etc).

Unfortunately, I can't make a PR since packages are downloaded from @davidsansome google drive account (I guess).


jerobarraco:
Awesome! 
I've been testing the dev-ppa and i've found at least 2 bugs that seems to be working now.
1) Changing the tag info while playing, then change track and the info doesnt stay persisted seems to be working (though not really sure)
2) Library stuck at updating: seems to work now :)

great :)

PR #5131, opened 2015-11-25
Fix bug in edit tag dialog when fetching

terrorfisch:
Before: When only one song is selected and the tag fetcher returns the value a field already has, the value of the field gets erased. The reason for this became not 100% clear to me. This bug was introduced in #5094 

The changes prevent that a field gets updated when the value did not change. Furthermore the fields changed by the tag fetcher are no longer hard coded to allow easy extending and more readable code.

### DIFF
--- a/src/musicbrainz/tagfetcher.cpp
+++ b/src/musicbrainz/tagfetcher.cpp
@@ -27,6 +27,9 @@
 #include <QUrl>
 #include <QtConcurrentMap>
 
+const std::array<QString, 5> TagFetcher::kFetchedFields{
+    {"title", "artist", "album", "track", "year"}};
+
 TagFetcher::TagFetcher(QObject* parent)
     : QObject(parent),
       fingerprint_watcher_(nullptr),
--- a/src/musicbrainz/tagfetcher.h
+++ b/src/musicbrainz/tagfetcher.h
@@ -24,6 +24,8 @@
 #include <QFutureWatcher>
 #include <QObject>
 
+#include <array>
+
 class AcoustidClient;
 
 class TagFetcher : public QObject {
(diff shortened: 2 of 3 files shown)


src/musicbrainz/tagfetcher.cpp
```diff
@@ -27,6 +27,9 @@
 #include <QUrl>
 #include <QtConcurrentMap>
 
+const std::array<QString, 5> TagFetcher::kFetchedFields{
```
hatstand: Please avoid static non-POD variables.

terrorfisch: Do you think it's reasonable to switch to an enum class for the field id?

hatstand: IIRC, c++11 enum class doesn't compile on some of our platforms.

hatstand: I take it back, we're using it elsewhere already so it must work now.

terrorfisch: Have you any objections adding the methods
```
QVariant get_attribute( AttributeId ) const;
void set_attribute( AttributeId, const QVariant& );
```
to the Song class and make AttributeId a nested enum? My feeling is, that it logically belongs there and may be useful somewhere else.

hatstand: Not sure, `Song::kColumns` is a similar id list; could merge them but I think there might be fields in `Song` that aren't necessarily in the database.

abika:
I just stumbled across this. The problem is that in `InitFieldValue()` the `editor->clear()` call emits the signal `textChanged()` (if the editor already had a value) and this triggers the field to be erased in `UpdateFieldValue()`.

I would suggest a solution without calling `InitFieldValue()`, the code is already complicated enough.

And I cannot confirm that this patch actually fixes the bug!?

terrorfisch:
> And I cannot confirm that this patch actually fixes the bug!?

I'm quite shure it did for me. But if the reason you worked out is true it indeed shouldn't.

I will close this pull request, as I dont have the time atm to refine it. Maybe you can come up with a better fix.

Issue #1241, opened 2013-12-07
Moving files within library loses ratings/playcounts


Clementine-Issue-Importer:
_From [Gavin77](https://code.google.com/u/111310800507815713391/) on January 04, 2011 17:26:06_

What steps will reproduce the problem? 1. I have 2 main folders both included in the library, first is music I need to backup, 2nd is my main archive.
2. Once I've backed up some music I move it to my main archive folder.
3. The ratings and playcounts on those files are then lost. What is the expected output? What do you see instead? Ratings/playcounts should be kept intact.  They are not. What version of the product are you using? On what operating system? 0.6 r2562 Kubuntu 10.10

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=1241_

sabret00the:
I can accept that some files I change are losing ratings. But some files I've not changed are losing ratings and that's just frustrating. Clementine in its current state just isn't reliable.

abika:
If somebody is still interested: #5135

v-fox:
Wouldn't using checksums of accompanying moodbar-files, which basically are fingerprints, be perfect for that ? Even replacing filepaths in playlists with them. Moodbar really is the best thing that ever happened to music players.

torotil:
This bug means that you can either use the library management functionality or dynamic playlists based on ratings (and last played time). I think both are major features of this player. For me this is a blocker in order to switch over from amarok.

thenewnano:
Can you please merge the #5135  !

PR #5135, opened 2015-11-26
New feature: song fingerprinting and file tracking

abika:
A new  feature for tracking filename changes for playlist songs.

On inserting songs into the playlist a fingerprint is created for each song and the directory of the song is watched for changes. If a songs filename changes, the new file is found by searching in the same directory with the fingerprint of the song.
Of course, this makes tracking currently very limited: only renames are detected. But its only kind of a proof-of-concept, the real benefit comes when fingerprints are added to the database.

### How is the fingerprint calculated?
-> `hash( chromaprint( first10SecsOfSong ) + lengthOfSong )`

There is also a fallback using the first 512byte of a file if Chromaprint fails.

### What are the properties of this calculation method?
* agnostic to file meta/ID2/ID3 tags: YES! E.g. mp3 tags can be inserted/removed/changed, the fingerprint stays the same.
* agnostic to media format/bitrate: unfortunately no. AcoustID is too sensitive.
* uniqueness: tested with 10000+ songs, all FPs were unique 
* calculation speed: ~ 0.1secs per song. This not very fast and I'm open to suggestions for improvement here.

### Why is resampling turned of in Chromaprint for FP creation?
When repeatedly processing the same track GStreamer produces slightly different output streams if resampling is done. The result is different fingerprints for the same file.
(This also happens when requesting Musicbrainz but its not that important here.)

### Why watching directories and not the files directly?
The number of (inotify) filesystem watches is limited. This limit would be easily reached for a large number of playlist songs.
***

As a next step, I'm planning to insert fingerprints to the playlist-backend (=database). But I would like to have some feedback first...)


hatstand:
I'm not really sure if we want this or not yet but you can definitely improve performance by running multiple Chromaprints in parallel. The transcoder does something similar and I believe Chromaprinter is used in parallel elsewhere too.

abika:
Well, the performance gain of using concurrent threads is not that big (less then half of computation time on my machine with 4 CPU cores). And I would like to avoid massive parallel IO work for something that should be running only additionally in the background. The Tagreader is already reading every file and user will certainly complain if Clementine is using almost all system resources, even if its only for a short time.

BTW, in the long run this will eventually close #1241 .

Kwaadpepper:
Hey thanks for this work, i would also say that in ```preference -> automatic updates -> save stats in file tag if possible``` is regularly failing here on linux, but i saw the same behavior on windows too. I don't know the reason of this yet, but sure this option ( disabled by default) is the reason. When activated you will have ```Updating Library 0%```. I believe something is failing on some files for unknown reason  and fall back is not handled properly. This would also fix benefit #1241.
This option is in the same spirit as you are trying to do, i have not tested your branch, but merging this with your behaviour could be nice and in the same time maybe fixing it :smile:

uaEquals42:
I would love to be able to reorganize files without having to worry about loosing statistics/scores for music files.  Any news on whether or not this will be merged?

abika:
after a long time... is there anything that prevents this from merging? I mean the "detect file name changes" option is working fine, it's off by default, and the code doesn't interfere with anything (9 months and still no conflicts).

Can somebody decide this, please?

zuf:
I want this feature in clementine, please merge this PR.

v-fox:
If I had any money I would pay top dollar for this feature to be collection-wide, along with Qt5 build, proper FLAC/cue support and obnoxious tagreader hang fix. After that, you can just retire because you'd made the perfect audio player.

If there are concerns about performance, Clementine really should do any I/O operations besides playing a song as ["idle" ioclass](http://man7.org/linux/man-pages/man1/ionice.1.html) thread.

abika:
first conflict (minor problem in *.ui file), rebased.

> If there are concerns about performance, Clementine really should do any I/O operations besides playing a song as "idle" ioclass thread.

The thread of the songtracker has already `QThread::IdlePriority`.

zuf:
Why it still not merged yet?

Kwaadpepper:
This should really be considered to be include.

abika:
Rebased again.

And moved/deleted songs are now grayed out in the playlist.

comfy33:
So am I understanding correctly that we still can't move files without graying out the file/song in any playlists that song is in? I've been trying all night and can't seem to find a way to enable this.

abika:
Currently there is only support (with this patch and Playlist->"Detect file name changes" enabled) for rename detection of playlist files while Clementine is running.

---
Still waiting for some response e.g. from @hatstand ...

tmarplatt:
#1241 is one issue that bugs me most (just this one, actually) about Clementine. I can't just reorganize my music library by changing directories because it will break custom ratings.

Please consider merging this feature into master. 

Also @abika if you need further testing (5000+ song sample to test under Linux Mint) please walk me over the required steps. I'll be glad to provide results.

abika:
@tmarplatt (and everybody else) If you know how, you can compile and test the renaming feature. Any feedback appreciated.

pschichtel:
I just recently implemented a rename detection approach in [my music synchronizer](https://github.com/pschichtel/ConvertedSync/issues/16) by using the user_xattr support most filesystems offer these days.

Short version: Store the known location in an user-attribute upon first sight. When a new file is detected that already has the attribute, update playlists referencing the old location and update the attribute in the file.

- No complicated fingerprint calculations
- No heavy IO (a "known location" lookup table could be build during indexing)
- Renames can be arbitrary

hatstand:
xattr won't work on Windows or Mac.

pschichtel:
https://en.wikipedia.org/wiki/Extended_file_attributes
The wikipedia and the Java 8 source code tell me that there is xattr support at least in Windows. But if it is not well enough support you could just use your existing database in the same way.

hatstand:
Using the database is the exact problem you are trying to solve by using xattr

pschichtel:
Not really. The intention was to avoid heavy IO and fingerprinting

thenewnano:
Let me tell a short story here : 
I do DJ for Argentine Tango events, we do have many many playlists organised in folders etc , each playlist has 4 songs 4x2.5m , so for a long **night** I use and create **maybe 100 playlists.** I am depending on this function, as clementine seems to be the only player having folder structure for playlists, even if it has no support to search in playlist I still use it as the folders are useful. 

Now to the **nightmare** I have now , I did evolve my metadata use which is a bit scary, but still my library of 42000 songs became crazy unmanageable, too many duplicate etc , I have spend days now reorganising and renaming all files... and next step is **to manually create my playlists again**, and I really can't trust that files don't change name again for ever etc, **syncing between 3 computers where library path has not been the same** is the next nightmare I have coped with in 3 years now! 

**If we fingerprint files then all these issues are gone! there is a merge request so can we please merge it ?**

abika:
Rebased again and made a minor change to be compatible with Qt5/the qt5 branch.

PR #5137, opened 2015-12-01
use iconloader & cleanup previous remnants

narunlifescience:
1) clean up .ui, .h files and use IconLoader when possible
2) use uniform color across tiny start/pause & star on/off
before:
snapshot43
snapshot44
after:
snapshot46
snapshot45
3) compress .svg icon files


ArnaudBienner:
I liked the gradient on the play icon :(
The icon is also used on the Mac tray icon (where it is much bigger than in the playlist view) and it looked better to me previously (but maybe this is a personal taste or I just got used to it).

narunlifescience:
If we encounter more objections to this commit, we can revert it back :)

TheUbuntuGuy:
My problem with the play icon is the fact that all the icons in Clementine are sort of 3D and glossy. This new icon is very flat, and while I agree it is modern looking, it looks very out of place next to all the other icons in the main window. Also, the star now looks kind of "anaemic" and the points are very thin and sharp.

ArnaudBienner:
Commit 2603cf2088e0b032be3ea0cca3a823e0868064f4 about cleanup things and using IconLoader everywhere looks fine to me, but I can't get use to the new icons, sorry. And indeed, I agree with @TheUbuntuGuy that the star icon was better previously.
So if you don't mind, I think I will revert commit aa813fa372ff1a4cd14a13c77c774e036668dad9 for now.
I think it's fine to keep commit bc680f27554f7a1d0f79b0553d42752746fb4d9b since it's just compressing svg without modifying them, right?

narunlifescience:
hmm.. That should be fine..

ArnaudBienner:
commit aa813fa reverted.
Also the star change mentioned by @TheUbuntuGuy isn't related to this commit, but to pull request #5089

PR #5138, opened 2015-12-01
Add "Smart Playlists" for subsonic

nicklan:
Notes:
- These playlists allow access to the subsonic feature of pulling new, frequently played, and so on albums.
- See: http://www.subsonic.org/pages/api.jsp#getAlbumList for the subsonic description
- The subsonic api is inherently album oriented.  Therefore at the moment the "count" argument for GenerateMore is used as the number of albums to pull, which hopefully isn't a problem.
- Could be made more efficient by multi-threading the fetch of the songs for each album

### DIFF
--- a/src/internet/subsonic/subsonicdynamicplaylist.cpp
+++ b/src/internet/subsonic/subsonicdynamicplaylist.cpp
@@ -1,0 +1,259 @@
+/* This file is part of Clementine.
+   Copyright 2015, Nick Lanham <nick@afternight.org>
+
+   Clementine is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation, either version 3 of the License, or
+   (at your option) any later version.
+
+   Clementine is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with Clementine.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "subsonicdynamicplaylist.h"
+
+#include <QEventLoop>
+#include <QFileInfo>
+#include <QSslConfiguration>
+#include <QXmlStreamReader>
+
+#include "core/application.h"
+#include "core/logging.h"
+#include "core/network.h"
+#include "core/taskmanager.h"
+#include "core/timeconstants.h"
+#include "internet/core/internetplaylistitem.h"
+
+SubsonicDynamicPlaylist::SubsonicDynamicPlaylist()
+  : stat_(QueryStat_Newest),
+    offset_(0) {}
+
+SubsonicDynamicPlaylist::SubsonicDynamicPlaylist(const QString& name,
+                                                 QueryStat stat)
+  : stat_(stat),
+    offset_(0) {
+  set_name(name);
+}
+
+void SubsonicDynamicPlaylist::Load(const QByteArray& data) {
+  QDataStream s(data);
+  s >> *this;
+}
+
+void SubsonicDynamicPlaylist::Load(QueryStat stat) {
+  stat_ = stat;
+}
+
+QByteArray SubsonicDynamicPlaylist::Save() const {
+  QByteArray ret;
+  QDataStream s(&ret, QIODevice::WriteOnly);
+  s << *this;
+  return ret;
+}
+
+// copied from SubsonicService
+QNetworkReply* SubsonicDynamicPlaylist::Send(QNetworkAccessManager& network, const QUrl& url, const bool usesslv3) {
+  QNetworkRequest request(url);
+  // Don't try and check the authenticity of the SSL certificate - it'll almost
+  // certainly be self-signed.
+  QSslConfiguration sslconfig = QSslConfiguration::defaultConfiguration();
+  sslconfig.setPeerVerifyMode(QSslSocket::VerifyNone);
+  if (usesslv3) {
+    sslconfig.setProtocol(QSsl::SslV3);
+  }
+  request.setSslConfiguration(sslconfig);
+  QNetworkReply* reply = network.get(request);
+  return reply;
+}
+
+PlaylistItemList SubsonicDynamicPlaylist::Generate() { return GenerateMore(10); }
+
+PlaylistItemList SubsonicDynamicPlaylist::GenerateMore(int count) {
+  SubsonicService* service = InternetModel::Service<SubsonicService>();
+  int task_id = service->app_->task_manager()->StartTask(tr("Fetching Playlist Items"));
+  QUrl url = service->BuildRequestUrl("GetAlbumList");
+  QNetworkAccessManager network;
+
+  switch (stat_) {
+    case QueryStat::QueryStat_Newest:
+      url.addQueryItem("type","newest");
+      break;
+    case QueryStat::QueryStat_Highest:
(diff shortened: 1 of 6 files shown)


src/internet/subsonic/subsonicdynamicplaylist.h
```diff
@@ -0,0 +1,75 @@
...
+  // need our own one since we run in a different thread from service
+  QNetworkReply* Send(QNetworkAccessManager& network, const QUrl& url, const bool usesslv3);
+
+ private:
+  QueryStat stat_;
+  int offset_;
+  SubsonicService* service_;
+};
+
+#include "subsonicservice.h"
```
hatstand: Remove

src/internet/subsonic/subsonicdynamicplaylist.cpp
```diff
@@ -0,0 +1,259 @@
...
+  }
+  request.setSslConfiguration(sslconfig);
+  QNetworkReply* reply = network.get(request);
+  return reply;
+}
+
+PlaylistItemList SubsonicDynamicPlaylist::Generate() { return GenerateMore(10); }
+
+PlaylistItemList SubsonicDynamicPlaylist::GenerateMore(int count) {
+  SubsonicService* service = InternetModel::Service<SubsonicService>();
```
hatstand: It would be nice if you could avoid this singleton call here, is it possible to pass it in the constructor?

src/internet/subsonic/subsonicdynamicplaylist.cpp
```diff
@@ -0,0 +1,259 @@
...
+  int task_id = service->app_->task_manager()->StartTask(tr("Fetching Playlist Items"));
+  QUrl url = service->BuildRequestUrl("GetAlbumList");
+  QNetworkAccessManager network;
+
+  switch (stat_) {
```
hatstand: Refactor most of this into a separate function `QString GetType(QueryStat) const;`

src/internet/subsonic/subsonicdynamicplaylist.cpp
```diff
@@ -0,0 +1,259 @@
...
+      break;
+  case QueryStat::QueryStat_Starred:
+      url.addQueryItem("type","starred");
+      break;
+    case QueryStat::QueryStat_Random:
+      url.addQueryItem("type","random");
+      break;
+  }
+
+  if (count > 500) count = 500;  // 500 limit per subsonic api
```
hatstand: Add named constants for these magic numbers

src/internet/subsonic/subsonicdynamicplaylist.cpp
```diff
@@ -0,0 +1,259 @@
...
+  if (offset_ != 0) { // 0 is default
+    url.addQueryItem("offset",QString::number(offset_));
+  }
+
+  PlaylistItemList items;
+
+  QNetworkReply* reply = Send(network,url,service->usesslv3_);
+
+  // wait for reply
+  {
```
hatstand: There's a utility for this somewhere in clementine.

src/internet/subsonic/subsonicdynamicplaylist.cpp
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  if (reply->error() != QNetworkReply::NoError) {
+    qLog(Warning) << "HTTP error returned from Subsonic:" << reply->errorString()
+                  << ", url:" << url.toString();
+    service->app_->task_manager()->SetTaskFinished(task_id);
+    return items; // empty
+  }
+
+  reply->deleteLater();
```
hatstand: You should do this before the other return above

src/internet/subsonic/subsonicdynamicplaylist.cpp
```diff
@@ -0,0 +1,259 @@
...
+      return items;
+    }
+
+    qLog(Debug) << "Getting album: "<<reader.attributes().value("album").toString();
+    GetAlbum(service,items,reader.attributes().value("id").toString(),network,service->usesslv3_);
+
+    reader.skipCurrentElement();
+  }
+  offset_+=count;
+  service->app_->task_manager()->SetTaskFinished(task_id);
```
hatstand: Might be an idea to write a helper class or use `BOOST_SCOPE_EXIT` to make sure you don't forget to call `SetTaskFinished()` before every return.

src/internet/subsonic/subsonicdynamicplaylist.cpp
```diff
@@ -0,0 +1,259 @@
...
+  return items;
+}
+
+void SubsonicDynamicPlaylist::GetAlbum(SubsonicService* service, PlaylistItemList& list, QString id, QNetworkAccessManager& network, const bool usesslv3) {
+  QUrl url = service->BuildRequestUrl("getAlbum");
+  url.addQueryItem("id", id);
+  QNetworkReply* reply = Send(network, url, usesslv3);
```
hatstand: This is leaked I think

hatstand:
Please run `make format` too

src/internet/subsonic/subsonicdynamicplaylist.cpp
nicklan: I don't think so, because these lists get serialized and then reconstructed in Generator (generator.cpp line 40).  I could maybe put the singleton call there, but it's not clear that would be any better.  Let me know if you'd prefer that and I can see about getting it to work.

nicklan:
Fixed most of the issues you mentioned, and ran format.  Thanks for the review, and let me know if you want me to address some of the stuff that hasn't changed.

src/internet/subsonic/subsonicdynamicplaylist.cpp
```diff
@@ -0,0 +1,257 @@
...
+  PlaylistItemList items;
+
+  QNetworkReply* reply = Send(network, url, service->usesslv3_);
+
+  // wait for reply
+  {
+    QEventLoop loop;
+    connect(reply, SIGNAL(finished()), &loop, SLOT(quit()));
+    loop.exec();
+  }
```
hatstand: Use `WaitForSignal(reply, SIGNAL(finished()));` instead.

nicklan: Right, sorry, forgot to address that before

nicklan:
Okay, that covers everything except the singleton call, which I think can't be avoided, only moved to generator.cpp if I move it to the contructor.  If that's preferred, let me know and I can change that too.  Thanks for the feedback again.

src/internet/subsonic/subsonicdynamicplaylist.cpp
```diff
@@ -0,0 +1,246 @@
...
+  BOOST_SCOPE_EXIT_END
+
+  QUrl url = service->BuildRequestUrl("GetAlbumList");
+  QNetworkAccessManager network;
+
+  url.addQueryItem("type", GetTypeString());
+
+  if (count > kMaxCount) count = kMaxCount;
+  if (count != kDefaultCount) {
```
hatstand: Any reason not to just always set the `size` and `offset` parameters? It would make this code more robust to server changes.

nicklan: Mostly just an efficiency thing, not sending redundant parameters.  I'm happy to just always send it though. 

I will note that the subsonic api specifies these defaults, so they are unlikely to change, but either way.

hatstand: I think the explicit version is better; the 16 byte saving is irrelevant.

nicklan: Okay, no problem, done.

PR #5152, opened 2015-12-11
Add option to save current library grouping

nicklan:
- Adds saved groupings to the Group by menu
- Currently missing a way to manage saved groupings
- Saved groupings are in a new section of the config file as
  "GroupingName"->ByteArray serialization

I'm planning on writing a Saved Groupings Manager, similar to the Queue Manager, that will allow adding/removing saved groupings.  I can make it part of this request, or leave it for another one.

PR #5153, opened 2015-12-12
Add dialog for managing saved files

nicklan:
This obviously depends on PR #5152


nicklan:
See last commit for description.

PR #5163, opened 2015-12-20
Improve the star icon of `favoritewidget`

Ede123:
* Use the star icon in its native size (16 px × 16 px) instead of unnecessarily scaling it down by 1 px
* Center the star icon vertically

### DIFF
--- a/src/widgets/favoritewidget.cpp
+++ b/src/widgets/favoritewidget.cpp
@@ -22,18 +22,18 @@
 #include <QSize>
 #include <QStyle>
 #include <QStylePainter>
+#include <QFontMetrics>
 
 #include "core/logging.h"
 
-const int FavoriteWidget::kStarSize = 15;
+const int FavoriteWidget::kStarSize = 16;
 
 FavoriteWidget::FavoriteWidget(int tab_index, bool favorite, QWidget* parent)
     : QWidget(parent),
       tab_index_(tab_index),
       favorite_(favorite),
       on_(":/star-on.png"),
-      off_(":/star-off.png"),
-      rect_(0, 0, kStarSize, kStarSize) {}
+      off_(":/star-off.png") {}
 
 void FavoriteWidget::SetFavorite(bool favorite) {
   if (favorite_ != favorite) {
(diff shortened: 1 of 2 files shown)


src/widgets/favoritewidget.cpp
```diff
@@ -22,18 +22,18 @@
...
-const int FavoriteWidget::kStarSize = 15;
+const int FavoriteWidget::kStarSize = 16;
 
 FavoriteWidget::FavoriteWidget(int tab_index, bool favorite, QWidget* parent)
     : QWidget(parent),
       tab_index_(tab_index),
       favorite_(favorite),
       on_(":/star-on.png"),
-      off_(":/star-off.png"),
-      rect_(0, 0, kStarSize, kStarSize) {}
```
narunlifescience: what about this variable in favoritewidget.h ? any way i don't think deleting this variable is a good idea..

src/widgets/favoritewidget.cpp
```diff
@@ -50,12 +50,16 @@ QSize FavoriteWidget::sizeHint() const {
 }
 
 void FavoriteWidget::paintEvent(QPaintEvent* e) {
-  QStylePainter p(this);
+  // ignore text descent when vertically centering
+  QFontMetrics fontMetrics = QFontMetrics(this->font());
+  QRect rect = this->rect();
+  rect.setBottom(rect.bottom() - fontMetrics.descent());
```
narunlifescience: If you put this code here, it will be called whenever paintEvent gets called, which is unnecessary. why not move this to the constructor ?

Ede123: I'm afraid the height of `this->rect()`is not yet defined (or at least does not have the final value) when calling the constructor.
I'm not too familiar with Qt and/or Clementine code myself but I'm open for suggestions if you have another idea...

src/widgets/favoritewidget.cpp
Ede123: I overlooked that one, removed. Thanks!

src/widgets/favoritewidget.cpp
narunlifescience: I made a pull request here https://github.com/Ede123/Clementine/pull/1
This way you can avoid call to this->rect() and move everything to constructor...

Ede123: I'm afraid your suggested code does not work:
* `Qt::AlignVCenter` will center the pixmap in the QRect provided to `drawItemPixmap()`. That's why we need `this->rect()` here (the drawing area of the widget or its height specifically) and can't just use some arbitrary rectangle.
* Your code produces a rectangle with a height of 14px on my system (since the font descent is 2 px) and therefore would even cut the icon (since `QStyle.drawItemPixmap()` won't rescale the pixmap).

As long as we do not want to make guesses regarding the widget height (which we probably could, given that we make a `sizeHint()`) which would be pretty bad style and prone to errors in the future we somehow have to get a valid height...

narunlifescience: Hmm.. you got a point. what about handling this in resize event like this ?
void FavoriteWidget::resizeEvent(QResizeEvent *) {
  rect_ = this->rect();
  rect_.setBottom(rect_.bottom() - QFontMetrics(this->font()).descent());
}

Ede123:
Great! @narunlifescience's suggestion with `QWidget::resizeEvent()` works nicely!

I updated the commit...

Issue #5178, opened 2015-12-30
Play (or add to playlist) by pressing Enter on Files


otoman:
In Colection you can add track to playlist (or just play song - depends on settings) by pressing Enter, but in Files you can't... Please add playing or adding by pressing Enter to Files.

ArnaudBienner:
Closing since #5180 has been merged, which fixes this issue.

PR #5180, opened 2016-01-02
Enable Enter key shortcut in file view

santigl:
When selecting a file from the file browser, pressing Enter now has the same effect than double-clicking it. (#5178)

### DIFF
--- a/src/widgets/fileview.cpp
+++ b/src/widgets/fileview.cpp
@@ -227,7 +227,10 @@ void FileView::keyPressEvent(QKeyEvent* e) {
     case Qt::Key_Backspace:
       ui_->up->click();
       break;
+    case Qt::Key_Enter:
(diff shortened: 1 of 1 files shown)


src/widgets/fileview.cpp
```diff
@@ -227,7 +227,12 @@ void FileView::keyPressEvent(QKeyEvent* e) {
     case Qt::Key_Backspace:
       ui_->up->click();
       break;
+    case Qt::Key_Enter:
+    case Qt::Key_Return:
+      ItemDoubleClick(ui_->list->currentIndex());
+      break;
   }
+}
```
ArnaudBienner: This doesn't compil, because of the extra } here

santigl:
Sorry about that. I tested it, but then I must have broken it before committing. I have now amended the commit.

PR #5182, opened 2016-01-04
Fix bug in edit tag dialog when fetching (2)

abika:
A fix for old tag values getting erased (instead of overwritten) in the edit tag dialog when fetching tags with MusicBrainz.

Replacement for pull request #5131 ; issue #3355 is still fixed.

[Maybe the `textEdited` signal can be connected instead of `textChanged` but I didn't want to break anything else. Specially for the next stable release.]

### DIFF
--- a/src/ui/edittagdialog.cpp
+++ b/src/ui/edittagdialog.cpp
@@ -99,7 +99,8 @@ EditTagDialog::EditTagDialog(Application* app, QWidget* parent)
         connect(widget, SIGNAL(Reset()), SLOT(ResetField()));
       }
 
-      // Connect the edited signal
+      // Connect the changed signal (emitted when value is changed
+      // programmatically or non-programmatically)
       if (qobject_cast<QLineEdit*>(widget)) {
         connect(widget, SIGNAL(textChanged(QString)), SLOT(FieldValueEdited()));
       } else if (qobject_cast<QPlainTextEdit*>(widget)) {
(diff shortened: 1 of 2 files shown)


fz72:
This fixed the bug for me.

src/ui/edittagdialog.cpp
```diff
@@ -833,27 +834,33 @@ void EditTagDialog::FetchTagSongChosen(const Song& original_song,
...
 
   // Find the song with this filename
-  auto data_it =
-      std::find_if(data_.begin(), data_.end(), [&filename](const Data& d) {
-        return d.original_.url().toLocalFile() == filename;
-      });
-  if (data_it == data_.end()) {
-    qLog(Warning) << "Could not find song to filename: " << filename;
-    return;
+  int id;
```
hatstand: Why are we going back to using an integer index instead of iterators now?

abika: For finding out if the song is currently displayed (see below).
(Otherwise a range-based for loop could be used here)

hatstand: You could still use an iterator and `std::distance(data_.begin(), it)`

abika: Is that really necessary? In the end its just a matter of personal preference but I want to produce code that is easy to understand and verifiable for everybody and does not include another new bug (which is the reason for this patch).

hatstand: The intent is clearer with an iterator.

abika: well, whatever. Reverted

PR #5193, opened 2016-01-09
Disable digitally imported for free user (premium account required)

Chocobozzz:
Digitally Imported stopped public streaming on 3rdparty clients to non premium users.

https://diradio.uservoice.com/forums/183400-digitally-imported-feedback/suggestions/11272656-restore-3rd-party-clients

Unfortunately I don't have a premium account so I can not test if this PR don't break premium access.


hatstand:
It's a good start. We should probably redirect users directly to the settings page too as soon as they try to expand one of these.

PR #5209, opened 2016-01-14
move duplicated code to annonymous namespace static function

narunlifescience:


hatstand:
Just put it all in a static function in an anonymous namespace in the .cpp file.

narunlifescience:
Done! now we wont be able to use const though..

src/widgets/fancytabwidget.cpp
```diff
@@ -55,6 +55,38 @@ using namespace Internal;
 const int FancyTabBar::m_rounding = 22;
 const int FancyTabBar::m_textPadding = 4;
 
+
+namespace {
+static void selectTab(QPainter *&p, const QRect rect) {
```
hatstand: `void SelectTab(QPainter* p, const QRect& rect) {`

narunlifescience:
Done!

Issue #3157, opened 2013-12-12
Equalizer details


Clementine-Issue-Importer:
_From [Aikino...@gmail.com](https://code.google.com/u/110277695185855645930/) on September 11, 2012 16:50:58_

its simple, but it would be good if, like in VLC, the EQ told you where its at currently. Check vlc's EQ to see what i mean, below each HZ there is the nummber of dB involved.
Its a minor detail that could be useful to most of us.

Cheers.

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=3156_

PR #5218, opened 2016-01-20
Add gain label to EqualizerSlider

santigl:
Display the selected gain (in dB) under each equalizer bar. (#3157)

### DIFF
--- a/src/widgets/equalizerslider.cpp
+++ b/src/widgets/equalizerslider.cpp
@@ -21,9 +21,25 @@
 EqualizerSlider::EqualizerSlider(const QString& label, QWidget* parent)
     : QWidget(parent), ui_(new Ui_EqualizerSlider) {
   ui_->setupUi(this);
-  ui_->label->setText(label);
+  ui_->band->setText(label);  // Band [Hz]
 
-  connect(ui_->slider, SIGNAL(valueChanged(int)), SIGNAL(ValueChanged(int)));
+  QFontMetrics fm = ui_->gain->fontMetrics();
+  int longestLabelWidth = fm.width(tr("%1 dB").arg(-99.99));
+  ui_->gain->setMinimumWidth(longestLabelWidth);
+  ui_->gain->setText(tr("%1 dB").arg(0));  // Gain [dB]
+
+  ui_->slider->setValue(0);
+
+  connect(ui_->slider, SIGNAL(valueChanged(int)), this,
+          SLOT(onValueChanged(int)));
+}
+
+void EqualizerSlider::onValueChanged(int value) {
+  // Converting % to dB as per GstEnginePipeline::UpdateEqualizer():
+  float gain = (value < 0) ? value * 0.24 : value * 0.12;
+
+  ui_->gain->setText(tr("%1 dB").arg(gain));  // Gain [dB]
+  emit ValueChanged(value);
 }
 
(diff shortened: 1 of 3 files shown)


hatstand:
This makes the size of the window jump around as the text width of the gain label changes.

santigl:
Thanks for noting that. I have managed to fix it, though first I wanted to ask you if it is the "correct" way of doing it.

The way I implemented the fix is:

1. Set the horizontal policy of the label to `QSizePolicy::Minimum`.
2. Using the label's `QFontMetrics.width()`, calculate the width of the string "-99.99 dB" in pixels (which is the longest that will ever be displayed).
3. Set the label's `MinimumWidth` to that size.

src/widgets/equalizerslider.cpp
```diff
@@ -21,9 +21,25 @@
...
     : QWidget(parent), ui_(new Ui_EqualizerSlider) {
   ui_->setupUi(this);
-  ui_->label->setText(label);
+  ui_->label->setText(label);  // Band [Hz]
 
-  connect(ui_->slider, SIGNAL(valueChanged(int)), SIGNAL(ValueChanged(int)));
+  QFontMetrics fm = ui_->label_2->fontMetrics();
+  int longestLabelWidth = fm.width("-99.99 dB");
+  ui_->label_2->setMinimumWidth(longestLabelWidth);
+  ui_->label_2->setText("0 dB");  // Gain [dB]
```
hatstand: This should really be translated

src/widgets/equalizerslider.cpp
```diff
@@ -21,9 +21,25 @@
 EqualizerSlider::EqualizerSlider(const QString& label, QWidget* parent)
     : QWidget(parent), ui_(new Ui_EqualizerSlider) {
   ui_->setupUi(this);
-  ui_->label->setText(label);
+  ui_->label->setText(label);  // Band [Hz]
 
-  connect(ui_->slider, SIGNAL(valueChanged(int)), SIGNAL(ValueChanged(int)));
+  QFontMetrics fm = ui_->label_2->fontMetrics();
+  int longestLabelWidth = fm.width("-99.99 dB");
```
hatstand: You need to use the translated version here too.

santigl:
Sorry about that. I amended the commit. Now it calls `tr()` on all strings.

src/widgets/equalizerslider.ui
```diff
@@ -74,6 +83,28 @@
    </item>
    <item>
     <widget class="QLabel" name="label">
```
hatstand: Can you give these two labels meaningful names?

santigl:
Sure. `label` was already there (displays the equalizer band frequency), so I followed that convention.

Do you want me to just modify `label_2` here and do a separate commit for the other one?

hatstand:
Change them both to something useful like `label_gain` and `label_band`

santigl:
Done.

src/widgets/equalizerslider.h
```diff
@@ -19,6 +19,7 @@
 #define EQUALISERSLIDER_H
 
 #include <QWidget>
+#include <QFontMetrics>
```
ArnaudBienner: Seems like you need this include only in the cpp file, so please move it there: it's better to include files only where necessary.

santigl:
I applied that change in #5251.

Issue #5197, opened 2016-01-11
Volume bar font has become grey


Yanpas:
Not sure, maybe it's a feature, but it looks like volume is muted
before
![_2016-01-11_16-58-18](https://cloud.githubusercontent.com/assets/7345761/12235429/b8bee1dc-b884-11e5-97ed-e2e7c3904acb.png)
after
![_2016-01-11_16-59-08](https://cloud.githubusercontent.com/assets/7345761/12235430/b8bf1918-b884-11e5-927e-aa70b94cf550.png)

TheUbuntuGuy:
This is a side effect of #5030 and #5035 which I mentioned previously. The colour depends on the system theme now.

fitojb:
The lighter outline looks much, much better, but the volume text should be the same color as the rest of the labels. It makes no sense for it to be lighter in color if the control is not disabled. Its small size already works as a de-emphasizing factor.

narunlifescience:
I made a pull request to make the volume text match normal window text (#5219). I don't want to change the volume slider outline. As @fitojb pointed out that looks even. There is probably another way to make the volume slider like before and still use system colours by taking the pixmap & overlaying the outline with system colours (some thing like #5042). I purposefully avoided this to make clementine look uniform.

narunlifescience:
volume font should be fine now..

PR #5219, opened 2016-01-21
use normal text color for volume

narunlifescience:
related to #5197


fitojb:
+1. Forcing the disabled appearance for an *enabled* control is crazy. Thanks!

PR #5228, opened 2016-01-30
Use https and not deprecated authentication with lastfm

Chocobozzz:
As said [here](http://www.last.fm/api/show/auth.getMobileSession), the `authToken` is deprecated. So we use directly the password in plain text, but force https.

### DIFF
--- a/src/internet/lastfm/lastfmservice.cpp
+++ b/src/internet/lastfm/lastfmservice.cpp
@@ -82,6 +82,8 @@ LastFMService::LastFMService(Application* app, QObject* parent)
       scrobbling_enabled_(false),
       connection_problems_(false),
       app_(app) {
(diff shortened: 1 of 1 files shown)


amuttsch:
Hm, I think we should use [desktopauth](http://www.last.fm/api/desktopauth) rather than mobileauth. It's a bit more work but the user doesn't have to enter the last.fm password into Clementine but on the last.fm website to gain access to last.fm.

Chocobozzz:
I completely agree with you but infortunately I don't have time to make this implementation :(

amuttsch:
Okay. I'm currently working on it, should be done today or tomorrow.

Chocobozzz:
Great! So we don't need this PR anymore.

PR #5243, opened 2016-02-07
Added support for Intergalactic FM

PanderMusubi:
This PR adds support for Intergalactic FM that works with the same channels.xml file as Soma FM does. At the moment the live channels.xml file doesn't get playing now information updated but that will happen soon. This support is at the moment also available in KODI, see http://addons.kodi.tv/show/plugin.audio.intergalacticfm/

There is only a minor problem. I think Clementine cannot handle the PLS files correctly for Intergalactic FM, while the same format for SomaFM does work. Can you advice on how to fix this?

Second question, I added functionality to use highest quality playing list if fast quality is not available. Should this also be added to the support for SomaFM and should also AAC be used if MP3 is not available? This will also provide more channels for SomaFM as not all offer streaming in MP3.


hatstand:
The playlists aren't working because instead of using the urls from the channels.xml you're hacking them around a bit by changing the scheme. If you request http versions of the pls urls (like you're currently doing) you get a 302 redirect which is not followed by default. HTTPS works just fine.

PanderMusubi:
I have changed the urls to HTTP in 1496d58 and the channels.xml has now also only HTTP. That is now also how they are being used in KODI. I have tested Clementine but, nevertheless, still no listing of channels in Clementine. Is there something else I am missing?

hatstand:
You're still generating the wrong urls in `IntergalacticFMUrlHandler`

PanderMusubi:
Could you be more specific?

All is HTTP and not HTTPS anymore.

hatstand:
IIRC, they need to be HTTPS

PanderMusubi:
Ah, perhaps some confusion.The implementation for SomaFM is all in HTTPS (because their HTTPS server is faster and sooner up to date). For Intergalactic FM, all should be in HTTP. So I changed all to HTTP. Can't see where it prevents showing the available channels in Clementine. Can you pinpoint it?

PanderMusubi:
Let me know if I can help or test a new build form your side.

hatstand:
https://builds.clementine-player.org/

PanderMusubi:
Super, this can be closed.

Issue #4774, opened 2015-03-10
Add ability to copy stream URL to clipboard


Beep6581:
I have a bunch of radio streams in a playlist, and it would be nice to be able to right-click on one and to copy its URL so I can share it. As is, its not only impossible to find the actual URL, but its even impossible to figure out where a stream in the playlist came from (Icecast? SKY.fm? SomaFM? other?) without manually searching for it. Again, this would be a great addition for a right-click context menu.

Please note, the URL should be the one directly to the stream, not some general one. E.g. right-click on "Cafe de Paris" in Internet > SKY.fm, and the closest option to getting a URL is "Open www.sky.fm in browser" but that's a dead end since you then have to search for that station again on their website and their website requires Adobe Flash to play a stream, which more and more people refuse to use for safety and other reasons.

jlsjonas:
+1 for adding copy URL support for services like soundcloud

coconutlulz:
I've been looking into this request. You can add the 'filename' column to your playlist display and this will show you both the URL and the service. Copying the URL might not help as some of them (e.g. Soundcloud) require the Clementine API key. While this would be copied, too, I'm not sure what restrictions there are on the Soundcloud API in terms of sharing it to be used from other applications.

If you are unhappy with the above workaround, I'll attempt to develop the feature and open a pull request.

Beep6581:
Thank you for looking into this.

I couldn't see the stream URLs, but found that when I replace "File name (without path)" with "File name" then they appear. This is good to know, but it would be far more practical to be able to right-click an entry and "Copy path" to clipboard, ending up with `/music/Mike Oldfield/2003 - Tubular Bells 2003/2 - Fast Guitars.ogg` or `http://streaming203.radionomy.com:80/OrabidooblogRadio` or whatever.

k3dar:
i too vote... please add "Copy Path/URL" to song context menu, or maybe "Show Information" for showing window with informations about song and with selectable&copyable text... thx :)

suther:
+10

jbauer180266:
On a related note: For Digitally Imported you show di:// domains, which aren't too useful. Replacing this prefix with http://pub7.di.fm/di_ and appending the name makes those URLs valid and much more useful, e.g. http://pub7.di.fm/di_deeptech [no warranty for linked page]

[...]

PR #5246, opened 2016-02-11
Allow copying of browser-ready URLs for Internet streams.

coconutlulz:
This is the first step in allowing users to directly copy browser-ready URLs which link to a selected stream via the context menu.

I have introduced a new QAction and QUrl in InternetService, so that all services can use this in future.

The new feature has been integrated into the Soundcloud, SomaFM, Icecast and Digitally Imported services.

While I have tested the Digitally Imported functionality previously, this was before a premium account was required. As such, I cannot modify my changes to be in line with what I have for the other services, as I would not be able to test such changes. Would someone with a premium account be so kind as to test this? A commit to tidy up the DI code (i.e. the if else) would be most welcome.

This is my first ever pull request for free software, so go easy! :-)

Issue #4774


hatstand:
Is there a plan for what to do for streams where we can't provide a browser-ready URL for some reason (e.g. Amazon)?

coconutlulz:
If we can't provide one for the service, I'm not sure what can be done other than the default (i.e. don't provide the option).

hatstand:
That's fine. As long as we don't provide URLs that don't work.

src/internet/digitally/digitallyimportedservicebase.cpp
```diff
@@ -181,11 +183,59 @@ void DigitallyImportedServiceBase::ShowContextMenu(const QPoint& global_pos) {
...
+
+void DigitallyImportedServiceBase::ShowContextMenu(const QPoint& global_pos) {
+  EnsureMenuCreated();
+  QStandardItem* item = model()->itemFromIndex(model()->current_index());
+
+  if (item) {
+    int type = item->data(InternetModel::Role_Type).toInt();
+
+    // Digitally Imported streams have a Role_Type of 0.
+    if (type == 0) {
```
coconutlulz: This conditional could be tidied up by replacing with `bool can_play`.

This bit of code was successfully tested **before** a premium account became mandatory, so any changes will need to be tested by someone with such an account.

hatstand:
Run `make format`

coconutlulz:
@hatstand Doesn't show me any problems; might be doing it incorrectly. Did I introduce incorrect formatting? Thanks.

hatstand:
It's still wrong in `DigitallyImportedServiceBase::ShowContextMenu` at least

PR #5263, opened 2016-02-18
moodbar: check QSlider max/min before using them in division

### LINKED ISSUE #5261: Crash when switching from a song w/omoodbar to a song w/ moodbar
I can reproduce reliably a crash that is related to the moodbar styling.  I built Clementine from master, and still see the issue.

I have a playlist with two items:

1. Internet stream: http://2qmtl0.akacast.akamaistream.net:80/7/953/177387/v1/rc.akacast.akamaistream.net/2QMTL0, for which clementine does not display a moodbar.
2. A mp3 file, but I suppose that any file for which a moodbar can be displayed would work too.

When I first play the internet stream, then switch (double click) on the mp3 file, Clementine crashes.

Steps to reproduce:

1. Open Clementine
2. Add the Internet stream mentioned above to the playlist
3. Start playing the Internet stream, if it does not do so automatically 
4. Enqueue an local music file
5. Double click on the local music file

It's at this point that Clementine crashes with a SIGFPE.

```
#0  0x0000000000c5fe23 in MoodbarProxyStyle::subControlRect (this=0x2e53a00, cc=QStyle::CC_Slider, opt=0x7ffcae353cb0, sc=QStyle::SC_ScrollBarSubLine, widget=0x2bccbc0) at /home/emaisin/src/clementine-player/src/moodbar/moodbarproxystyle.cpp:274
#1  0x00007fa80b565880 in QGtkStyle::drawComplexControl(QStyle::ComplexControl, QStyleOptionComplex const*, QPainter*, QWidget const*) const () from /usr/lib/x86_64-linux-gnu/libQtGui.so.4
#2  0x0000000000c5f934 in MoodbarProxyStyle::Render (this=0x2e53a00, control=QStyle::CC_Slider, option=0x7ffcae354190, painter=0x7ffcae354250, widget=0x2bccbc0) at /home/emaisin/src/clementine-player/src/moodbar/moodbarproxystyle.cpp:200
#3  0x0000000000c5f704 in MoodbarProxyStyle::drawComplexControl (this=0x2e53a00, control=QStyle::CC_Slider, option=0x7ffcae354190, painter=0x7ffcae354250, widget=0x2bccbc0) at /home/emaisin/src/clementine-player/src/moodbar/moodbarproxystyle.cpp:171
```

The following computation is done here:
https://github.com/clementine-player/Clementine/blob/master/src/moodbar/moodbarproxystyle.cpp#L274

```
const int x = (slider_opt->sliderValue - slider_opt->minimum) *
                        (opt->rect.width() - kArrowWidth) /
                        (slider_opt->maximum - slider_opt->minimum);
```

At this point, slider_opt->maximum and slider_opt->minimum are both 0, so it ends up in a division by 0.  0 is the default value for both of those fields, so my guess is that this is called before the widget is properly initialized.

It is possible to workaround this crash by disabling the moodbar.  I also managed to workaround it by adding a 

```
          if (slider_opt->maximum <= slider_opt->minimum)
        	  break;
```

just before the line that ends up dividing by 0.

EDIT:

I forgot to add version details:
- Ubuntu 14.04
- Qt 4:4.8.5+git192-g085f851+dfsg-2ubuntu4.1

simark:
When switching playback from an item for which the moodbar is not
displayed (e.g. an internet stream), to an item for which it is
displayed (e.g. a local mp3 file), Clementine sometimes crashes.

This happens because the slider_opt->maximum and slider_opt->minimum
have the value 0 (their default value), and the difference is used as a
divisor.  This gives a division by 0, and a SIGFPE.

This problem has been fixed in commit af42cce.  However, when a9f9b0e
reverted 3f79fa5, a little to much was reverted and we lost what af42cce
did.  This patch re-introduces the fix.

Fixes #5261

### DIFF
--- a/src/moodbar/moodbarproxystyle.cpp
+++ b/src/moodbar/moodbarproxystyle.cpp
@@ -270,10 +270,15 @@ QRect MoodbarProxyStyle::subControlRect(ComplexControl cc,
         case SC_SliderHandle: {
           const QStyleOptionSlider* slider_opt =
               qstyleoption_cast<const QStyleOptionSlider*>(opt);
+          int x = 0;
 
-          const int x = (slider_opt->sliderValue - slider_opt->minimum) *
-                        (opt->rect.width() - kArrowWidth) /
-                        (slider_opt->maximum - slider_opt->minimum);
+          /* slider_opt->{maximum,minimum} can have the value 0 (their default
+             values), so this check avoids a division by 0. */
+          if (slider_opt->maximum > slider_opt->minimum) {
+            x = (slider_opt->sliderValue - slider_opt->minimum) *
+                (opt->rect.width() - kArrowWidth) /
+				(slider_opt->maximum - slider_opt->minimum);
+          }
 
           return QRect(QPoint(opt->rect.left() + x, opt->rect.top()),
                        QSize(kArrowWidth, kArrowHeight));


hatstand:
Given the trouble this seems to have caused, can you add an explanatory comment?

simark:
Sure, good idea.  Do you prefer if I add it as a new commit, or if I force push?

hatstand:
Just add it to this pull request and squash the commits

simark:
Ok, commit updated.

Issue #695, opened 2013-12-06
16k is inverted in equalizer


Clementine-Issue-Importer:
_From [himynameiszacHandiamapirate](https://code.google.com/u/116275170679458301339/) on September 05, 2010 04:50:54_

Never thought I'd see the day. Maybe it's just my setup (though I tested on speakers and headphones), but give this a shot: Go into the equalizer while anything is playing, and adjust the 16k slider up and down. On my system, sound is actually clearer with the slider all the way down... much clearer. Pumping up the 16k seems to actually be creating some interference or something, for me. So, assuming this is happening on other machines, either the equalizer we're using is incorrectly adjusting this band, or it really is just way too high for us to be messing with.

Thoughts continue below, on the assumption this needs to be changed, and isn't just happening for me.

I'm not really sure what's to be done here... a nine-band equalizer would be kinda odd, and I definitely use the other bands (if only to adjust the low end so my subwoofer doesn't eat me... damned lack of a subwoofer volume control). Thoughts? /am I crazy and only experiencing this on my end?

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=695_

ArnaudBienner:
Now we migrated to Gstreamer 1.0, can someone try again to see if the problem is gone?

dancemasterglenn:
I got the same result from the most recent compiled windows build... but maybe I had some leftover config files from my previous install? This doesn't seem like the kind of issue that would carry over in that way, though. :(

Can anyone else test and confirm?

KitsuneRale:
I'm still having this on my Windows machine, with recent nightly builds.

santigl:
The bug is still there with GStreamer 1.0 in Debian 64 bits.

However, I have been playing with the values, and noticed that the behavior could also be replicated with the first band. 

It turns out that for the first and last bands GStreamer uses shelve filters, calling different routines than for the rest of the frequencies.

So, what I did as a workaround (assuming that there is a bug in GStreamer's implementation of those corner cases), is to generate two dummy bands at each end of the spectrum. They are dummies because their bandwidth is set to 0, and they are centered at 20Hz and 20000Hz.

While doing that, I also tried moving the bands to follow the [ISO 266](http://www.zytrax.com/tech/audio/equalization.html#iso-frequencies) standard. 

[...]

PR #5277, opened 2016-03-02
 Add dummy bands to avoid #695. Change frecuencies to be ISO 266 compliant

santigl:
Assuming there is a bug in GStreamer's implementation of the first and last bands of the equalizer, we now set up two dummy bands, cutting from 20Hz and 20KHz, and with a bandwidth of 0 (what makes them dummies). That forces all actual bands to be implemented with band-pass filters, avoiding GStreamer's possibly-buggy shelve-filters implementation and preventing #695.

Also, we make bands compliant with ISO standard 266.

### DIFF
--- a/src/engines/gstenginepipeline.cpp
+++ b/src/engines/gstenginepipeline.cpp
@@ -41,8 +41,10 @@ const int GstEnginePipeline::kGstStateTimeoutNanosecs = 10000000;
 const int GstEnginePipeline::kFaderFudgeMsec = 2000;
 
 const int GstEnginePipeline::kEqBandCount = 10;
-const int GstEnginePipeline::kEqBandFrequencies[] = {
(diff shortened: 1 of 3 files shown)


TheUbuntuGuy:
I love that this is finally being done. This will however affect the preset equalizers in `src/ui/equalizer.cpp`. Are there plans to update them according to the new bands using some kind of conversion curve? As it stands they will sound very different (with the 12k and 14k bands now being 4k and 8k respectively).

Ferroin:
I'm not 100% certain that the curves should be changed to account for this.  Things will sound different of course, but the curves exactly match a number of other pieces of software I've seen that use the ISO 266 frequency bands instead of what we are using, so they may have been wrong in the first place.

If we do decide to update the curves, then the same type of interpolation that gets used for re-sampling audio should theoretically work just fine for converting the curves, but we'd have to redo the entire curve, as all the bands are different except for 1k, which is getting moved right one slot.

PR #5280, opened 2016-03-07
Require Qt4.8.6 and CMake deprecate warnings

JulianVolodia:
"com.trolltech.*" changed to "org.qtproject.*"


hatstand:
Ubuntu Precise is still on 4.8.1

JulianVolodia:
@hatstand, we could leave there 4.8.1 and fix that deprecation, what you think?

hatstand:
I think 4.8.1 would be fine

JulianVolodia:
Merged #5287 .

PR #5292, opened 2016-03-11
Fix mpris dbus notifications

printesoi:
Clementine does not send the correct signal about the changed properties such as CanGoNext, CanGoPrevious, CanSeek and may cause programs that listen to those changes use the wrong information. 

To check the corect signals are send one can use:

`dbus-monitor --session --monitor "type='signal',interface=org.freedesktop.DBus.Properties,member=PropertiesChanged,path=/org/mpris/MediaPlayer2"`

### DIFF
--- a/src/core/mpris2.cpp
+++ b/src/core/mpris2.cpp
@@ -171,6 +171,12 @@ void Mpris2::EmitNotification(const QString& name) {
     value = Volume();
   else if (name == "Position")
     value = Position();
+  else if (name == "CanGoNext")
+    value = CanGoNext();
+  else if (name == "CanGoPrevious")
+    value = CanGoPrevious();
+  else if (name == "CanSeek")
+    value = CanSeek();
 
   if (value.isValid()) EmitNotification(name, value);
 }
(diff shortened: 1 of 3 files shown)


src/core/mpris2.cpp
```diff
@@ -404,6 +421,14 @@ bool Mpris2::CanSeek() const {
   }
 }
 
+bool Mpris2::CanSeek(Engine::State state) const {
+    if (mpris1_->player()) {
```
hatstand: Use a ternary expression

hatstand:
Can you also run `make format`?

printesoi:
@hatstand I did run `make format`, I get:

    src/core/mpris2.cpp: formatting is correct!
    src/core/mpris2.h: formatting is correct!
    src/playlist/playlist.cpp: formatting is correct!

printesoi:
@hatstand I ran `make format`

src/core/mpris2.cpp
printesoi: Fixed.

PR #5314, opened 2016-03-26
Move global search settings loading later in "MainWindow" initialisation

### LINKED ISSUE #5312: Search pane only shows local library as a source
### Before posting

- [x] I tried the latest Clementine build from [here](https://builds.clementine-player.org/) - *I [build from git](https://aur.archlinux.org/packages/clementine-git) instead.*

### System information

- Operating System: Arch Linux
- Clementine version: 23c0fe9

The search pane only shows the local library as a source, despite others existing (both enabled and disabled).  This appears to be merely a visual issue, as I still get results from the secondary sources when searching.

1. If no other search sources are enabled, enable them from preferences (User interface -> Search).
1. Restart clementine.
1. Observe the search pane: 
image
1. Revisit the search preferences and hit 'ok' without changing anything.
1. Observe the search pane again:  
image

ivan-leontiev:
Fixes #5312.

To be able to populate the sources list `GlobalSearch` requires `InternetModel` to be initialised before. The issue #5312 started happening since the commit f2daa772c1090e866cacb7fce68a9240ffe3e766.

### DIFF
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -251,9 +251,6 @@ MainWindow::MainWindow(Application* app, SystemTrayIcon* tray_icon, OSD* osd,
       app_->library_backend(), tr("Library"), "library",
       IconLoader::Load("folder-sound", IconLoader::Base), true, app_, this));
 
-  app_->global_search()->ReloadSettings();
-  global_search_view_->ReloadSettings();
-
   connect(global_search_view_, SIGNAL(AddToPlaylist(QMimeData*)),
           SLOT(AddToPlaylist(QMimeData*)));
 
@@ -987,6 +984,11 @@ MainWindow::MainWindow(Application* app, SystemTrayIcon* tray_icon, OSD* osd,
   ui_->splitter->setChildrenCollapsible(false);
 
   ReloadSettings();
+
+  // The "GlobalSearchView" requires that "InternetModel" has already been
+  // initialised before reload settings.
+  app_->global_search()->ReloadSettings();
+  global_search_view_->ReloadSettings();
 
   // Reload pretty OSD to avoid issues with fonts
   osd_->ReloadPrettyOSDSettings();


narunlifescience:
:+1:

hatstand:
Can you add a comment explaining why those statements have to be in that position?

amuttsch:
Hm, this change doesn't really fix this issue but only makes it less common, does it? No checks are implemented if `InternetModel` is available at this point or am I overlooking something? It would still be a matter of timing if this issue occurs or not.

ivan-leontiev:
Actually those statements don't have to be placed in that particular line, rather it would be enough to place them just after the line where we have access the `InternetModel` first time, since we lazy initialise it(e.g. just after [this](https://github.com/clementine-player/Clementine/blob/ca487a51663472fa443e5d33748fb3c65b720713/src/ui/mainwindow.cpp#L750) line). The rationale to place them there was that after a look into the [`MainWindow::ReloadAllSettings()`](https://github.com/clementine-player/Clementine/blob/ca487a51663472fa443e5d33748fb3c65b720713/src/ui/mainwindow.cpp#L1071) method, you can see that `global_search()->ReloadSettings()` call is under the `MainWindow::ReloadSettings()` (actually it does not matter, but just for consistency look).

But more interesting is why we need to care about that order. We should look a bit deeper in `InternetModel` initialisation, there we register our internet services via the `InternetModel::AddService()` method [here](https://github.com/clementine-player/Clementine/blob/ca487a51663472fa443e5d33748fb3c65b720713/src/internet/core/internetmodel.cpp#L89-L122). We instantiate each service and each service registers its globalsearch provider by calling `GlobalSearch::AddProvider()` method. Thereby we have populated `GlobalSearch`'s `providers_` map. 

If we reload settings too early then the `providers_` is not populated by internet services and the `GlobalSearchView` displays only a local library search provider which we have registered before [here](https://github.com/clementine-player/Clementine/blob/ca487a51663472fa443e5d33748fb3c65b720713/src/ui/mainwindow.cpp#L250).

hatstand:
Yes, we should probably fix it properly; sort of what I was hinting at ;-)

xiongchiamiov:
Thanks, and confirming I see #5312 fixed on my end as of ed13d02231ba6b6cc1d02239edac59fb71b55750.

Issue #3584, opened 2013-12-13
.nomedia support


Clementine-Issue-Importer:
_From [ozone89.DP](https://code.google.com/u/ozone89.DP/) on March 29, 2013 01:51:01_

As mobile devices are getting more and more diffused for listening to music, and clementine being able to transfer to/from this equipment, I think it should be useful to have .nomedia files support.
(right now I have the music files from the app "beats" being listed into my phone's library into clementine when connected to my PC, LOL)

little background facts:
What are .nomedia files?

a .nomedia file is a hidden file that makes a media scanner (like the one found on Android) to ignore the content of the folder where the file is placed

example:

music
|
+--dir 1
|  +--1.mp3
|  +--2.flac
|  +--subdir 1
|     +--.nomedia
|     +--7.ogg
|
+--dir 2
|  +--.nomedia
|  +--3.ape
|  +--4.mp3
+--5.ogg
+--6.wav

in this case, when the content of the "Music" directory gets scanned, the whole content or "dir 2" and "subdir 1" is ignored and not being listed in the library because of the .nomedia file, causing "3.ape", "4.mp3" and "7.ogg" to be missing from DB

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=3584_

mildred:
I'd also add support for a `.nomusic` file. There are some directories that might contain media files but no music and we don't want Clementine to scan those.

I have a practical use case for this. I have a 1TB hard drive that contain music and many other things. Currently, clementing is stuck at 99% reading this device, and I can't use it.

Implementation for this could be very simple. It happens in `LibraryWatcher::ScanSubdirectory` (`src/library/librarywatcher.cpp`). If we detect a `.nomedia` file, we just return early.

PR #5327, opened 2016-04-08
Don't scan directories with .nomedia or .nomusic files

mildred:
Should solve issue #3584. I am currently testing it out on my large hard drive that contain music and many other things (that slows down the crawling)

### DIFF
--- a/src/library/librarywatcher.cpp
+++ b/src/library/librarywatcher.cpp
@@ -43,6 +43,11 @@
 #undef RemoveDirectory
 #endif
 
+namespace {
+static const char *kNoMediaFile = ".nomedia";
+static const char *kNoMusicFile   = ".nomusic";
+}
+
 QStringList LibraryWatcher::sValidImages;
 
 const char* LibraryWatcher::kSettingsGroup = "LibraryWatcher";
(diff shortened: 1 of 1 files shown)


src/library/librarywatcher.h
```diff
@@ -26,6 +26,9 @@
 #include <QStringList>
 #include <QMap>
 
+#define LIBRARYWATCHER_NOMEDIA_FILE ".nomedia"
```
hatstand: Put these in the cpp file and change them to:
```c++
namespace {
static const char* kNoMediaFile = ".nomedia";
...
```

mildred:
By the way, it works really well on a slow computer with a hard disk containing lots of other files.

src/library/librarywatcher.cpp
```diff
@@ -43,6 +43,11 @@
 #undef RemoveDirectory
 #endif
 
+namespace {
+static const char *kNoMoMediaFile = ".nomedia";
```
hatstand: `kNoMediaFile`

mildred:
I fixed my commit according to your comment

PR #5345, opened 2016-04-19
Add Favorites support to SoundCloud

Vavooon:


src/internet/soundcloud/soundcloudservice.cpp
```diff
@@ -137,6 +137,10 @@ void SoundCloudService::EnsureItemsCreated() {
                           InternetModel::Role_PlayBehaviour);
     root_->appendRow(user_tracks_);
 
+
```
ArnaudBienner: Don't need to add another empty line here

src/internet/soundcloud/soundcloudservice.h
```diff
@@ -112,6 +114,7 @@ class SoundCloudService : public InternetService {
   QStandardItem* user_tracks_;
   QStandardItem* user_playlists_;
   QStandardItem* user_activities_;
+  QStandardItem* user_favorites_;
```
ArnaudBienner: This should be init to nullptr in the constructor, like other POD types

PR #5348, opened 2016-04-19
Qt5

davidgfnet:
Sorry I forgot some other fixes.
Still I'm facing issues when the server works over HTTPS instead of old plain HTTP, I will probably come up with a new patch soon. BTW --verbose doesn't really work for me, nothing out the stdout/stderr, any ideas?


davidgfnet:
I just found the problem, turns out that SSL does strict certificate validation, but only for audio streams, not for the actual SubSonic song/album enumeration. Anyone can point me to the right place please? I think we might already have an option to ignore SSL certificates in case we want to.

PR #5370, opened 2016-05-09
[Qt5] Fix spotify blob config path

Chocobozzz:
If we compile spotify blob ourselves.


FadeMind:
@Chocobozzz 
`clementine-player-qt5.org` is valid domain?

Chocobozzz:
@FadeMind Thanks, it's fixed.

PR #5374, opened 2016-05-11
Typo

hartgen:
Fixed typo in German translation.


hatstand:
Please fix this on transifex instead: https://www.transifex.com/davidsansome/clementine/

PR #5434, opened 2016-07-07
Fix fetching cover art for taglib >= 1.11

beedaddy:
TagLib::Ogg::XiphComment now handles METADATA_BLOCK_PICTURE separately from other fields. Thus, we have to use XiphComment::pictureList() in order to get cover art.


hatstand:
Please run `make format`

beedaddy:
@hatstand Thanks for the hint. Done.

PR #5443, opened 2016-07-16
Fix compilation with Clang

### LINKED ISSUE #4227: Building with clang

Fat-Zer:
Another try to fix  #4227
Also this fixes several warnings about missing overrides. And a messed up include guard.

Previous PR had troubled building chromaprint which is currently removed.
I've test myself that it builds fine with both clang-3.7.1 and gcc-4.9.3 on gentoo linux. But this one still requires some additional testing.

### DIFF
--- a/tests/mock_networkaccessmanager.h
+++ b/tests/mock_networkaccessmanager.h
@@ -15,7 +15,7 @@
    along with Clementine.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-#ifndef MOCK_NETWORKACCESSAMANGER_H
+#ifndef MOCK_NETWORKACCESSMANAGER_H
 #define MOCK_NETWORKACCESSMANAGER_H
 
 #include <QByteArray>
--- a/3rdparty/vreen/vreen/cmake/CommonUtils.cmake
+++ b/3rdparty/vreen/vreen/cmake/CommonUtils.cmake
@@ -92,7 +92,7 @@ macro(UPDATE_COMPILER_FLAGS target)
 
     if(FLAGS_CXX11)
         update_cxx_compiler_flag(${target} "-std=c++0x" CXX_11)
-	update_cxx_compiler_flag(${target} "-stdlib=libc++" STD_LIBCXX)
+        #update_cxx_compiler_flag(${target} "-stdlib=libc++" STD_LIBCXX)
         #add check for c++11 support
     endif()
 
--- a/ext/clementine-spotifyblob/CMakeLists.txt
+++ b/ext/clementine-spotifyblob/CMakeLists.txt
@@ -7,7 +7,7 @@ include_directories(${CMAKE_SOURCE_DIR}/ext/libclementine-spotifyblob)
 include_directories(${CMAKE_SOURCE_DIR}/ext/libclementine-common)
 include_directories(${CMAKE_SOURCE_DIR}/src)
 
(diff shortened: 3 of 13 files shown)


tests/mock_networkaccessmanager.h
```diff
@@ -15,7 +15,7 @@
    along with Clementine.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-#ifndef MOCK_NETWORKACCESSAMANGER_H
+#ifndef MOCK_NETWORKACCESSMANAGER_H
```
ArnaudBienner: :+1:  good catch!

ArnaudBienner:
I'm not sure what removing  -U__STRICT_ANSI__ implies, nor does the change you did in vreen.
Regarding the change in the header, good catch: worth to merge that. And the override you added is worth to merge too, though this clang's specific warning doesn't sound like a big deal to me: IIRC it warns only when other methods have a override keyword, but doesn't when none have (while this is arguably a good reason to have a warning too) and FWIW I don't like changing code when not needed.

Fat-Zer:
I'm pretty sure about changes in vreen: For a linux box there is no sense to default a system library to libc++ due to absolute most of linux boxes doesn't have it at all and definitely it is not desirable to be a default one. Apple users should be safe here because clementine adds the same flag for them in the toplevel CMakeLists.txt.

I'm not sure there -U__STRICT_ANSI__ came from... may be it was required during the migration to c++11 and prevented build failures because those times clementine used some deprecated features or whatever... not sure for it. But it seems now it's not really needed anymore.

>IIRC it warns only when other methods have a override keyword, but doesn't when none have (while this is arguably a good reason to have a warning too)

yes it is, this behavior is triggered by -Winconsistent-missing-override. If you want override to be strictly required everywhere, you may pass -Wsuggest-override. Both was introduced in clang 3.7.

> Regarding the change in the header, good catch: worth to merge that.

This one was also noticed by clang.

Fat-Zer:
There is no response for quite a while on this PR, so what's about it?

Note, If there will be some problems cosed by the patch, I'm wondering to actively participate in resolving them as far as I can.

hatstand:
Compiles fine on all our platforms according to buildbot (including mac which is using clang 3.9) so not inclined to merge this.

Fat-Zer:
@hatstand, just to make sure, does the build bot compiles with clang on a linux?

hatstand:
I think all the native Linux builds are gcc, but the mac buildbot is actually clang on Linux.

Fat-Zer:
> I think all the native Linux builds are gcc, but the mac buildbot is actually clang on Linux.

Not quite true. There is at least one (although it seems to be the only one by now) distribution claimed to be compiled with clang: [openmandrive-3.0](https://blog.openmandriva.org/2016/08/openmandriva-lx-3-0-final-release-is-out/). Also gentoo supports ability to build packages (or more or less even the whole system) with clang (actually it's the reason I'm concerned about this). The debian guys more or less regularly [try to build](http://clang.debian.net/) the debian package archive with clang too (more for testing/PR purpose IMO); The clementine's failure build log may be seen [here](https://github.com/opencollab/clang.debian.net/blob/master/logs/2016-08-30/clementine_1.3.1%2Bdfsg-1_unstable_clang.log). Besides those examples I believe that there are developers which use clang as a dev compiler (either as part of IDE or for test builds).

I don't know a lot about specific of building for MacOS, but I'm pretty sure it uses at least different libc/libcxx. So the fact that the build is run on linux doesn't really mean anything.

After all I'm pretty sure that all patches here except "Remove -U__STRICT_ANSI__ from CXXFLAGS" defiantly won't hurt the build in any way. If you wish I may provide them in a separate PR for merging and keep discussing the last one, awaiting the feedback from users building for mac.

Issue #5445, opened 2016-07-17
Playlist sorting by Filename / filename (without path) erroneously


kekkc:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [X] I checked the issue tracker for similar issues
- [X] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [X] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Win (OS independent)
- Clementine version: 1.3.1

### Expected behaviour / actual behaviour
just discovered Clementine (amazing player and much smaller on Windows than Amarok), but I realized a strange behavior when new files are added to the playlist. If a folder "My Music" is added that contains a folder "Brooklyn", Clementine displays it as the following:

1. "My Music" songs starting with "A" are displayed first
2. "My Music" songs starting with "B" are displayed afterwards
3. "Brooklyn" songs starting with "A" are displayed afterwards
4. "My Music" songs starting with "C" are displayed afterwards

Usually players like Winamp, Foobar etc.. will display logically sorted based on the folder structure:

1. "My Music" songs starting with "A" are displayed first
2. "My Music" songs starting with "B" are displayed afterwards
3. "My Music" songs starting with "C" are displayed afterwards (and all remaining files in "My Music")
4. "Brooklyn" songs starting with "A" are displayed afterwards

All files within a folder are displayed first, then the subfolders files & folders are displayed

### Steps to reproduce the problem (only for bugs)
1. Add & folder with subfolders to the playlist
2. Sort by filename

santigl:
Thanks for reporting this bug. I can confirm that it is not working properly.

So far I could think of a solution that would return a *breadth-first* order, at least implemented the way sorting functions are now (by defining a less-than comparison function).

A sample output would be:
* /Music/a.mp3
* /Music/c.mp3
* /Music/Brooklyn/b.mp3
* /Music/Brooklyn/z.mp3
* /Music/Charlie/a.mp3
* /Music/Charlie/e.mp3
* /Music/Brooklyn/X/l.mp3

[...]

PR #5452, opened 2016-07-25
Ignore case in path comparison (#5445)

santigl:
This makes sorting by `filename` (path) return the expected result.

For example:
* /Music/a.mp3
* /Music/c.mp3
* /Music/Brooklyn/b.mp3
* /Music/Brooklyn/z.mp3
* /Music/Charlie/a.mp3
* /Music/Charlie/e.mp3
* /Music/Brooklyn/X/l.mp3


ArnaudBienner:
Looks good to me.
Thanks for the patch! :)

Issue #777, opened 2013-12-06
copy album covers to portable devices (e.g. ipods)


Clementine-Issue-Importer:
_From [multicolor.mood](https://code.google.com/u/107644531612457542952/) on September 20, 2010 23:17:01_

Are you planning to add covers' support for ipod?

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=777_

JulianVolodia:
@Chocobozzz , @hatstand is it actual still?

blokant:
I have fixed it. I did a pull request with my changes to clementine/qt5. I am not sure should i create a similar  pull request for the master branch but that patch will work on the master branch too

blokant:
https://github.com/clementine-player/Clementine/pull/5477

blokant:
and yes: it works only with embedded covers, but i am working on the covers from folders

JulianVolodia:
So it's still there but only for *covers from folders*. Any explanation where could be buggy part? This week I probably have more time for contributions.

blokant:
Well, this patch is only about embedded to the file covers. Or what do you mean saying "covers from folder"

lesebas:
Hello,

Importing cover in my Ipod nano does not work for me even after exporting covver on my disk.

jernejZup:
will this ever be added?

PR #5477, opened 2016-08-27
Cover art transferes to the iPod and displayed appropriately(issue #777)

blokant:
Now cover art embedded to your files will be transfered appropriately to your iPod


blokant:
If it is better to do this PR on the Master branch say so, i see no activity here, may be this port is dropped...

Issue #5483, opened 2016-09-07
[Feature Request] Expanded Playlist Tabs


wapisani:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [x] I checked the issue tracker for similar issues
- [x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [ ] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Linux Mint 17.3
- Clementine version: 1.3.1

Feature Request:

I would like the capability to expand the playlist tabs so that the full names are visible. I have a lot of playlists loaded and it's annoying to have to click around until I find the playlist I want. Since there is already the capability to scroll through the playlists, it shouldn't be a big deal to make the tabs longer.

Thanks,
Will

PR #5487, opened 2016-09-12
Add ability to show full playlist name in PlaylistTab. #5483

selumin:


hatstand:
Please follow the style guide: https://google.github.io/styleguide/cppguide.html

selumin:
Fixed and checked with cpplint. Looks like it's correct now.

PR #5506, opened 2016-10-03
Fix undefined symbol error conversion from std::string to TagLib::String

izissise:
I had some undefined symbols errors:
```
undefined reference to `TagLib::String::String(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, TagLib::String::Type)'
```
This pr fix it.
I had the same errors on master, might be a good idea to cherry pick ;)


hatstand:
I'd prefer this were applied to master instead.

Issue #1079, opened 2013-12-06
Make CPU usage lower


Clementine-Issue-Importer:
_From [madura.x86](https://code.google.com/u/100668712151901585546/) on December 12, 2010 17:32:21_

Clementine is using too much CPU for a player(when considered with Amarok 1.4 and IMHO). It uses about 12&#37; of CPU when playing a MP3 without any visualization. I think this should improve, other than this it's great!

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=1079_

bwat47:
I've used clementine on both yosememite and mavericks on my mba 2013 (both clean installs). Clementine uses absurd amount of cpu while playing music on both. when I'm playing a song it just sits there using constant 20-30% cpu usage at all times.

JulianVolodia:
@xgdgsc , @bwat47 is it actual still?

bwat47:
Yes, I still see this with clementine 1.2.2 on OSX 10.11.3

vukasin0:
still issue on latest version on El Cap.

Even if it is idle it consume arround 17 - 20 % of cpu on MBP mid 2015

ginsterbusch:
Still surfacing once in a while, even with the latest nightly. 

System: Thinkpad T520, i5, 8 GB RAM
OS: Linux Mint 17.2 / x86_64
Kernel: 3.16

cu, w0lf.

JulianVolodia:
Any idea why could it be? Some plugins, cores? Maybe some scrobble timer?

ginsterbusch:
No plugins (are there still any around?) installed or enabled. Happens with or without enabled last.fm. Mostly without, because I dont have it set up for clementine (doesnt make sense; I use clementine mostly for quick prelistening songs and testing whether that nasty cue file worked out or not; MoC is my main player).

cu, w0lf.

JulianVolodia:
Even if I do not scrobble could notice:

> 01:54:31.227 DEBUG MainWindow:1473                  position 0 scrobble point 31 status 0 
01:54:41.227 DEBUG MainWindow:1473                  position 10 scrobble point 31 status 0 
01:54:51.227 DEBUG MainWindow:1473                  position 20 scrobble point 31 status 0

JulianVolodia:
I notice sth about 5% and only when window is open. Could you agree that CPU usage is lower when active "in tray" and window is closed?

iHile:
up!
10% CPU if window closed. 
[...]

PR #5511, opened 2016-10-10
Reduce some CPU load by disabling/fixing some UI stuff

davidgfnet:
- Disabling playlist row glow by default
- Track slider popup visible by default, add hide()
  to ensure not visible by default (no more repaints)
- Set kTrackSliderUpdateTimeMs to 500. IMO 40 is too low

This should partially fix issue #1079

### DIFF
--- a/src/playlist/playlistview.cpp
+++ b/src/playlist/playlistview.cpp
@@ -120,7 +120,7 @@ PlaylistView::PlaylistView(QWidget* parent)
       last_height_(-1),
       last_width_(-1),
       force_background_redraw_(false),
-      glow_enabled_(true),
+      glow_enabled_(false),
       currently_glowing_(false),
       glow_intensity_step_(0),
       rating_delegate_(nullptr),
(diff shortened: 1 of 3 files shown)


kernc:
Any reason this hasn't gotten the attention it deserves? Seems to apply cleanly on master as well as qt5 branch, and works exactly as observed by the last few commenters in https://github.com/clementine-player/Clementine/issues/1079. 😄

rickysarraf:
Thanks for these changes. They make clementine much power efficient.

raphj:
I just tried compiling clementine-qt5 with this pull request and this halves CPU usage (from 17-20% to 8-9%) when the window is shown. This makes Clementine CPU usage the almost the same when hidden or shown.

The slider animation is less smooth, but this is not a problem as me. I'd rather have an energy efficient player than smooth slider animation.

kernc:
60 FPS smooth slider, that's Western values right there. 😆

davidgfnet:
I'd gladly make any amendments so it can get pushed into the official repo, I think this adds value to the program.

davidgfnet:
Just set up a copr repo for any Fedora users interested.
https://copr.fedorainfracloud.org/coprs/davidgf/Clementine/

Will try to do the same for qt5 branch, since I use that one with my patches

davidgfnet:
Just sent the patch to fedora pkg mantainer. In theory it should get there next release. Yay! Its fedora only though, lets hope it gets merged later.

PR #5514, opened 2016-10-14
Clarified translation for "Scroll over icon...

ghandmann:
...to change track"

The translation for this config entry was rather unhelpfull and not consistent in german.


hatstand:
Please fix on transifex instead otherwise it will get overwritten: https://www.transifex.com/davidsansome/clementine

PR #5519, opened 2016-10-20
Fix Subsonic endpoint

nicolasmartinelli:
According to the documentation, the endpoint is `getArtistInfo`, not
`GetArtistInfo`. This typo makes the complete feature not working at
all.

Source: http://www.subsonic.org/pages/api.jsp#getArtistInfo


hatstand:
The commit doesn't match your description

PR #5520, opened 2016-10-20
Fix Subsonic endpoint

nicolasmartinelli:
According to the documentation, the endpoint is `getAlbumList`, not
`GetAlbumList`. This typo makes the complete feature not working at
all.

Source: http://www.subsonic.org/pages/api.jsp#getAlbumList


nicolasmartinelli:
Thanks

PR #5553, opened 2016-12-08
Fix compilation with chromaprint >= 1.4

Chocobozzz:
See https://github.com/acoustid/chromaprint/blob/master/NEWS.txt#L7

I did not find a more elegant solution, I'm open to suggestions.

### DIFF
--- a/src/musicbrainz/chromaprinter.cpp
+++ b/src/musicbrainz/chromaprinter.cpp
@@ -143,16 +143,24 @@ QString Chromaprinter::CreateFingerprint() {
   ChromaprintContext* chromaprint =
       chromaprint_new(CHROMAPRINT_ALGORITHM_DEFAULT);
   chromaprint_start(chromaprint, kDecodeRate, kDecodeChannels);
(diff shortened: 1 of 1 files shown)


src/musicbrainz/chromaprinter.cpp
```diff
@@ -143,16 +143,24 @@ QString Chromaprinter::CreateFingerprint() {
   ChromaprintContext* chromaprint =
       chromaprint_new(CHROMAPRINT_ALGORITHM_DEFAULT);
   chromaprint_start(chromaprint, kDecodeRate, kDecodeChannels);
-  chromaprint_feed(chromaprint, reinterpret_cast<void*>(data.data()),
+  chromaprint_feed(chromaprint, reinterpret_cast<int16_t *>(data.data()),
```
hatstand: I imagine this fails on older chromaprint versions.

Chocobozzz: This compiles well on my GCC. There is an implicit conversion `int16_t* -> void*`.

See http://en.cppreference.com/w/cpp/language/implicit_conversion (**Pointer conversions**).

Issue #5558, opened 2016-12-17
[qt5] Podcast episodes forgotten


beojan:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [x] I checked the issue tracker for similar issues
- [x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [ ] ~~I tried the latest Clementine build from [here](https://builds.clementine-player.org/)~~ 
*I'm building from git*

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Arch Linux
- Clementine version: 1.3.1-318-g36d617a (qt5 branch)

### Expected behaviour / actual behaviour
When I update podcast episodes, the list of episodes disappears when Clementine is restarted.

### Steps to reproduce the problem (only for bugs)

1. Add a podcast
2. Right-click and update podcast 
 *There is now a list of episodes.*
4. Quit and restart Clementine 
*The list of episodes has been forgotten.*

Chocobozzz:
Thanks for the report :) It should be fixed.

PR #5559, opened 2016-12-18
Fix podcast episodes forgotten

Chocobozzz:
Fixes https://github.com/clementine-player/Clementine/issues/5558


Chocobozzz:
Oup's, the bug is on master too.

PR #5572, opened 2016-12-29
Fix compilation with Clang on linux

Fat-Zer:
### tl;dr version:
Fix compilation with clang on linux, which is broken for a while.

### Background
Building of Clementine on linux with clang is broken for a long time: see #4227.
First attempt to fix it (#4228) failed the chromaprint building which is currently removed from 3rdparty.
Last attempt (#5443) was rejected because of, as I believe, a bit obscure messages for patches and explanation.

#### vreen: remove -stdlib=libc++ compiler flag
The string commented out added to clang flag "-stdlib=libc++" causing clang try to link against [libcxx](http://libcxx.llvm.org/) which is (as I suppose) a desired for MacOS, but neither not available nor the preferred one on a linux desktop.
Removing it is safe because:
 - gcc and other compilers doesn't support this flag and thus it is not passed to them (cmake checks if the compiler supports it).
 - for MacOS the flag is set by the uplevel [CMakeLists.txt](https://github.com/clementine-player/Clementine/blob/master/CMakeLists.txt#L21)

#### Remove -U__STRICT_ANSI__ from CXXFLAGS
Frankly speaking I'm not sure why -U__STRICT_ANSI__  was added. It's an internal macro which should be triggered by switching the "-std*" for C++ and for non-gnu++ code it should be defined. Anyway it seems it isn't really needed anymore. Although, it would be nice to get some feedback if the patch doesn't break the compilation on supported systems.


Fat-Zer:
@hatstand, please at least run it through build bot at let me an opportunity to fix the patch if it break some build before closing this PR...
Sorry if I'm making noise, but I wouldn't insist on these changes if I haven't thought they are essential.

Fat-Zer:
Removed vreen patch due to it's removal.

PR #5581, opened 2017-01-08
fix Windows build libmygpo-qt5

pavelvat:
libmygpo-qt5 is static library.
Clementine/3rdparty/libmygpo-qt5/src/CMakeLists.txt:
```
add_library( ${MYGPO_QT_TARGET_NAME} STATIC ${LIBMYGPO_QT_SRC} ${LIBMYGPO_QT_MOC_SRC} )
```

log: https://gist.github.com/pavelvat/2072ae11fb8c94fba317c371eaeba871


3rdparty/libmygpo-qt5/src/mygpo_export.h
```diff
@@ -25,6 +25,8 @@
 
 #include <QtCore/qglobal.h>
 
+#define MYGPO_EXPORT
```
hatstand: This is pretty weird with the `#ifndef MYGPO_EXPORT` below

pavelvat: What you propose?

Something like this:
```
#ifndef MYGPO_BUILD_TYPE_STATIC
# if defined(MYGPO_MAKEDLL)
/* We are building this library */
#  define MYGPO_EXPORT Q_DECL_EXPORT
# else
/* We are using this library */
#  define MYGPO_EXPORT Q_DECL_IMPORT
# endif
#else
# define MYGPO_EXPORT
#endif
```
?

hatstand: LGTM

pavelvat:
Done.

PR #5588, opened 2017-01-08
Fix windows7thumbbar.cpp and osdpretty.cpp

pavelvat:
http://jpnurmi.kapsi.fi/blog/2012/11/13/qpixmaptowinhicon-in-qt5/

logs:
https://gist.github.com/pavelvat/54c99441a5971e7663034d185428e3ae 
https://gist.github.com/pavelvat/b76eff38b43863a18e6dd92cefa67e48

### DIFF
--- a/src/ui/windows7thumbbar.cpp
+++ b/src/ui/windows7thumbbar.cpp
@@ -53,10 +53,13 @@ void Windows7ThumbBar::SetActions(const QList<QAction*>& actions) {
 }
 
 #ifdef Q_OS_WIN32
+
+extern HICON qt_pixmapToWinHICON(const QPixmap &p);
+
 static void SetupButton(const QAction* action, THUMBBUTTON* button) {
   if (action) {
     button->hIcon =
-        action->icon().pixmap(Windows7ThumbBar::kIconSize).toWinHICON();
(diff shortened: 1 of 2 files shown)


src/ui/windows7thumbbar.cpp
```diff
@@ -53,10 +53,17 @@ void Windows7ThumbBar::SetActions(const QList<QAction*>& actions) {
 }
 
 #ifdef Q_OS_WIN32
+
+#if QT_VERSION >= 0x050000
+extern HICON qt_pixmapToWinHICON(const QPixmap &p);
+#else
+#define qt_pixmapToWinHICON(p) p.toWinHICON()
```
hatstand: I'd prefer a wrapper function that hides all this rather than using macros.

pavelvat: Currently qt5 branch have CMakeLists.txt files only for Qt5.
Has Clementine-team any plans to add support for Qt4 and Qt5 simultaneously to one git-branch?
If Qt4 support is not needed then maybe only this one line:
```
 +extern HICON qt_pixmapToWinHICON(const QPixmap &p); 
```

hatstand: No, I don't think we care about simultaneous support.

pavelvat:
Done.

Issue #5574, opened 2016-12-31
Spotify plugin search crashes Clementine


gbrauen:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [ x] I checked the issue tracker for similar issues
- [ x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [ x] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Ubuntu 16.04 GNOME
- Clementine version: clementine 1.3.1-228-gd9b3a93~xenial [origin: LP-PPA-me-davidsansome-clementine] - also with latest - same behaviour so reverted to released version

### Expected behaviour / actual behaviour

Clementine is my main music player because it nicely integrates spotify streaming and playback of local files. Until recently. Mid-December 2016 both of my installations of Clementine on Ubuntu 16.04 (different hardware architectures) stopped working after package upgrades - not sure which package has caused the problem.

Expected behaviour:
1. Select Internet tab in Clementine; Click on Spotify.
    * Expected result: my spotify playlists should be listed.
2. Type search token into spotify search box.
    * Expected result: matching search results should be listed.

Actual results:
For matching steps above:
1. Spotify playlists are no longer listed. No information from my spotify account is ever listed.
2. Whenever I complete a search token by typing a space, for example "join ", spotify crashes.

Crash report: `clementine crashed with SIGSEGV in QStandardItem::rowCount()`

I deleted the spotify plugin (all of ~/.config/Clementine/spotify*) and restarted.  Problem persists once plugin reinstalled. The spotify plugin connects and logs in to my spotify account without problem.

#### SegvAnalysis
> Segfault happened at: 0x7f261c3b7100 <_ZNK13QStandardItem8rowCountEv>:	mov    0x8(%rdi),%rax
> PC (0x7f261c3b7100) ok
> source "0x8(%rdi)" (0x00000008) not located in a known VMA region (needed readable region)!
> destination "%rax" ok

#### SegvReason
> Reading NULL VMA

#### Signal
> 11

#### Disassembly
> => 0x7f261c3b7100 <_ZNK13QStandardItem8rowCountEv>:	mov    0x8(%rdi),%rax
>    0x7f261c3b7104 <_ZNK13QStandardItem8rowCountEv+4>:	mov    0x28(%rax),%eax
[...]

PR #5593, opened 2017-01-12
Workaround to spotify loading playlist issue

Chocobozzz:
Some Spotify playlists never load (libspotify/spotify server bug I guess). This is a workaround in order that Clementine does not stuck on unloaded playlists.

Fixes https://github.com/clementine-player/Clementine/issues/5574

### DIFF
--- a/ext/clementine-spotifyblob/spotifyclient.cpp
+++ b/ext/clementine-spotifyblob/spotifyclient.cpp
@@ -433,8 +433,8 @@ void SpotifyClient::SendPlaylistList() {
                 << sp_playlist_name(playlist);
 
     if (!is_loaded) {
-      qLog(Info) << "Playlist is not loaded yet, waiting...";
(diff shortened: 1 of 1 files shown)


Fermman:
Same problem here.

rsec:
This commit solved the problem with spotify playlists! Thanks!

akademy:
This appears to be a problem with Playlists created specifically by Spotify (They say "By Spotify" in my list of playlists). I couldn't get the spotify service to work at all (it also crashed when I tired to search) until I deleted all the followed Spotify created playlists, including:

- Discover Weekly
- Throwback Thursday
- 88 Keys
- Easy 00s
- Music for Concentration
- (Others too)

But I do have one playlist "By Spotify UK" which worked fine.

I can also confirm that re-adding one of the "By Spotify" playlists that it once again causes a problem.

Issue #5457, opened 2016-07-31
Can't login to Digitally Imported (or Radiotunes) premium account


fkelbe:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [x] I checked the issue tracker for similar issues
- [x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [x] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)
    - clementine-1.3.2-0.rc1.fc23.x86_64.rpm  and ClementineSetup-1.3.2rc1.exe

### System information
- Operating System: Fedora 24 x86_64 and Windows 10 64-bit
- Clementine version: 1.3.1 and later

### Expected behaviour / actual behaviour
Expected:
Run Clementine and login to a Digitally Imported (or Radiotunes) premium account

Actual:
   The latest versions of Clementine will no longer login to Digitally Imported (DI) Premium accounts. Since Radiotunes and Digitally Imported (DI) are the same company and use the same configuration/login, this applies to both. 
   The error is a small pop-up that has a title of "Authentication Failure" and the text in the box says "Unknown error". The DI account has been verified to still be valid and works fine through web logins.

### Steps to reproduce the problem (only for bugs)
- Start Clementine
- Select "Internet"
- Right click on "Digitally Imported" or "Radiotunes" and select "Configure"
- Enter Premium account username and password
- Error pop-up then appears

major:
I also tried removing `~/.cache/Clementine` and that didn't fix it. For what it's worth, I can't listen to DI/RadioTunes in Clementine as a guest user, either.

major:
If I dump the network traffic from Clementine with `ngrep`, I see that DI's API is responding with success:

```
HTTP/1.1 200 OK.
Content-Type: application/json; charset=utf-8.
Vary: Accept-Encoding.
Status: 200 OK.
Cache-Control: max-age=0, private, must-revalidate.
Content-Language: en.
X-Request-Id: xxx.
ETag: W/"xxx".
X-Runtime: 0.111967.
Date: Tue, 02 Aug 2016 13:15:48 GMT.
X-Powered-By: Phusion Passenger Enterprise.
Server: nginx/1.8.1 + Phusion Passenger.
Content-Encoding: gzip.
X-Frame-Options: SAMEORIGIN.
X-Varnish: 620267412.
Age: 0.
Via: 1.1 varnish-v4.
[...]

PR #5603, opened 2017-01-18
Fix OAuthenticator reply parsing.

selumin:
Provide QByteArray instead of QIODevice to QJson parser.
Bug link: https://github.com/flavio/qjson/issues/52

### DIFF
--- a/src/internet/box/boxservice.cpp
+++ b/src/internet/box/boxservice.cpp
@@ -125,7 +125,7 @@ void BoxService::FetchUserInfoFinished(QNetworkReply* reply) {
   reply->deleteLater();
 
   QJson::Parser parser;
-  QVariantMap response = parser.parse(reply).toMap();
+  QVariantMap response = parser.parse(reply->readAll()).toMap();
 
   QString name = response["name"].toString();
   if (!name.isEmpty()) {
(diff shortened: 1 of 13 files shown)


selumin:
Probably will fix #5457

hatstand:
If this really fixes things, then I'd like to know why because we have this pattern everywhere.

selumin:
This pattern is correct, but there is a regression in libqjson, which was introduced between 0.8.1 and 0.9.0 releases (774c7785099be575d3d39fa698c6186919e4521a - break commit if someone is interested).

selumin:
I've updated all problem places in code. It resolves issues with loggin into Google Drive and Dropbox, for example.

hatstand:
We should rely on the upstream fix I think and perhaps check for QJson > 0.9.0 at compile time.

selumin:
@hatstand But clementine use shared version of qjson, so compile time check wouldn't solve the problem. 
Could we temporary add qjson to project's repo as 3rd party and build it as static library? 
When qjson version with fixed behavior will be released, we will drop dependency from repo.

hatstand:
I realise it's not perfect but it's best if we can avoid adding more things to 3rdparty. Perhaps we can add a wrapper function for now that we use everywhere.

The best thing to do would be to harass distros into updating their qjson for the bugfix.

abika:
This bug in QJson also breaks fetching tags from Musicbrainz.
How about using the [JSON support build-in in Qt](http://doc.qt.io/qt-5/json.html)? It came with 5.0 which is 4 years old now. We could remove QJson everywhere.

abika:
Nevermind, Clementine is still using Qt4.

Well, I can help if somebody decides how to solve this (globally).

Issue #5623, opened 2017-02-02
Album art (cover) manager uses artist tag instead of album artist tag


ShabazzKing:
Please make Clementine use "album artist" tag instead of "artist". If tracks have multiple artists in the "artist" tag (for example many rappers perform on particular song), then cover manager shows many duplicates of the same album's cover. If someone doesn't use "album artist" tag (because doesn't know about its existence), make cover manager use "artist" tag only when "album artist" tag is empty.

gsantner:
You can actually change whats visible by clicking on one of the columns. Just take `Artists` out and `Album Artists` in

ShabazzKing:
I'm not talking about the main window of the program, but about the cover manager.
As one can see in the screenshot attached, there are no columns in the cover manager window.
![screenshot_20170205_215526](https://cloud.githubusercontent.com/assets/25502034/22629830/18d27cc6-ebee-11e6-84ba-b91355fe1e91.png)

santigl:
It should now behave like that, @ShabazzKing. (Many thanks to @hatstand for the merges!)

PR #5647, opened 2017-03-03
Prioritize album artist in cover search and manager

santigl:
As mentioned in #5623, currently the cover manager and search only take into account an album `artist` field.

These changes make it so that the `effective_albumartist` field is used instead, which is defined as: `albumartist_.isEmpty() ? artist_ : albumartist_`.

Also, the cover manager now shows a list of artists composed by all names that appear in some album as either:
* a) `albumartist`, or
* b) `artist` when `albumartist` is empty.

### DIFF
--- a/src/library/librarybackend.cpp
+++ b/src/library/librarybackend.cpp
@@ -536,10 +536,12 @@ QStringList LibraryBackend::GetAllArtistsWithAlbums(const QueryOptions& opt) {
   query2.AddWhere("album", "", "!=");
   query2.AddWhere("albumartist", "", "=");
 
-  QMutexLocker l(db_->Mutex());
-  ExecQuery(&query);
-  ExecQuery(&query2);
-  l.unlock();
+  {
+    QMutexLocker l(db_->Mutex());
+    if (!ExecQuery(&query) || !ExecQuery(&query2)) {
+      return QStringList();
+    }
+  }
 
   QSet<QString> artists;
   while (query.Next()) {
@@ -870,12 +872,14 @@ LibraryBackend::AlbumList LibraryBackend::GetAlbums(const QString& artist,
     query.AddWhere("artist", artist);
   }
 
-  QMutexLocker l(db_->Mutex());
-  if (!ExecQuery(&query)) return ret;
-  l.unlock();
+  {
+    QMutexLocker l(db_->Mutex());
+    if (!ExecQuery(&query)) return ret;
+  }
 
   QString last_album;
-  QString last_artist, last_album_artist;
+  QString last_artist;
+  QString last_album_artist;
   while (query.Next()) {
     bool compilation = query.Value(3).toBool() | query.Value(4).toBool();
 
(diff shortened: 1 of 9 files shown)


santigl:
The last two commits introduce the `effective_album` field to the `Song` class. It is defined as: `album_.isEmpty() ? title_ : album_`. 

This is now used by the `AlbumCoverChoiceController` to attempt to find covers for tracks without an album name set.

src/library/librarybackend.cpp
```diff
@@ -523,29 +523,48 @@ QStringList LibraryBackend::GetAllArtists(const QueryOptions& opt) {
...
+  query2.SetColumnSpec("DISTINCT artist");
+  query2.AddCompilationRequirement(false);
+  query2.AddWhere("album", "", "!=");
+  query2.AddWhere("albumartist", "", "=");
+
   QMutexLocker l(db_->Mutex());
-  if (!ExecQuery(&query)) return QStringList();
+  ExecQuery(&query);
+  ExecQuery(&query2);
+  l.unlock();
```
hatstand: Never explicitly call `unlock()`. Add a scope if you need it instead.

src/library/librarybackend.cpp
```diff
@@ -830,58 +849,74 @@ void LibraryBackend::UpdateCompilations(QSqlQuery& find_songs,
...
     query.AddWhere("artist", artist);
   }
 
   QMutexLocker l(db_->Mutex());
   if (!ExecQuery(&query)) return ret;
+  l.unlock();
 
   QString last_album;
-  QString last_artist;
+  QString last_artist, last_album_artist;
```
hatstand: One per line

src/ui/albumcovermanager.cpp
```diff
@@ -815,3 +825,34 @@ void AlbumCoverManager::UpdateExportStatus(int exported, int skipped, int max) {
...
+  QPainter p(&square_nocover);
+  p.setOpacity(0.4);
+  p.drawImage((120 - nocover.width()) / 2, (120 - nocover.height()) / 2,
+              nocover);
+  p.end();
+
+  return square_nocover;
+}
+
+bool AlbumCoverManager::ItemHasCover(const QListWidgetItem* item) const {
```
hatstand: This can be a reference instead of a pointer.

src/ui/albumcovermanager.cpp
```diff
@@ -815,3 +825,34 @@ void AlbumCoverManager::UpdateExportStatus(int exported, int skipped, int max) {
     msg.exec();
   }
 }
+
+QString AlbumCoverManager::EffectiveAlbumArtistName(
+    const QListWidgetItem* item) const {
```
hatstand: Reference instead of pointer.

santigl:
Thanks a lot, @hatstand!

I made the last commit (refactoring `EffectiveAlbumArtistName()`) a couple of seconds later than the merge, and I'm not seeing it appear here. Is it showing for you? Should I include it in another PR?

hatstand:
Go for it

PR #5649, opened 2017-03-05
Add contextual searches to playlists, library and global search.

### LINKED ISSUE #5567: Enable search-by-artist and search-by-album when right-clicking on a song in playlist or search.
### Before posting

- [X] I checked the issue tracker for similar issues
- [X] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [X] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

- Operating System: Arch
- Clementine version: 1.3.1

Clementine should allow the user to search by artist or search by album by right-clicking on a search result, or a song in a playlist, and selecting the corresponding option. 

Example: User searches for song X, knowing it's on album Y. Instead of having to then search for album Y manually, the user right clicks on the result for song X, and selects "search by/for album", which returns the entire album.

JacobHenner:
Closes #5567

### DIFF
--- a/src/globalsearch/globalsearchview.cpp
+++ b/src/globalsearch/globalsearchview.cpp
@@ -465,15 +465,14 @@ bool GlobalSearchView::ResultsContextMenuEvent(QContextMenuEvent* event) {
       SLOT(AddSelectedToPlaylistEnqueue()));
 
   context_menu_->addSeparator();
+
   if (ui_->results->selectionModel() &&
       ui_->results->selectionModel()->selectedRows().length() == 1) {
     context_actions_ << context_menu_->addAction(
         IconLoader::Load("system-search", IconLoader::Base),
-        tr("Search for artist"), this, SLOT(SearchForArtist()));
-    context_actions_ << context_menu_->addAction(
-        IconLoader::Load("system-search", IconLoader::Base),
-        tr("Search for album"), this, SLOT(SearchForAlbum()));
+        tr("Search for this"), this, SLOT(SearchForThis()));
   }
+
   context_menu_->addSeparator();
   context_menu_->addMenu(tr("Group by"))
       ->addActions(group_by_actions_->actions());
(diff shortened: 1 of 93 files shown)


src/globalsearch/globalsearchview.cpp
```diff
@@ -515,6 +525,32 @@ void GlobalSearchView::OpenSelectedInNewPlaylist() {
   emit AddToPlaylist(data);
 }
 
+void GlobalSearchView::SearchForArtist() {
+  if (const SongMimeData* data = qobject_cast<const SongMimeData*>(SelectedMimeData())) {
+    if (data && !data->songs.isEmpty() && !data->songs.first().artist().isEmpty()) {
+      StartSearch(data->songs.first().artist().simplified());
+    }
+  }
```
hatstand: Should be the same line as the `else if`. Run `make format` to fix.

JacobHenner:
I've noticed that this does not work as expected if you select an album rather than a song under the library view. Will investigate sometime this week. This is not an issue with global search.

src/globalsearch/globalsearchview.cpp
JacobHenner: `make format` did not fix this for me. I'll do it manually.

JacobHenner:
@hatstand Fixed the issues I was encountering, and the formatting. Should be good to go.

JacobHenner:
Found a minor bug where selecting "search for this" on an album in Library view results in a prefixed entry in the search box. Will investigate.

src/library/libraryview.cpp
```diff
@@ -709,6 +737,29 @@ void LibraryView::FilterReturnPressed() {
...
+  SaveFocus();
+  if (!last_selected_song_.album().isEmpty()) {
+    filter_->ShowInLibrary(last_selected_song_.album().simplified());
+  }
+}
+
+void LibraryView::SearchForThis() {
+  SaveFocus();
+  if (!last_selected_container_.isEmpty()) {
+    filter_->ShowInLibrary(last_selected_container_.simplified());
```
JacobHenner: Is there any way to get a more predictable result for last_selected_container_? Right now, it varies based upon the sorting/grouping method used. For example: 

2chainz

selecting "search for this" results in a search for `2013B.O.A.T.S. II #MeTime`. 

Searching for 
screenshot from 2017-03-14 23-00-40

results in a search for `10302009 i  festival 8 ca`

JacobHenner: Figured it out. Will tackle tomorrow.

JacobHenner: So I see two options here: 
1. Search for exactly what's being displayed when a user clicks "search for this" on a divider, which might not return results if the album year is included as per the sorting preferences.
2. Modify librarymodel.cpp to record the album, album artist, or artist for each divider (depending on type of divider), and search for that. 

Which is more suitable?

Fat-Zer:
If in library view settings "Show only duplicates/untaged" is set thus the Filter field is disabled (why/should it?), but it can still be populated by "Search for *" actions.

JacobHenner:
@Fat-Zer I will take a look at this.

Also, just noticed a bug/inconsistency in global search when right-clicking on a container/divider. Going to bring it in-line with the Library search behavior of "search for this" if possible.

src/library/libraryview.cpp
hatstand: It would be pretty weird if a user could initiate a search from there that resulted in zero results.

JacobHenner:
@hatstand Would a schema update be acceptable in order to complete this feature?

hatstand:
If you need it, sure.

JacobHenner:
Updated to show "search for this" instead of album/artist search in both global and library views for a more consistent experience. User can click on headers in tree to search for artist or album, or whatever other attributes user decides to sort tree by.

JacobHenner:
@hatstand Any thoughts?

JacobHenner:
Anyone available to review this?

JacobHenner:
Rebased

hatstand:
I think you forgot to include the new schema

JacobHenner:
@hatstand Indeed. Fixing now, standby.

JacobHenner:
@hatstand Fixed.

JacobHenner:
@hatstand  This should be ready

hatstand:
I still don't see a schema

JacobHenner:
@hatstand Added. Must have gotten lost in a bad rebase, sorry.

JacobHenner:
This should be good to go.

hatstand:
Travis says this doesn't build, please fix.

JacobHenner:
That was embarrassing. Looks like some extraneous text made its way into that commit. It should build now...

JacobHenner:
Yep, it's good.

JacobHenner:
Thank you, sir!

PR #5658, opened 2017-03-11
Refactor FHT class (fixes buffer overrun)

santigl:
These changes make the oldish `FHT` class code to use newer, safer structures. 

This avoids a potential buffer overrun [here](https://github.com/clementine-player/Clementine/compare/master...santigl:fht-refactor?expand=1#diff-1b0025c3a227af0f536329d279c4a2eeL92), where the `scope` size can be smaller than `fht_->size()`.

Because the code is too "convoluted" I didn't modify the algorithms, so pointers are still being passed as arguments.

Along modifications to the different analyzers to adopt those changes I made a few refactorings in their code as well, improving readability and making vector members use the `QVector` type everywhere it was possible.


hatstand:
Maybe we should just use something from fftw now instead.

Issue #5366, opened 2016-04-30
Hide the left part of the toolbar?


Movunh:
Hi,

I'm on Windows 7. I'm new on Clementine and I think it's close to be the best audio player ever for my personal use. However, a little problem bothers me, the left toolbar which pop up every time. I have a huge library and every time I see the left toolbar it reminds me that I will never get the courage to fix covers for 6to of folders (yeah, I know Album Art Downloader but I have some concerns about the fails). 
It looks like that for almost every single album: http://www.imagebam.com/image/97547f480776682
I don't need track informations too, I have them in the playlist. 
I would feel very grateful if you could find a way to hide it.

Thank you in advance.

Movunh:
I just modified the presentation. I apologize for my broken english, my first language is french.

Fat-Zer:
Are you talking about hiding album image on the nowplaying widget, or the whole nowplaying widget, or the whole sidebar with library browser/artist&track info/etc?

Movunh:
Thanks for the response. I want to keep the library, I'm talking about hiding the whole nowplaying widget (album image+_track title_+artist+album), this is what I really want but hiding album image would already be a significant improvement for me.

Movunh:
Is there any chance that you could at least add an option to hide album image and realign properties in the next releases?

PR #5667, opened 2017-03-15
NowPlayingWidget: add option to disable cover display

santigl:
(Feature request #5366)


Fat-Zer:
I was experimenting with it a while ago... don't you think it looks just a bit crappy?

santigl:
It maybe looks kind of weird, after being used to seeing some image there all the time.

emre-aydin:
I also need this functionality. Since it's optional, I don't think it would matter if it looks good or not.

Issue #5659, opened 2017-03-13
Unable to add stream url


andrew-filippov:
Hi,
I have a troubles with adding new streams with urls as follow:
`http://8.38.78.173:8240/listen.pls?sid=1`

(this url is was picked from http://www.internet-radio.com).
Url without any parameters work fine.

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: XUbuntu 16.04
- Clementine version: v1.2.3

santigl:
This seems to be a problem with GStreamer. It has been reported in the past (#5399 for example).

I think a possible solution would be to try taking a look at the URL's MIME type. The current code seems to assume that GStreamer would do that, but for some reason it is not. 

(Perhaps that functionality changed in 1.0? Maybe @hatstand, @davidsansome or @amuttsch know.)

Meanwhile, @andrew-filippov, you can add the direct stream URL, [http://8.38.78.173:8240/stream](http://8.38.78.173:8240/stream), or download the .pls file and open it locally. Both alternatives seem to be working.

Issue #5399, opened 2016-06-05
.PLS files in Ubnutu 16.04


fwd079:
Good morning,

### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [x] I checked the issue tracker for similar issues
- [x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [x] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System:  **Ubuntu 16.04 64bit**
- Clementine version: 1.3.1

### Expected behaviour / actual behaviour
**Expected:** Paste URL in search bar / Add via Add Stream:
http://stereoscenic.com/pls/pill-hi-aac.pls
Listen to music.

**Actual:** Paste URL in search bar / Add via Add Stream:
http://stereoscenic.com/pls/pill-hi-aac.pls
Nothing happens. (also tried http://stereoscenic.com/pls/pill-hi-mp3.pls)

### Steps to reproduce the problem (only for bugs)
Since I know this is a bug, follow the steps please:
Paste URL in search bar / Add via Add Stream:
http://stereoscenic.com/pls/pill-hi-aac.pls
Result: Nothing happens. (also try http://stereoscenic.com/pls/pill-hi-mp3.pls)

I have installed Audacious and above step works fine, so I know for a fact this is **not a Ubuntu 16.04 issue**. Please I do not want to back-port/install old packages etc. 

May I please request to ping here when the bug is fixed in Clementine?

Thanks.

santigl:
Thanks for reporting the bug. I could recreate in Debian 64 bits. 

Apparently, in function [`SongLoader::LoadRemote()`](https://github.com/clementine-player/Clementine/blob/fde64e31b8b7016050512c46c551b5098821fb27/src/core/songloader.cpp#L369) GStreamer's`typefind` is not detecting the format of the playlist. 

In addition, callback function [`SongLoader::DataReady()`](https://github.com/clementine-player/Clementine/blob/fde64e31b8b7016050512c46c551b5098821fb27/src/core/songloader.cpp#L450) is not being triggered. That function would normally detect the file as a PLS playlist (I could verify that, by forcing it to execute, the playlist format is correctly identified through the *magic*).

santigl:
Sorry, @fwd079. I should have added that meanwhile, if you download the .pls and add it through the *Add file* button, the stream plays.

[...]

Issue #4977, opened 2015-08-04
Can't play Shoutcast Stream


TyrasNistru:
Version: 1.2.3-1241-gfbe310b
OS: Windows 10

I'm trying to open the stream http://radioarquivodorock.com.br:9998/listen.pls on Clemetine, But it got stuck in "Loading Tracks: 0%"

Trying to reopen the stream without closing and reopening Clementine result in the status bar getting stacked entries ("Loading Tracks: 0%, Loading Tracks: 0%, Loading Tracks: 0%...")

Nelsyv:
This is happening to me as well, whenever I try to add the following streams:

http://www.bigblueswing.com:8002/listen.pls
http://50.7.173.162:2199/tunein/jazz.pls
http://50.7.173.162:2199/tunein/enieopyy.pls

OS: Windows 7 SP1
Version: 1.3.1 (the main stable release from the website, I downloaded it today)

seanbruno:
Choosing this issue as there appear to be two in the tracker for ShoutCast streams.

Command line attempt to access a known working shoutcast stream (with debug enabled) looks like this in the failure case:

10:02:19.966 DEBUG MainWindow:301                   Creating UI 
10:02:20.188 DEBUG MainWindow:846                   Creating equalizer 
10:02:20.190 DEBUG MainWindow:875                   Creating now playing widget 
10:02:20.323 DEBUG MainWindow:937                   Loading settings 
10:02:20.343 INFO  NetworkRemote:80                 Network Remote deactivated 
10:02:20.344 DEBUG MainWindow:1016                  Started 
10:02:20.344 DEBUG SongLoader:370                   Loading remote file  QUrl( "http://www.radio-kol.net/64/listen.pls" )  
10:02:20.377 INFO  DeviceManager:405                Device added: "/dev/cd0" 
10:02:21.007 ERROR SongLoader:517                   Could not determine type of stream. 
10:02:21.007 ERROR SongLoader:518                   gsttypefindelement.c(992): GstFlowReturn gst_type_find_element_chain_do_typefinding(GstTypeFindElement *, gboolean, gboolean) (): /GstPipeline:pipeline0/GstTypeFindElement:typefindelement0 
10:02:21.007 ERROR SongLoader:517                   Could not determine type of stream. 
[...]

PR #5671, opened 2017-03-17
Fix remote-playlist load

santigl:
As reported in #4977, #5659, #5399 and many others, the loading of remote playlists was broken.

These changes seek to fix that.

### DIFF
--- a/src/core/songloader.cpp
+++ b/src/core/songloader.cpp
@@ -34,6 +34,7 @@
 #include "config.h"
 #include "core/logging.h"
 #include "core/player.h"
+#include "core/utilities.h"
 #include "core/signalchecker.h"
 #include "core/song.h"
 #include "core/tagreaderclient.h"
@@ -115,6 +116,11 @@ SongLoader::Result SongLoader::Load(const QUrl& url) {
     return Success;
   }
 
+  // It could be a playlist, we give it a shot.
+  if (LoadRemotePlaylist(url_)) {
+    return Success;
+  }
+
   url_ = PodcastUrlLoader::FixPodcastUrl(url_);
 
   preload_func_ = std::bind(&SongLoader::LoadRemote, this);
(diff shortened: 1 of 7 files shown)


src/core/songloader.cpp
```diff
@@ -617,3 +620,64 @@ void SongLoader::StopTypefindAsync(bool success) {
...
+  // If the URL does not point to a playlist file we could handle,
+  // it returns false.
+
+  NetworkAccessManager manager;
+  QNetworkRequest req = QNetworkRequest(url);
+
+  // Getting headers:
+  QNetworkReply* reply = manager.head(req);
+  {
+    QEventLoop loop;
```
hatstand: `WaitForSignal(reply, SIGNAL(finished()));`

src/core/songloader.cpp
```diff
@@ -617,3 +620,64 @@ void SongLoader::StopTypefindAsync(bool success) {
...
+    QObject::connect(reply, SIGNAL(finished()), &loop, SLOT(quit()));
+    loop.exec();
+  }
+  if (reply->error() != QNetworkReply::NoError) {
+    qLog(Debug) << url.toString() << reply->errorString();
```
hatstand: `qLog(Error)`

src/core/songloader.cpp
```diff
@@ -617,3 +620,64 @@ void SongLoader::StopTypefindAsync(bool success) {
...
+  if (parser == nullptr) {
+    qLog(Debug) << url.toString() << "seems to not be a playlist";
+    return false;
+  }
+
+  // We know it is a playlist!
+  // Getting its contents:
+  reply = manager.get(req);
+  {
+    QEventLoop loop;
```
hatstand: `WaitForSignal(reply, SIGNAL(finished()))`

src/core/songloader.cpp
```diff
@@ -617,3 +620,64 @@ void SongLoader::StopTypefindAsync(bool success) {
...
+
+  ParserBase* parser = playlist_parser_->ParserForMimeType(mime_type);
+  if (parser == nullptr) {
+    qLog(Debug) << url.toString() << "seems to not be a playlist";
+    return false;
+  }
+
+  // We know it is a playlist!
+  // Getting its contents:
+  reply = manager.get(req);
```
hatstand: Use a different variable name to avoid confusion

src/core/utilities.cpp
```diff
@@ -231,6 +236,23 @@ QString GetTemporaryFileName() {
   return file;
 }
 
+QString SaveToTemporaryFile(const QByteArray& data) {
+  QString filename = GetTemporaryFileName();
```
hatstand: You should just directly use `QTemporaryFile` here to avoid any race conditions.

src/core/songloader.cpp
```diff
@@ -632,52 +634,49 @@ bool SongLoader::LoadRemotePlaylist(const QUrl& url) {
...
+  }
+
+  qLog(Debug) << url.toString() << "with MIME" << mime_type << "loading from"
               << playlist_filename;
 
   // ...and load it.
   LoadPlaylist(parser, playlist_filename);
 
-  file.remove();
+  QFile(playlist_filename).remove();
```
hatstand: It would probably be better if `Utilities::SaveToTemporaryFile()` returned a `QFile` so you don't need to do this.

src/core/utilities.cpp
santigl: Ok. I was trying to avoid having duplicate code.

src/core/songloader.cpp
santigl: You're right. But then the function in SongLoader would have to check the actual file to see that it was written, wouldn't it?

santigl: Should I change it to `QFile SaveToTemporaryFile(const QByteArray& data, bool* success)`?

PR #5677, opened 2017-03-24
add file .gitattributes and remove -j8 from make-parameters

mbrehmer:
The commits of this pull-request do two things:

- Adding a .gitattributes-file, to ensure correct tracking and diffing for different filetypes. (see commit 019dddc for details)
- Removing of the parameter -j8 from make-command in README.md. (see commit 1b729ad for reasons)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,19 @@
+* text=auto
+
+*.cpp text diff=cpp
+*.h text diff=cpp
+*.in text
+*.txt text
+*.pem text
+*.css text diff=css
+*.md text
+*.sh text
+*.py text
+*.conf text
+*.README text
+*.js text
+
+*.png binary
+*.ttf binary
+*.icns binary
+*.ico binary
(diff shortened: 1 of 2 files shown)


mbrehmer:
@hatstand Have you read the whole commit descriptions of  019dddc and 1b729ad?
Both commits contain descriptions of what they do and why.

Anyway, let me explain it a bit more.

- Usually the files .gitignore and .gitattributes are the very beginning of a good Git repository.
Without a proper .gitattributes file, Git tracks binary files like simple (sourcecode-)textfiles. That causes inconsistent commits, if you want to track any binary files with Git. So you have to tell git which files are textfiles and which have to be tracked binary. By default Git tracks every file as a textfile.

- Running make with eight parallel processes (parameter -j 8) is a bad idea, if your machine hasn't at least eight **physical** cores. Hyper-Threating doesn't count here.
Make has the power to drive your system to its knees, if you run it with more threads than your number of physical cores. I had to learn that the hard way.

PR #5679, opened 2017-03-25
Fix some warnings shown on start-up

santigl:
> QColor::fromRgb: RGB parameters out of range 

> Couldn't load icon "find"

> QPainter::begin: Paint device returned engine == 0, type: 2

### DIFF
--- a/src/analyzers/analyzerbase.cpp
+++ b/src/analyzers/analyzerbase.cpp
@@ -241,9 +241,9 @@ QColor Analyzer::Base::getPsychedelicColor(const Scope& scope,
   rgb[1] = sqrt(rgb[1]) * ampFactor + bias;
   rgb[2] = sqrt(rgb[2]) * ampFactor + bias;
 
-  const int r = qMax(0, qMin(255, static_cast<int>(rgb[0])));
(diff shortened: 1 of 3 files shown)


src/analyzers/analyzerbase.cpp
```diff
@@ -236,12 +236,16 @@ QColor Analyzer::Base::getPsychedelicColor(const Scope& scope,
...
-  for (int i = 0; i < 3; ++i) {
-    // bias colours for a threshold around normally amplified audio
-    rgb[i] = qMin(255, (int)((sqrt(rgb[i]) * ampFactor) + bias));
-  }
+  // bias colours for a threshold around normally amplified audio
+  rgb[0] = sqrt(rgb[0]) * ampFactor + bias;
+  rgb[1] = sqrt(rgb[1]) * ampFactor + bias;
+  rgb[2] = sqrt(rgb[2]) * ampFactor + bias;
+
+  const int r = qMax(0, qMin(255, static_cast<int>(rgb[0])));
```
Fat-Zer: Note that there is a qBound() function for such stuff...

santigl: Thanks for pointing that out, @Fat-Zer!

Issue #5701, opened 2017-04-09
Using Clementine to delete a song from disk should not add song to the undo stack.


gdolphy2015:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [x] I checked the issue tracker for similar issues
- [x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [x] I tried the latest Clementine build from [code](https://github.com/clementine-player/Clementine)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Ubuntu 16.04.2
- Clementine version:  9f00e02

### Expected behaviour / actual behavior
Actual behavior: Undo option says you can undo a removed song, but does not actually restore the song.
Expected Behavior: Clementine should not add the deleted song from disk into the undo stack, and it should not be an option to undo a deleted song from disk anyway, because user is warned that it will be permanently deleted.

### Steps to reproduce the problem (only for bugs)
1. Open up Clementine with a playlist open.
2. Right Click on a song
3. Select "Delete from disk..."
4. Select "yes"
5. Right click anywhere in the playlist area again.
6. "Undo 1 remove song" will appear in the menu.

Issue #5576, opened 2017-01-03
Removing duplicates from playlist doesn't remove only duplicates


crissimaster:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [x] I checked the issue tracker for similar issues
- [x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [x] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Kubuntu 16.04
- Clementine version: 1.3.1-268-g789c492

###
Expected behaviour:
Only duplicate entries will be removed from playlist to get single entries

Actual behaviour:
Duplicates gets removed and some other entries too

### Steps to reproduce the problem (only for bugs)
Add some files from file browser (see bug 1785) which adds all file twice. Then select on the context menu of playlist "Remove duplicates from playlist"

fralix:
For me too.

A few months ago, my music library had missing title information like there were missing information about the artist and the song title, often there were no tags:

Today I added most of them and now it removes less files wrongly.

I guess that clementine checks the file information and if they are missing it interpretes them as similar?

billcarlson:
An addition problem, removing is based on something other than filename. This doesn't make sense, often released songs versus live performance will have the same name, artist, etc.

Confirmed this via checking one specific song that I knew had two versions and testing.

soundsbeard:
same thing
os: manjaro kde
take any amount of tracks, delete all info about them, and clementine will consider them similar, even if they have different length, name, bitrate etc.

rmpbklyn:
> same thing
> os: manjaro kde
> take any amount of tracks, delete all info about them, and clementine will consider them similar, even if they have different length, name, bitrate etc.

same issue, the songs mostly different  filee sizes and  file names, lost 2days works to make the playlist

Svidro:
Still an issue in March, 2020, version 1.3.1

Remove duplicates removed a dozen or so songs that were all missing Artist information.

KaterFindus:
Still unresolved, as of July 2020...
v. 1.3.1 in Xubuntu 18.04.4 LTX x64.

PR #5702, opened 2017-04-09
Delete Undo Fix Attempt - Should Solve Issue #5701 Using Clementine to delete a song from disk should not be added to the undo stack

gdolphy2015:
Hello again, our Software team has come up with another solution, this time to #5701. When calling Playlist::removeRows, the third parameter (which was not used before in the method) is now used to differentiate between a song deleted from the disk and a song deleted from the playlist by the new added boolean. To check between the two, range.topLeft should be a valid index (which we can say is a delete from playlist), and by default QModelIndex() is invalid (which we can say is the delete from disk). 

This allows a deletion from disk to not be pushed onto the undo stack. And this allows a deletion just from the playlist to be pushed onto the undo stack. I hope you will consider this fix into Clementine, thank you for your time.

### DIFF
--- a/src/playlist/playlist.cpp
+++ b/src/playlist/playlist.cpp
@@ -1593,6 +1593,8 @@ bool Playlist::removeRows(int row, int count, const QModelIndex& parent) {
     // might have been invalidated.
     RemoveItemsWithoutUndo(row, count);
     undo_stack_->clear();
+  } else if (parent.isValid() == false && parent.operator ==(QModelIndex())){
+      RemoveItemsWithoutUndo(row, count);
   } else {
     undo_stack_->push(new PlaylistUndoCommands::RemoveItems(this, row, count));
   }
--- a/src/playlist/playlist.cpp
+++ b/src/playlist/playlist.cpp
@@ -1593,8 +1593,8 @@ bool Playlist::removeRows(int row, int count, const QModelIndex& parent) {
     // might have been invalidated.
     RemoveItemsWithoutUndo(row, count);
     undo_stack_->clear();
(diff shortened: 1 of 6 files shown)


src/playlist/playlist.cpp
```diff
@@ -1593,6 +1593,8 @@ bool Playlist::removeRows(int row, int count, const QModelIndex& parent) {
     // might have been invalidated.
     RemoveItemsWithoutUndo(row, count);
     undo_stack_->clear();
+  } else if (parent.isValid() == false && parent.operator ==(QModelIndex())){
```
hatstand: This condition is pretty weird

HighMans:
Formatting should be fixed.

src/playlist/playlist.cpp
```diff
@@ -1593,6 +1593,8 @@ bool Playlist::removeRows(int row, int count, const QModelIndex& parent) {
     // might have been invalidated.
     RemoveItemsWithoutUndo(row, count);
     undo_stack_->clear();
+  } else if (parent.isValid() == false && parent.operator==(QModelIndex())) {
```
hatstand: Surely this can just be `} else if (parent == QModelIndex()) {`

gdolphy2015:
Just updated it, definitely is a better way to simplify the condition. Thank you.

jonaski:
@hatstand  @gdolphy2015 
This PR should be reverted, it creates a bigger issue than it solves, it completely breaks the undo stack because it calls RemoveItemsWithoutUndo directly making it impossible to undo removing duplicate songs which is frequently removing to many songs for users, see issue #5917 and #5576

PR #5706, opened 2017-04-13
Simpler iconloader::Load() logic

latach:
This fixes a bug when clementine uses an icon from the system, even though it exists in it the *data* folder.

For example: the icons at the top of *playlistlistview.ui*, and at the top of *fileview.ui*


latach:
O thank you!

PR #5723, opened 2017-05-03
New shortcut: remove current playing song from playlist

abika:
The title says it. It's a new shortcut (no default set) for removing the current playing song from the playlist.

Usage:
Clementine runs in the background, a horrible song starts playing - oh no! never want to hear it again -, you press *your favourite shortcut* to remove it from the playlist and *your second favourite shortcut* to start the next() song.

### DIFF
--- a/src/playlist/playlistmanager.cpp
+++ b/src/playlist/playlistmanager.cpp
@@ -494,6 +494,10 @@ void PlaylistManager::RemoveItemsWithoutUndo(int id,
   playlists_[id].p->RemoveItemsWithoutUndo(indices);
 }
 
+void PlaylistManager::RemoveCurrentSong() {
+  active()->removeRows(active()->current_index().row(), 1);
+}
+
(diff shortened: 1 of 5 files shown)


src/playlist/playlistmanager.cpp
```diff
@@ -494,6 +494,11 @@ void PlaylistManager::RemoveItemsWithoutUndo(int id,
   playlists_[id].p->RemoveItemsWithoutUndo(indices);
 }
 
+void PlaylistManager::RemoveCurrentSong()
+{
```
hatstand: `{` on the same line as the function like everywhere else or just run `make format`.

abika: Sorry, I'm used to another format. And `clang-format` on the whole file does also change other code because not everything is formatted.
How should `make format` supposed to work?
>make: *** No rule to make target 'format'.  Stop.

PR #5747, opened 2017-05-29
Passes network proxy information through to gstreamer

nicklan:
This uses the already defined network proxy settings, which before
did not apply to the actual streams.  This is useful for something
like a caching proxy that prevents re-downloading of network
resources (like tracks from a subsonic server for instance).

A few notes:
- gstreamer has a bug that prevents the use of socks proxies at the
  moment.  I've added a reference to the bug report in a comment and
  when that bug is resolved this code can be modified to also use
  socks proxies.

- I didn't make any changes to the UI.  In theory there could be
  another checkbox on the setting pane with something like "Use proxy
  for media streams" to control the behavior.  My feeling was that it
  would be odd to want to proxy things like last.fm API requests but
  not network streams, so I didn't add that.

### DIFF
--- a/src/engines/gstengine.cpp
+++ b/src/engines/gstengine.cpp
@@ -80,6 +80,7 @@ using std::vector;
 
 const char* GstEngine::kSettingsGroup = "GstEngine";
 const char* GstEngine::kAutoSink = "autoaudiosink";
+const char* GstEngine::kProxySettingsGroup = "Proxy";
 const char* GstEngine::kHypnotoadPipeline =
     "audiotestsrc wave=6 ! "
     "audioecho intensity=1 delay=50000000 ! "
@@ -106,6 +107,9 @@ GstEngine::GstEngine(TaskManager* task_manager)
       buffer_min_fill_(33),
       mono_playback_(false),
       sample_rate_(kAutoSampleRate),
+      proxy_url_(QString::null),
+      proxy_user_(QString::null),
+      proxy_passwd_(QString::null),
       seek_timer_(new QTimer(this)),
       timer_id_(-1),
       next_element_id_(0),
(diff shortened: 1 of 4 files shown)


src/engines/gstengine.cpp
```diff
@@ -106,6 +107,9 @@ GstEngine::GstEngine(TaskManager* task_manager)
       buffer_min_fill_(33),
       mono_playback_(false),
       sample_rate_(kAutoSampleRate),
+      proxy_url_(QString::null),
```
hatstand: You can skip these initialisers

src/engines/gstengine.cpp
```diff
@@ -210,6 +214,44 @@ void GstEngine::ReloadSettings() {
...
   mono_playback_ = s.value("monoplayback", false).toBool();
   sample_rate_ = s.value("samplerate", kAutoSampleRate).toInt();
+
+  s.endGroup();
+
+  s.beginGroup(kProxySettingsGroup);
+  const int mode = s.value("mode", 0).toInt();
+  proxy_url_.clear(); // make it null, assume no proxy
+
+  if (mode == 2) {
```
hatstand: This needs to be an enum so that it's obvious what the numbers actually mean.

src/engines/gstengine.cpp
```diff
@@ -210,6 +214,44 @@ void GstEngine::ReloadSettings() {
...
+    // 2 = manual config, we need to do something
+    const int type = s.value("type", 0).toInt(); // 1 = SOCKS, 3 = HTTP
+    if (type == 1) {
+      // See https://bugzilla.gnome.org/show_bug.cgi?id=783012
+      // When that bug is resoved we can append socks:// here
+      qLog(Warning) << "GStreamer doesn't currently support socks proxies." <<
+        " Will not use proxy for gstreamer streams";
+    } else if (type == 3) {
+      proxy_url_.append("http://");
```
hatstand: Use a `QUrl` here instead so you can use `setScheme()` etc.

src/engines/gstenginepipeline.cpp
```diff
@@ -75,6 +75,9 @@ GstEnginePipeline::GstEnginePipeline(GstEngine* engine)
       next_end_offset_nanosec_(-1),
       ignore_next_seek_(false),
       ignore_tags_(false),
+      proxy_url_(QString::null),
```
hatstand: Skip these too

src/engines/gstenginepipeline.cpp
```diff
@@ -986,6 +998,24 @@ void GstEnginePipeline::SourceSetupCallback(GstURIDecodeBin* bin,
     g_object_set(element, "ssl-strict", TRUE, nullptr);
 #endif
   }
+
+  const QString proxy_url = instance->proxy_url();
+  if (!proxy_url.isNull() && g_object_class_find_property(G_OBJECT_GET_CLASS(element), "proxy")) {
```
hatstand: `isEmpty()` is preferred

src/engines/gstengine.h
```diff
@@ -213,6 +214,11 @@ class GstEngine : public Engine::Base, public BufferConsumer {
   bool mono_playback_;
   int sample_rate_;
 
+  // proxy_url_.isNull => no proxy
+  QString proxy_url_;
+  QString proxy_user_;
+  QString proxy_passwd_;
```
hatstand: `proxy_password_` here and everywhere

src/engines/gstenginepipeline.cpp
```diff
@@ -986,6 +998,24 @@ void GstEnginePipeline::SourceSetupCallback(GstURIDecodeBin* bin,
...
+    g_object_set(element, "proxy", proxy_url.toLocal8Bit().constData(), nullptr);
+
+    const QString user = instance->proxy_user();
+    const QString passwd = instance->proxy_passwd();
+    if (!user.isEmpty() && // we set them together so only check one
+        g_object_class_find_property(G_OBJECT_GET_CLASS(element), "proxy-id") &&
+        g_object_class_find_property(G_OBJECT_GET_CLASS(element), "proxy-pw")) {
+      qLog(Debug) << "Setting proxy login to: " << user << " (password redacted)";
+      g_object_set (element,
+                    "proxy-id", user.toLocal8Bit().constData(),
```
hatstand: `toUtf8()` here and above

nicklan:
thanks for the comments!  i'm a little swamped right now, but will address these asap

PR #5758, opened 2017-06-10
Fix broken artist info tab

Atrament666:
Hi guys,
this fixes the broken fetching of artist info which hasn't been working for some time. 

Tested on up-to-date Manjaro Linux.

### DIFF
--- a/src/songinfo/artistbiography.cpp
+++ b/src/songinfo/artistbiography.cpp
@@ -70,8 +70,9 @@ void ArtistBiography::FetchInfo(int id, const Song& metadata) {
   NewClosure(reply, SIGNAL(finished()), [this, reply, id]() {
     reply->deleteLater();
 
+    QString json = reply->readAll();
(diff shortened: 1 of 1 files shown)


daddy33:
Working great. But why has this not been included in the master release so far?

hatstand:
The correct fix is to install a non broken version of QJson on your system.

daddy33:
Thanks a lot for the tip. Now the question is : where to find a non broken version of QJson for Manjaro/Arch?

Atrament666:
> Now the question is : where to find a non broken version of QJson for Manjaro/Arch?

You can clone https://github.com/flavio/qjson and checkout the old 0.8.1 release, build it and install it manually. Building the Clementine with this library really fixes the problem although I am not convinced that many Arch/Manjaro users will be willing to do that :( And I believe that the version 0.9.x will find it's way to the other distributions soon and the problem will affect even more users :(

Issue #5752, opened 2017-06-05
Nonrecursive library directory


Yanpas:
I suggest following feature. Mark some library directories as recursive / Non recursive. Here is mockup (sorry for my mad skillz, it would look as playlist table :) )
![2017-06-06 00_41_22-](https://cloud.githubusercontent.com/assets/7345761/26804820/07cc97b4-4a52-11e7-93da-51bc0e312adb.png)

If you give the greenlight to it, I will begin implementing it and will open PR.
By default I guess all directories would be recursive.

PR #5774, opened 2017-06-24
gtk icons by default on linux

Yanpas:
I'm not sure about this option, but this is how it was always for linux users, they won't be surprised after next clementine update


Yanpas:
BTW, I've opened proposal in issues #5752

Brli:
Should refer to as "system default icon" instead of "gtk icon" as this is actually an Qt app and not all DEs are gtk-based.

Please edit your title, thx.

PR #5793, opened 2017-07-22
Update fr.po

Fl0w3D:
Fixed translated search tag which shouldn't have been translated.


fitojb:
Thiese changes should be made at [Transifex ](https://www.transifex.com/projects/p/clementine/resource/clementineplayer) instead of PRs, as they would be overwritten each time the automatic Transifex update runs.

Fl0w3D:
Ok, correction done on Transifex.

PR #5841, opened 2017-08-28
↑↓ for "Up" and "Down"

comradekingu:


jbroadus:
For some reason, the QKeySequence constructor used in the generated code defaults to NativeText instead of PortableText. I assume this change was made and tested on a Mac, but it breaks the shortcut on Linux. It probably breaks Windows as well. I'm not sure if there's a way to force uic to specify PortableText.

The generated code also causes a compilation warning when it passes the unicode to QCoreApplication::translate, so at the very least, the translation should be disabled.

PR #5870, opened 2017-09-27
Submitting this snapcraft.yaml to upstream.

kz6fittycent:
Hopefully this file can provide a starting point for further development and creation of snap packages for other distributions or to provide an easier method of distribution of Clementine.


kz6fittycent:
See above comments.

Issue #5845, opened 2017-08-29
String language "Comment" doesn't appear in my language


Trudill:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [X] I checked the issue tracker for similar issues
- [X] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [ ] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Xubuntu 16.04
- Clementine version: 1.2.3

### Expected behaviour / actual behaviour
Hi and thanks for the good job !
I did create a dashboard on Xubuntu and every software has its comment in my language (French) but not Clementine wihich shows : "Plays music and last.fm streams" instead of "Écouter de la musique et les streams last.fm".
I did check the translations on transifex, but couldn't find this sentence.
I can force it, but it's better for everyone to have it in its language ;)

### Steps to reproduce the problem (only for bugs)
Create a dashboard on Xubuntu
Insert Clementine
See attached file (the rest of the icons are in my language)
![bug_clementine](https://user-images.githubusercontent.com/3287253/29818572-3575e316-8cbd-11e7-831a-ac6722a5c1c0.png)

hatstand:
I think you'd need to add a French translation here: https://github.com/clementine-player/Clementine/blob/master/dist/clementine.desktop

Trudill:
Thanks, I did that, but I think it would be nicer if it can be integrated to Transifex as translators and developers are not always the same one.
Hypnotoad forever !!!

PR #5874, opened 2017-10-03
Updated qt bug report page

mananaBanana:
The original bug report page in the comments ([this link](https://bugreports.qt-project.org/browse/QTBUG-32789)) is unreachable.


hatstand:
Please revert the translations and do them on transifex instead, otherwise they will be overwritten.

mananaBanana:
Done! Will add them later on transifex. I thought that this file cannot be updated using transifex (according to #5845). Was that fixed?

hatstand:
Oh, you're right, sorry! It would still be better as a separate pull request though.

mananaBanana:
Kein Problem! Will do so. I'll have to push to a new branch, right?

hatstand:
Manually merged in: https://github.com/clementine-player/Clementine/compare/c024b1340d6d...ba8a82bffab4

Issue #5868, opened 2017-09-23
Clementine QT5 remote


karl69942:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [ x] I checked the issue tracker for similar issues
- [ x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [x ] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System:
- Clementine version: 

### Steps to reproduce the problem (only for bugs)

Hi, I run Manjaro Linux which is an Arch variant, but I think I had this problem on Fedora also. 
I've tried this a few times & just installed clementine-qt5 today to try it again, my remote app won't connect to it... it will appear in the list of the app sometimes but it won't connect.
I always run the latest version, in this case it would be the Arch AUR version.

https://aur.archlinux.org/packages/clementine-qt5-git/
clementine 1.3.1.r231.g57d90ab-1

trougnouf:
I was able to replicate this problem with the "Only allow connections from the local network" setting enabled (I believe that's the default). The app gets disconnected as soon as I enable this setting back.

My local IP is in the 192.168.0.** range

btw the qt5 code is hosted in this same repo under the qt5 branch: https://github.com/clementine-player/Clementine/tree/qt5

kenny-wettlaufer:
QT5 Clementine now works OK with the remote but it crashes a lot...

PR #5896, opened 2017-11-04
fixed issue #5868: IpIsPrivate returns false when IPv4 are translated to IPv6

trougnouf:
fixed issue #5868: IPs were not recognized as private when Qt translates ipv4 to ipv6

As a result, the Network Remote would disconnect whenever the option to "Only allow connections from the local network" was selected.

Added translated IPs to NetworkRemote::IpIsPrivate


trougnouf:
@hatstand can you review this? It's a pretty simple fix.
Thanks in advance for your time!

hatstand:
I'd prefer to merge it to master first.

trougnouf:
Thank you for your response. As far as I know this only affects Qt5 (it worked before, and https://bugreports.qt.io/browse/QTBUG-47348 confirms the behavior changed in Qt5). Should I make a pull request on the master branch anyway?

PR #5942, opened 2018-01-03
Add appearance option to hide playlist filter toolbar

vikramambrose:
Gives the user the option of a cleaner looking interface if the filter toolbar isn't used.

I've tried to follow existing coding style to the best of my knowledge.

PS: This is my first ever pull request so be kind :)


ronjouch:
@hatstand @vikramambrose thanks for this! Works fine for me using latest AUR `clementine-git` 👍.

PR #5956, opened 2018-01-09
Replace Nokia based FancyTabWidget with standard QTabWidget

vikramambrose:
This is a rewrite of the sidebar widget. The old sidebar widget was written by Nokia 10 years ago and was a hand made hardcoded mess of widgets trying to emulate a QTabWidget. 
    
This commit completely replaces it with a widget of the same name (FancyTabWidget) but is a much simpler subclass of the standard QTabWidget allowing for a 50% code reduction.
    
There is still some manual draw code copied over to get the exact same look of the previous widget but this is not strictly necessary and can be later refactored with simpler drawControl code based on styling preferences or if a pixel perfect match of the Nokia version is not necessary.
    
Benefits:
  - Easier to maintain (50% code reduction)
  - Easier to understand (standard QTabWidget mechanics)
  - Built-in support for re-arranging and closing tabs (Not easily possible with the Nokia version)

The drag-n-drop animation for the custom modes (Large and Small Sidebar) isn't the prettiest thing yet but it works well.

I've added save and restore QSettings for re-arranging the tabs but not for closing tabs.

Adding back the floating NowPlayingWidget was a bit awkward to do with a standard QTabWidget. But I achieved the same look by transparently adding a parent VBoxLayout to each page. Seems to work well.

Has only been tested on Ubuntu 16.04. Looks perfect there. Needs to be tested on other platforms.

### DIFF
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -255,37 +255,37 @@ MainWindow::MainWindow(Application* app, SystemTrayIcon* tray_icon, OSD* osd,
           SLOT(AddToPlaylist(QMimeData*)));
 
   // Add tabs to the fancy tab widget
-  ui_->tabs->AddTab(global_search_view_,
+  ui_->tabs->addTab(global_search_view_,
                     IconLoader::Load("search", IconLoader::Base),
                     tr("Search", "Global search settings dialog title."));
-  ui_->tabs->AddTab(library_view_,
+  ui_->tabs->addTab(library_view_,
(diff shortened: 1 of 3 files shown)


ronjouch:
> *Has only been tested on Ubuntu 16.04. Looks perfect there. Needs to be tested on other platforms.*

@vikramambrose looks good here on Arch Linux through AUR package `clementine-git` 👍. Tested creating/closing playlist, re-ordering.

PR #5960, opened 2018-01-11
Add search filter to playlist list page

vikramambrose:
This allows power users who keep 100s of playlists to easily find a playlist either by searching for directory name, playlist name or a song artist/title a playlist might contain

This commit also adds track "$artist - $title" items to each playlist tree node


vikramambrose:
Thanks @hatstand

Issue #2617, opened 2013-12-10
Can't search for empty genre


Clementine-Issue-Importer:
_From [alphadel...@gmail.com](https://code.google.com/u/107136459110240960141/) on January 16, 2012 09:08:52_

What steps will reproduce the problem? 1.  Make a new smart playlist
2.  Add a search term for "Genre equals" and leave the third field empty What is the expected output? What do you see instead? I expect to see only tracks which have no set Genre.  Instead I see all tracks regardless of their Genre field. What version of the product are you using? On what operating system? 1.0 on Linux Please provide any additional information below. This app is the best music player, hands down.  Amarok 1.4 is dead!  Long live Clementine!

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=2617_

kleinfreund:
Smart Playlists currently can’t contain empty values in the Search Terms field.

It would be useful to either

- be able to leave the search term empty or
- have `is empty` and `is not empty` qualifiers

This would allow users to filter out tracks that are part of an album or tracks that don’t have a genre, yet.

Any chance this will get implemented?

kleinfreund:
I cloned the repository and tried implementing this on my own.

My first naive approach was changing the following line to `return true;`: https://github.com/clementine-player/Clementine/blob/e5ab3e786f9adde12cec3cc90cfe8c1cc6b06320/src/smartplaylists/searchterm.cpp#L146

This will crash the program when opening the create smart playlist dialog.

Next, I tried to add the operators `Op_Empty` and `Op_NotEmpty`. However, I didn’t find an easy way to remove the text field for these operators. Currently, the type of the current search term is determined by the field (e.g. Artist, [see here](https://github.com/clementine-player/Clementine/blob/e5ab3e786f9adde12cec3cc90cfe8c1cc6b06320/src/smartplaylists/searchterm.cpp#L167-L196)). The default type is `Type_Text`; however, `Op_Empty` and `Op_NotEmpty` are not of type text. Yet, they need to appear in the same combobox along with operators that require a text field.

**TL;DR:** The type of the input field should be determined by the selected operator, not the selected field. This would allow mixing input field types.

kleinfreund:
This was resolved in #5985.

PR #5985, opened 2018-02-12
Smart Playlists: Add empty/not empty operators

kleinfreund:
This adds two new operators "empty" and "not empty" to the smart playlist search terms, fixing #2617.

Disclaimer: I am completely inexperienced with regards to native app development and Qt. I have basic knowledge of C++.

A few notes:

- This pull request does not address the culprit that search terms cannot be empty. In general, it should be allowed to use empty value to create queries like *Album equals `$EMPTYSTRING`*. However, some internal logic depends on the behavior of considering empty fields as empty. In particular, starting the smart playlist dialog will crash if empty search terms are considered valid (i.e. changing [this line](https://github.com/clementine-player/Clementine/blob/e5ab3e786f9adde12cec3cc90cfe8c1cc6b06320/src/smartplaylists/searchterm.cpp#L146) to `return true;`)
- Currently, the search field (e.g. Album, Artist, etc.) determines the input field (e.g. text, date, etc.) when it should be a combination of search field and operator.
- This pull request might be incomplete. I tried to cover all relevant areas when adding new entities like `page_empty` or the operators.

So far, have a nice day. I hope this finds its way to a merge.

### DIFF
--- a/src/smartplaylists/searchtermwidget.cpp
+++ b/src/smartplaylists/searchtermwidget.cpp
@@ -163,7 +163,11 @@ void SearchTermWidget::FieldChanged(int index) {
       page = ui_->page_rating;
       break;
     case SearchTerm::Type_Text:
-      page = ui_->page_text;
+      if (ui_->op->currentIndex() == 4 || ui_->op->currentIndex() == 5) {
+        page = ui_->page_empty;
+      } else {
+        page = ui_->page_text;
+      }
       break;
     case SearchTerm::Type_Invalid:
       page = nullptr;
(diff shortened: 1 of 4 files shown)


src/smartplaylists/searchtermwidget.cpp
```diff
@@ -163,7 +163,11 @@ void SearchTermWidget::FieldChanged(int index) {
       page = ui_->page_rating;
       break;
     case SearchTerm::Type_Text:
-      page = ui_->page_text;
+      if (ui_->op->currentIndex() == 4 || ui_->op->currentIndex() == 5) {
```
hatstand: What are these magic numbers about?

kleinfreund: Same as in `SearchTermWidget::OpChanged`. The application uses the index of the entry in the combobox. I didn’t find a way how I would be able to do this check otherwise.

kleinfreund: I just mirrored what the original code does here: https://github.com/kleinfreund/Clementine/blob/4ab6f359d50bb23c116b6dca2a8fc02f26f5cd7a/src/smartplaylists/searchtermwidget.cpp#L214-L216

Issue #5993, opened 2018-02-17
clementine-qt5 fails at the last.fm authentication (invalid key) due to a different name


jorgicio:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [ x] I checked the issue tracker for similar issues
- [ x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [ x] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Gentoo
- Clementine version: 1.3.1 

### Expected behaviour / actual behaviour

Can log flawlessly to my last.fm account / It throws the message of "Invalid key".

### Steps to reproduce the problem (only for bugs)
* Open Clementine
* Go to the last.fm menu and try to login as usual
* It redirects to the page to authenticate, but it shows an "Invalid key" message. Tried with Firefox, Chrome and Opera.

As I reported the issue in the Gentoo tracker, I got this answer. The Clementine-qt5 app is treated as a different application aside the Clementine one. Hence the message.

More information in the screenshot.

![pantallazo-2018-02-17 11-45-59](https://user-images.githubusercontent.com/554710/36342215-d0ed5e98-13d9-11e8-94c4-3724efa08c9a.png)

trougnouf:
Same issue, I just noticed that my last update was on February 2nd.

trougnouf:
@Chocobozzz 

Logging in now leads to the page https://www.last.fm/api/auth?api_key=<api_key>=http://localhost:42437/ , clicking the allow button leads to https://localhost:41505/?token=<token> then Clementine sends a failure popup and crashes.

__logging_message__13:57:45.057 DEBUG unknown                          "<?xml version=\"1.0\" encoding=\"utf-8\"?>\n<lfm status=\"failed\">\n  <error code=\"6\">Invalid parameters - Your request is missing a required parameter</error>\n</lfm>\n"

I'm not sure this is caused by clementine as opposed to lastfm

Chocobozzz:
@trougnouf try to open the link in chrome?

trougnouf:
@Chocobozzz It works with Chrome. Firefox threw an error about SS(H/L) key too long I believe, I can replicate and provide the exact output if necessary.

Chocobozzz:
[...]

PR #5994, opened 2018-02-19
Fix "Invalid key" error in lastfm auth

Chocobozzz:
Fixes https://github.com/clementine-player/Clementine/issues/5993#issuecomment-366507718


trougnouf:
Thank you!

PR #6000, opened 2018-02-24
Hide TrackSliderPopup on exit from popup instead of from TrackSliderSlider

aerusso:
The previous behavior exited on leaving TrackSliderSlider, which causes the popup to flicker, as it is shown and hidden rapidly, at least for me (on Debian unstable).


aerusso:
@hatstand Shoot. The situation is more complicated than I realized. The TrackSliderPopup does not always generate a leaveEvent (causing a regression for, presumably, most users). I am testing a patch that reverts the above and validates the leave event directly instead of relying on TrackSliderPopup to generate a leave event.

hatstand:
Reverted for now

PR #6001, opened 2018-02-24
Hide TrackSliderPopup on appropriate leaveEvent

aerusso:
The previous fix introduces a regression on some platforms which did not generate a leaveEvent when the TrackSliderPopup was displayed. The previous fix is reverted and logic is introduced to identify bonafide mouse motion out of the TrackSliderSlider.

Sorry about the noise on this. I only have two systems to check this on, but since this patch is much simpler than my first one, I expect it is less likely to cause a regression.


hatstand:
Sorry, I reverted the previous one first as it makes the break/fix more obvious. Can you rebase this?

aerusso:
Rebased, as requested.

PR #6021, opened 2018-03-09
Workaround for broken CD playback in Qt5

jonaski:
QUrl() does not accept a single number as the host, it translates it to 0.0.0.1.
This is a simple workaround instead of replacing QUrl with QString or QVariant.
The first commit fixes it by adding a single character after the tracknumber, ie: cdda://1a for track number one. This URL will still work with gstreamer and can be played. The second commit removes the added "a" from the URL, just in case it would break in another version of gstreamer.

### DIFF
--- a/src/engines/gstenginepipeline.cpp
+++ b/src/engines/gstenginepipeline.cpp
@@ -181,8 +181,17 @@ bool GstEnginePipeline::ReplaceDecodeBin(const QUrl& url) {
         spotify_server, "StartPlayback", Qt::QueuedConnection,
         Q_ARG(QString, url.toString()), Q_ARG(quint16, port));
   } else {
+    QByteArray uri;
+    if (url.scheme() == "cdda") {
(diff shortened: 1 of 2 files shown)


src/engines/gstenginepipeline.cpp
```diff
@@ -181,9 +181,16 @@ bool GstEnginePipeline::ReplaceDecodeBin(const QUrl& url) {
         spotify_server, "StartPlayback", Qt::QueuedConnection,
         Q_ARG(QString, url.toString()), Q_ARG(quint16, port));
   } else {
+    QByteArray uri;
+    if (url.scheme() == "cdda") {
+      QString str = url.toString();
+      str.remove(str.lastIndexOf(QChar('a')), 1);
```
EGD: Perhaps it's better to use `QString.leftRef`?
`uri = str.leftRef(str.lastIndexOf(QChar('a'))).toUtf8();`

PR #6023, opened 2018-03-10
Fix channels loading for DI.fm services after API domain changes

Pulfer:
Clementine no longer loads channels list from DI.FM services. The issue is caused by obsolete api.v2.audioaddict.com domain usage. api.audioaddict.com should be used instead.

### DIFF
--- a/src/internet/digitally/digitallyimportedclient.cpp
+++ b/src/internet/digitally/digitallyimportedclient.cpp
@@ -38,7 +38,7 @@ const char* DigitallyImportedClient::kAuthUrl =
     "http://api.audioaddict.com/v1/%1/members/authenticate";
 
 const char* DigitallyImportedClient::kChannelListUrl =
-    "http://api.v2.audioaddict.com/v1/%1/mobile/"
+    "http://api.audioaddict.com/v1/%1/mobile/"
     "batch_update?asset_group_key=mobile_icons&stream_set_key=";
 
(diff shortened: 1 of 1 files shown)


sethhochberg:
Hi,

Lead engineer at DI.FM here - we apologize for breaking this! Our usage audit on the `v2` subdomain didn't turn up anything and it was internally retired many years ago, so we removed it recently. You can expect `https://api.audioaddict.com/v1` to provide a stable (or at least generally backwards compatible) interface.

We don't support public usage of our API right now but intend to provide a documented API in the future, so we try to keep an eye on the various good-faith third party users of our systems and offer a hand when we can. Please don't hesitate to reach out if we can assist with integrating things more efficiently in the future, we'll do our best to offer tips.

thodges314:
Is this working?  Do I need to download and compile from source?  I'm using the mac version and I can't update to listen to the new Synthwave station.

sethhochberg:
@thodges314 it does not look like there has been a new release of Clementine since this PR was merged. You will likely need to compile from source (or find a more recently-built version than the official downloads) to get the fix above.

thodges314:
@sethhochberg Thanks.  I grabbed yesterday's dev build.  It works great.

JulianVolodia:
@jjberry314 2 questions. First, I think the answer is yes - did you really try the version with fix?
Second, could you provide debug log while using more verbose (indeed debug) version for Win? 

Put that with info in the issue, not here, ok? Thanks.

PR #6028, opened 2018-03-15
Bring back album cover fetching from discogs.

jonaski:
Discogs has a large collection of album covers that can be a nice addition, especially for certain albums. 

However they use rate limiting to limit the requests to 60 per minute identified by the source IP: https://www.discogs.com/developers/#page:home,header:home-rate-limiting
There is no point in using it when fetching covers for all albums in the cover manager (Fetch Missing Covers) since they throttle the connection to fast. But it can be very useful and nice addition for manually setting album cover in the cover manager, and for automatically or manually downloading when playing in the now playing widget.
To work around this I've created a variable called fetchall in the cover provider class that can be set to true if the provider should be used for "Fetch Missing Covers" in the cover manager, or false if not.
The 'fetchall' variable in the album cover fetcher will identify if "Fetch Missing Covers" was pressed or not. Please suggest another way of doing this if it should be done differently.

I have registered the access key and secret to Clementine and encoded it using base64.

The code is using QUrlQuery, QJsonDocument and QJsonObject that are not available in Qt4 so I'm creating the pull request on the Qt5 branch instead of backporting it to Qt4. It does not need extra qjson library dependencies.


src/covers/albumcoverfetchersearch.cpp
```diff
@@ -63,6 +63,10 @@ void AlbumCoverFetcherSearch::TerminateSearch() {
 
 void AlbumCoverFetcherSearch::Start(CoverProviders* cover_providers) {
   for (CoverProvider* provider : cover_providers->List()) {
+    // Skip provider if it does not have fetchall set, and we are doing fetchall - "Fetch Missing Covers".
+    if ((provider->fetchall() == false) && (fetchall_ == true)) {
```
hatstand: `if (fetchall_ && !provider->fetchall())`

src/covers/albumcoverfetcher.h
```diff
@@ -92,6 +92,8 @@ class AlbumCoverFetcher : public QObject {
 
   void Clear();
 
+  bool fetchall_ = false;
```
hatstand: This should be private

src/covers/albumcoverfetchersearch.h
```diff
@@ -51,6 +51,8 @@ class AlbumCoverFetcherSearch : public QObject {
 
   CoverSearchStatistics statistics() const { return statistics_; }
 
+  bool fetchall_ = false;
```
hatstand: private

src/covers/coverprovider.h
```diff
@@ -35,10 +35,12 @@ class CoverProvider : public QObject {
   Q_OBJECT
 
  public:
-  explicit CoverProvider(const QString& name, QObject* parent);
+  explicit CoverProvider(const QString& name, const bool& fetchall,
```
hatstand: Just use `bool`

src/covers/discogscoverprovider.cpp
```diff
@@ -0,0 +1,373 @@
...
+    "QkJNb2tMVXVUVFhSRWRUVmZDc0ZGamZmSWRjdHZRVno=";
+
+DiscogsCoverProvider::DiscogsCoverProvider(QObject* parent)
+    : CoverProvider("Discogs", false, parent),
+      network_(new NetworkAccessManager(this)) {}
+
+bool DiscogsCoverProvider::StartSearch(const QString& artist,
+                                       const QString& album, int s_id) {
+  DiscogsCoverSearchContext* s_ctx = new DiscogsCoverSearchContext;
+  if (s_ctx == nullptr) return false;
```
hatstand: This can never be `nullptr` so there's no point in the check

src/covers/discogscoverprovider.cpp
```diff
@@ -0,0 +1,373 @@
...
+}
+
+void DiscogsCoverProvider::CancelSearch(int id) {
+  delete requests_search_.take(id);
+}
+
+bool DiscogsCoverProvider::StartRelease(DiscogsCoverSearchContext* s_ctx,
+                                        int r_id, QString resource_url) {
+  DiscogsCoverReleaseContext* r_ctx = new DiscogsCoverReleaseContext;
+  if (r_ctx == nullptr) return false;
```
hatstand: Again, this can never be null

src/covers/discogscoverprovider.cpp
```diff
@@ -0,0 +1,373 @@
...
+                           QUrl::toPercentEncoding(arg.second));
+    query_items << QString(encoded_arg.first + "=" + encoded_arg.second);
+    url_query.addQueryItem(encoded_arg.first, encoded_arg.second);
+  }
+
+  // Sign the request
+  const QByteArray data_to_sign =
+      QString("GET\n%1\n%2\n%3")
+          .arg(url.host(), url.path(), query_items.join("&"))
+          .toLatin1();
```
hatstand: Should this actually be latin1, or ASCII, or UTF-8?

src/covers/discogscoverprovider.cpp
```diff
@@ -0,0 +1,373 @@
...
+      QByteArray::fromBase64(kSecretKeyB64), data_to_sign));
+
+  // Add the signature to the request
+  url_query.addQueryItem("Signature",
+                         QUrl::toPercentEncoding(signature.toBase64()));
+
+  url.setQuery(url_query);
+  QNetworkReply* reply = network_->get(QNetworkRequest(url));
+
+  NewClosure(reply, SIGNAL(error(QNetworkReply::NetworkError)), this,
```
hatstand: Pretty sure signal arguments don't work with `NewClosure`

src/covers/discogscoverprovider.cpp
```diff
@@ -0,0 +1,373 @@
...
+             reply, s_ctx->id, r_ctx->id);
+  NewClosure(reply, SIGNAL(finished()), this,
+             SLOT(HandleReleaseReply(QNetworkReply*, int, int)), reply,
+             s_ctx->id, r_ctx->id);
+}
+
+void DiscogsCoverProvider::HandleSearchReply(QNetworkReply* reply, int s_id) {
+  reply->deleteLater();
+
+  DiscogsCoverSearchContext* s_ctx;
```
hatstand: Assign at the same time as declaration

src/covers/discogscoverprovider.cpp
```diff
@@ -0,0 +1,373 @@
...
+  if (!requests_search_.contains(s_id)) {
+    return;
+  }
+  s_ctx = requests_search_.value(s_id);
+  if (s_ctx == nullptr) return;
```
hatstand: This check is redundant

src/covers/discogscoverprovider.cpp
```diff
@@ -0,0 +1,373 @@
...
+
+  QString json_string = reply->readAll();
+  QByteArray json_bytes = json_string.toLocal8Bit();
```
hatstand: why `toLocal8Bit()` not `toUtf8()`?

src/covers/discogscoverprovider.cpp
```diff
@@ -0,0 +1,373 @@
...
+    return;
+  }
+  if (!json_doc.isObject()) {
+    qLog(Error) << "Discogs: JSON is not an object.";
+    EndSearch(s_ctx);
+    return;
+  }
+
+  QJsonObject json_obj = json_doc.object();
+  if (json_obj.isEmpty()) {
```
hatstand: Do all of these checks in one conditional

src/covers/discogscoverprovider.cpp
```diff
@@ -0,0 +1,373 @@
...
+
+  for (const QVariant& result : results) {
+    QVariantMap result_map = result.toMap();
+    int r_id = 0;
+    QString title;
+    QString resource_url;
+    if ((!result_map.contains("id")) || (!result_map.contains("resource_url")))
+      continue;
+
+    if (result_map.contains("id")) {
```
hatstand: You've done these checks already

src/covers/discogscoverprovider.cpp
```diff
@@ -0,0 +1,373 @@
...
+  }
+
+  QVariantList results = reply_map["results"].toList();
+  int i = 0;
+
+  for (const QVariant& result : results) {
+    QVariantMap result_map = result.toMap();
+    int r_id = 0;
+    QString title;
+    QString resource_url;
```
hatstand: Move these to where you use them

src/covers/discogscoverprovider.cpp
```diff
@@ -0,0 +1,373 @@
...
+    if (result_map.contains("id")) {
+      r_id = result_map["id"].toInt();
+    }
+    if (result_map.contains("title")) {
+      title = result_map["title"].toString();
+    }
+    if (result_map.contains("resource_url")) {
+      resource_url = result_map["resource_url"].toString();
+    }
+    StartRelease(s_ctx, r_id, resource_url);
```
hatstand: Just do something like:
```
if (result_map.contains("title") && result_map.contains(... {
  StartRelease(s_ctx, result_map["id"].toInt(),...);
  ++i
}
```

src/covers/discogscoverprovider.cpp
```diff
@@ -0,0 +1,373 @@
...
+void DiscogsCoverProvider::HandleReleaseReply(QNetworkReply* reply, int s_id,
+                                              int r_id) {
+  reply->deleteLater();
+
+  DiscogsCoverReleaseContext* r_ctx;
+  if (!requests_release_.contains(r_id)) {
+    return;
+  }
+  r_ctx = requests_release_.value(r_id);
+  if (r_ctx == nullptr) return;
```
hatstand: More redundant checks in this function.

src/covers/discogscoverprovider.cpp
```diff
@@ -0,0 +1,373 @@
...
+
+  DiscogsCoverSearchContext* s_ctx;
```
hatstand: More declarations separate from assignments too.

src/covers/discogscoverprovider.cpp
```diff
@@ -0,0 +1,373 @@
...
+  }
+  s_ctx = requests_search_.value(s_id);
+  if (s_ctx == nullptr) return;
+
+  EndSearch(s_ctx, r_ctx);
+}
+
+void DiscogsCoverProvider::EndSearch(DiscogsCoverSearchContext* s_ctx,
+                                     DiscogsCoverReleaseContext* r_ctx) {
+  (void)requests_release_.remove(r_ctx->id);
```
hatstand: You don't need the `(void)`

hatstand: You probably just want to do: `delete requests_release_.take(r_ctx->id);` anyway

src/covers/discogscoverprovider.cpp
```diff
@@ -0,0 +1,373 @@
...
+  delete r_ctx;
+
+  if (s_ctx == nullptr) return;
```
hatstand: Redundant

src/covers/discogscoverprovider.cpp
```diff
@@ -0,0 +1,373 @@
...
+  url_query.addQueryItem("Signature",
+                         QUrl::toPercentEncoding(signature.toBase64()));
+
+  url.setQuery(url_query);
+  QNetworkReply* reply = network_->get(QNetworkRequest(url));
+
+  NewClosure(reply, SIGNAL(error(QNetworkReply::NetworkError)), this,
+             SLOT(SearchRequestError(QNetworkReply::NetworkError,
+                                     QNetworkReply*, int)),
+             reply, s_ctx->id);
```
hatstand: I don't think you even need all these ids and the two `QHash`. You can just pass the context objects around instead via `NewClosure`.

src/covers/discogscoverprovider.h
```diff
@@ -0,0 +1,97 @@
...
+};
+Q_DECLARE_METATYPE(DiscogsCoverReleaseContext)
+
+class DiscogsCoverProvider : public CoverProvider {
+  Q_OBJECT
+
+ public:
+  explicit DiscogsCoverProvider(QObject* parent = nullptr);
+
+  static const char* kUrlSearch;
```
hatstand: These can be private I think. In fact, they probably don't need to be in the header.

src/ui/albumcovermanager.cpp
```diff
@@ -424,6 +424,7 @@ bool AlbumCoverManager::ShouldHide(const QListWidgetItem& item,
 }
 
 void AlbumCoverManager::FetchAlbumCovers() {
+  cover_fetcher_->fetchall_ = true;
```
hatstand: This shouldn't be a member variable at all but a parameter to the cover fetch method, possibly with a default.

jonaski: That's what I intended in the first place, but confused somewhere. I've redone it now, I'm creating a variable on the request instead, this allows automatically downloading while playing from discogs even when the user could be doing fetch all missing covers in the cover manager.

src/covers/albumcoverfetcher.h
jonaski: Agree

src/covers/albumcoverfetchersearch.h
jonaski: Agree

src/covers/discogscoverprovider.cpp
jonaski: Agree

src/covers/discogscoverprovider.cpp
jonaski: Agree

src/covers/discogscoverprovider.cpp
jonaski: Won't work without, throws errors. It works like this. it's already done like this in other parts of the code like gpodder.

src/covers/discogscoverprovider.cpp
jonaski: Agree

src/covers/discogscoverprovider.cpp
jonaski: Agree

src/covers/discogscoverprovider.cpp
jonaski: Agree

src/covers/discogscoverprovider.cpp
jonaski: Agree

src/covers/discogscoverprovider.cpp
jonaski: Agree

src/covers/discogscoverprovider.cpp
jonaski: Agree

src/covers/discogscoverprovider.cpp
jonaski: Agree

src/covers/discogscoverprovider.cpp
jonaski: Agree

src/covers/discogscoverprovider.cpp
jonaski: Agree

src/covers/discogscoverprovider.cpp
jonaski: Agree

src/covers/discogscoverprovider.cpp
jonaski: Agree

src/covers/discogscoverprovider.h
jonaski: Changing to private

src/covers/coverprovider.h
jonaski: Doesn't work. All the cover managers won't compile then.

PR #6060, opened 2018-05-23
Add option to verify subsonic server certificate.

ftiede:
Defaults to true, as it is safer to have a server certificate verified,
even more so, if the server is used over an insecure WAN link.
During subsonic configuration the checkbox can be deactivated, so that
no certificate verification will occur when talking to a subsonic
server, allowing for self-signed certificates.

With the proliferation of let's encrypt certificates there's probably
less need for this option but it has been requested and hard-coding
verify-off is IMHO bad security practice.
If a valid certificate has been installed, the configuration file can be
modified manually and after a restart Clementine will perform a proper
server certificate verification.

The patch might need some UI polishing and asks for string translations
but is operational so far.

### DIFF
--- a/src/internet/subsonic/subsonicservice.cpp
+++ b/src/internet/subsonic/subsonicservice.cpp
@@ -195,6 +195,10 @@ void SubsonicService::ReloadSettings() {
   username_ = s.value("username").toString();
   password_ = s.value("password").toString();
   usesslv3_ = s.value("usesslv3").toBool();
+  if (s.contains("verifycert"))
+    verifycert_ = s.value("verifycert").toBool();
+  else
+    verifycert_ = true;
 
   Login();
 }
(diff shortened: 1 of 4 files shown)


src/internet/subsonic/subsonicservice.cpp
```diff
@@ -195,6 +195,10 @@ void SubsonicService::ReloadSettings() {
   username_ = s.value("username").toString();
   password_ = s.value("password").toString();
   usesslv3_ = s.value("usesslv3").toBool();
+  if (s.contains("verifycert"))
```
hatstand: `verifycert_ = s.value("verifycert", true).toBool();` is simpler than this conditional

ftiede: Didn't know it existed, will change ASAP.

ftiede:
As I said: "probably". Whoever needs to use an "invalid" certificate now has the option to do so with a checkbox instead of a hack in the sources.
Using subsonic (or its forks) over the internet without properly checking the server certificate is IMHO risking his credentials to anyone eavesdropping on the connection.

I don't know the issue numbers anymore, but requests were there and there must be a reason why my distribution's package did say QSslSocket::VerifyNone at that line of code. So I patched it.

Issue #5364, opened 2016-04-29
Clementine 1.3.1 does not play ogg files with extension .oga


RipClaw2971:
Clementine 1.3.1 does not play ogg files with extension .oga. If you rename it to .ogg it works.

If you got a podcast (Example: http://freakshow.fm/feed/oga/) with such files and you download one episode Clementine hangs.

ArnaudBienner:
What is your operating system? Windows? Mac OS X? Linux (if so, which distro)?

At a first glance, this looks like something missing in Taglib, but latest version should support file with .oga extension (looking at their source code).
I will check if upgrading taglibg can help (we currently use 1.10 but 1.11 was released recently), but for Linux where we use the version installed on the system, I'm not sure we can do a lot until distribution updates taglib on their own.
Any case, it should fail currently but not hang: I will check why this happens.

birdip:
My Clementine 1.3.1 can't play .oga files either. I'm runnig Arch Linux 64bit with taglib 1.11 installed.
But my Clementine doesn't hang, it just doesn't play the file.

pyjamashark:
Hi, Clementine 1.3.1 isn't playing oga files for me either  ;(- I'm running Os X El Capitan

dtonhofer:
This is the case with Clementine 1.3.1 on Fedora 27. .oga is the (correct) suffix for audio-only .ogg files. What would cause Clementine to reject those files? 
(Please fix because I am appalled by complexifeaturificated Amarok)

PR #6076, opened 2018-06-06
Audio file detection by content supported by newer taglib

jonaski:
Since Clementine has builtin taglib anyway it might as well take advantage of the audio file detection feature not yet released in an official version of taglib.

- Bump taglib to current git revison
- Use taglib todetect if the file is a valid audio file using taglib fileref in song::InitFromFilePartial()
- Use builtin taglib as default


jonaski:
This might fix #5364

Issue #5352, opened 2016-04-20
Playback does not work on USB DAC with 1.3 and 1.3.1


SamuelGilbert42:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [x] I checked the issue tracker for similar issues
- [x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [x] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Arch Linux
- Clementine version: 1.3 and 1.3.1 (built from https://builds.clementine-player.org/)

### Expected behaviour / actual behaviour

After selecting the my DAC (NuForce USB Audio Analog Stereo) from Tools->Preferences->Playback->Audio Output, I should be able to hear music play with the headphones connected to my DAC.

It worked yesterday, but it does not work today.  My guess is that Clementine isn't using the proper end-point.  My DAC has 2 (Analog and SPDIF).  With the old interface in Clementine 1.2.3, I could write the proper ALSA device name (hw:Audio,0), but with this new interface, I can no longer specify nor select the end-point.

I confirmed that it's not the DAC or an hardware problem since playback works with VLC and mplayer.

### Steps to reproduce the problem (only for bugs)
1. Select "NuForce USB Audio Analog Stereo" from Tools->Preferences->Playback->Audio Output
2. Start playback of a track
3. Hear nothing, while seeing the song scroll-bar advance.  The issue isn't systematic.  It worked yesterday.
![clementine_output_selection_bug](https://cloud.githubusercontent.com/assets/14999789/14683471/d3d8e45a-06f8-11e6-994f-b92622dfea0e.png)

josephk84:
Maybe a compilation issue? 
With Clementine from arch "extra" repo it works for me (I have a HRT Musicstreamer II+).
Anyway I agree with your "not possible to specify alsa output" rant.

Ferroin:
It could just as easily be poor hardware design (I've seen USB audio cards that have both analogue and   optical digital outputs, and put the optical output first in enumeration), or even something with Pulseaudio (or ALSA) picking the wrong default output device (which I have regularly recurring issues with).

SamuelGilbert42:
> Maybe a compilation issue?
I don't think so since I was able to replicate the issue with 1.3 (Arch), 1.3.1 (https://builds.clementine-player.org/), and git.

[...]

PR #6079, opened 2018-06-07
Add ALSA Device Finder

### LINKED ISSUE #5344: Clementine 1.3: no more alsa output card options?
Hello,

with the new Clementine v1.3 , I sadly discovered that we are not able  to configure ALSA output card parameters anymore.
I fortunately own an audiophile DAC and I used to output streams to "hw:2,0" to avoid resampling.
Now this option is gone, while two new options are added ("Sample Rate" and "Minimal Buffer fill").
To me this is a regression. Before it worked well, now high resolution files are downsampled in any case, sounding worse than  44.1KHz ones.
This is a pity, since Clementine is by far the best Linux audio player.

Kind Regards
Alex

jonaski:
- Lists alsa devices only when Clementine was not compiled with pulseaudio or pulseaudio is disabled on the system.
- No advanced device string setting to avoid confusing users

Fixes #5344 and #5352


src/engines/gstengine.cpp
```diff
@@ -158,9 +162,22 @@ void GstEngine::InitialiseGstreamer() {
     plugin_names.insert(plugin.name);
   }
 
-  QList<DeviceFinder*> device_finders;
+  DeviceFinder* finder_pulse(nullptr);
```
hatstand: Use a `unique_ptr` and `unique_ptr::release()` to get RAII for this.

src/engines/alsadevicefinder.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+#include <core/logging.h>
+
+#include "alsadevicefinder.h"
+#include "devicefinder.h"
+
+AlsaDeviceFinder::AlsaDeviceFinder() : DeviceFinder("alsasink") {}
+
+QList<DeviceFinder::Device> AlsaDeviceFinder::ListDevices() {
+  QList<Device> ret;
```
hatstand: We don't have to write C89 here, please move the declarations next to their usages :-)

src/engines/alsadevicefinder.cpp
```diff
@@ -0,0 +1,110 @@
...
+  while (true) {
+    result = snd_card_next(&card);
+    if (result < 0) {
+      qLog(Error) << "Unable to get soundcard:" << snd_strerror(result);
+      break;
+    }
+    if (card < 0) break;
+
+    char name[32];
+    sprintf(name, "hw:%d", card);
```
hatstand: `snprintf`

src/engines/alsadevicefinder.cpp
```diff
@@ -0,0 +1,110 @@
...
+    if (result < 0) {
+      qLog(Error) << "Unable to open soundcard" << card << ":"
+                  << snd_strerror(result);
+      continue;
+    }
+    result = snd_ctl_card_info(handle, cardinfo);
+    if (result < 0) {
+      qLog(Error) << "Control hardware failure for card" << card << ":"
+                  << snd_strerror(result);
+      snd_ctl_close(handle);
```
hatstand: You could use boost scope exit to make these handle closures safer, something like:
```c++
BOOST_SCOPE_EXIT(&handle) {
  snd_ctl_close(handle);
} BOOST_SCOPE_EXIT_END
```

PR #6097, opened 2018-07-05
Add musicbrainz integration to the cd ripper

### LINKED ISSUE #6045: Clementine Does Not Preserve Metadata When Ripping From CD
### System information
- Operating System: Ubuntu 16.04 LTS
- Clementine version: 1.3.1

Clementine displays the metadata for the CD correctly when I import it into my library, presumably taking it from the CD. When I attempt to rip the music from the disc, it rips with no metadata. In the dialog box which shows the destination folder, etc it does not display metadata in the table for each track; instead it simply says "Track 1", "Track 2" and so on. Again, track information displays and displays correctly when _imported_, but not when _ripped_. I found [the exact same issue](https://askubuntu.com/questions/865920/why-doesnt-clementine-pick-up-track-information-when-ripping-cd) on an Ask Ubuntu post, however it is still pending an answer.

As reference, RhythmBox is able to rip the songs fine, although it is quite slow. Asunder hangs at different times and seems to rip the audio quickly before hanging. The optical drive seems to be reading a lot of data too, in order to keep up with the transcoding speed. The drive does not go nearly as fast with Clementine or RhythmBox. 

On my laptop running Xubuntu 16.04, Asunder works perfectly fine with ripping the audio from the discs. I ripped the same discs just the other day, and it also kept the metadata. The optical drive I'm using is connected via SATA, and not through USB. The discs are also in immaculate condition, so reading the discs themselves should not be an issue either.

How can I rip the audio files from the disc and preserve the metadata?

paperbagcorner:
Fixes #6045

### DIFF
--- a/src/ripper/ripcddialog.cpp
+++ b/src/ripper/ripcddialog.cpp
@@ -27,7 +27,7 @@
 #include "config.h"
 #include "core/logging.h"
 #include "core/tagreaderclient.h"
-#include "core/utilities.h"
+#include "devices/cddasongloader.h"
 #include "ripper/ripper.h"
 #include "ui_ripcddialog.h"
 #include "transcoder/transcoder.h"
@@ -53,7 +53,8 @@ RipCDDialog::RipCDDialog(QWidget* parent)
     : QDialog(parent),
       ui_(new Ui_RipCDDialog),
       ripper_(new Ripper(this)),
-      working_(false) {
+      working_(false),
+      loader_(new CddaSongLoader){
   // Init
   ui_->setupUi(this);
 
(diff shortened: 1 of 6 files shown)


src/ripper/ripcddialog.cpp
```diff
@@ -53,7 +53,8 @@ RipCDDialog::RipCDDialog(QWidget* parent)
     : QDialog(parent),
       ui_(new Ui_RipCDDialog),
       ripper_(new Ripper(this)),
-      working_(false) {
+      working_(false),
+      loader_(new CddaSongLoader) {
```
hatstand: `loader_` needs to be a `unique_ptr` otherwise this will leak.

paperbagcorner: How about just setting the parent and let Qt handle it, i.e. 
` loader_(new CddaSongLoader(QUrl(), this))`?

hatstand: Either is fine

paperbagcorner: Done

PR #6098, opened 2018-07-05
Device fixes for MacOs

jonaski:
- Fix setting device on newer MacOs where the property value is longlong.
- Some devices (ie one my USB Arcam DAC's) return an empty description, set this to "Unknown device" so it's possible to select it.


jonaski:
The clang-formatter is not working:
Failed to update HEAD: Request failed: {"message":"Reference does not exist","documentation_url":"https://developer.github.com/v3/git/refs/#update-a-reference"}

PR #6109, opened 2018-07-23
Update README.md

JeffChuaZiLe:
Correct grammar errors for README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,8 @@ Please:
  
  * Try the latest developer build (http://builds.clementine-player.org/) to see if the bug is still present (**Attention**, those builds aren't stable so they might not work well and could sometimes break things like user settings). If it works like a charm even though you see an open issue, please comment on it and explain that the issue has been fixed
  * Check if another person has already opened the same issue to avoid duplicates
- * If there already is an open issue you could comment on it to add precisions about the problem or confirm it
- * In case there isn't, you can open a new issue with an explicit title and as much information as possible (OS, Clementine version, how to reproduce the problem...)
+ * If there is already an open issue you could comment on it to add precisions about the problem or confirm on it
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,8 +24,8 @@ Please:
  
  * Try the latest developer build (http://builds.clementine-player.org/) to see if the bug is still present (**Attention**, those builds aren't stable so they might not work well and could sometimes break things like user settings). If it works like a charm even though you see an open issue, please comment on it and explain that the issue has been fixed
  * Check if another person has already opened the same issue to avoid duplicates
- * If there already is an open issue you could comment on it to add precisions about the problem or confirm it
- * In case there isn't, you can open a new issue with an explicit title and as much information as possible (OS, Clementine version, how to reproduce the problem...)
+ * If there is already an open issue you could comment on it to add precisions about the problem or confirm on it
+ * In case there isn't, you can open a new issue with an explicit title with as much information as possible (OS, Clementine version, how to reproduce the problem...)
```
matkoniecz: Are you sure that
> with an explicit title with as much information as possible

is better than

> with an explicit title and as much information as possible

? It changes meaning and advises to put all possible information in the title, it is rarely possible


Also, why "confirm on it" is better than "confirm it"?

Disclaimer: not a native speaker.

brutalcrozt:
update please @JeffChuaZiLe

2br-2b (changes requested):
I like the first change (if there is already), but the other changes seem to make the sentence more awkward.

README.md
marcusmueller: Definitely agreeing that the *title* doesn't have to contain as much information as possible (it really shouldn't), and "with" clearly associates the clause with "title" not with "issue". 

"Confirm on it" would basically mean "signing a document, confirming that what is stated in the document is true". That's not what was intended here.

That leaves us with the "already is" vs "is already" change… not a strong preference for either.

marcusmueller (changes requested):

Issue #6041, opened 2018-04-06
Clementine doesnt work in Gnome Shell - Fedora 27


Joseluismantilla:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [x] I checked the issue tracker for similar issues
- [x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [x] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Fedora 27
- Clementine version: 1.3.1

### Expected behaviour / actual behaviour
Unfortunately clementine does not work in Gnome-shell -- please try it in xorg and wayland--, it does not open its interface and I can not reproduce anything, I have the system updated without errors - it seems that some dependencies or requirements are missing by being added.

### Steps to reproduce the problem (only for bugs)
$ clementine
10:04:57.996 WARN  unknown                          libpng warning: iCCP: known incorrect sRGB profile 
10:04:58.017 WARN  unknown                          QPixmap: It is not safe to use pixmaps outside the GUI thread 
10:04:58.017 WARN  unknown                          QPixmap: It is not safe to use pixmaps outside the GUI thread 
10:04:58.017 WARN  unknown                          QPixmap: It is not safe to use pixmaps outside the GUI thread 
10:04:58.017 WARN  unknown                          QPixmap: It is not safe to use pixmaps outside the GUI thread 
10:04:58.017 WARN  unknown                          QPixmap: It is not safe to use pixmaps outside the GUI thread 
10:04:58.017 WARN  unknown                          QPixmap: It is not safe to use pixmaps outside the GUI thread 
10:04:58.017 WARN  unknown                          QPixmap: It is not safe to use pixmaps outside the GUI thread 
10:04:58.017 WARN  unknown                          QPixmap: It is not safe to use pixmaps outside the GUI thread 
10:04:58.017 WARN  unknown                          QPixmap: It is not safe to use pixmaps outside the GUI thread 
10:04:58.746 WARN  unknown                          QPainter::begin: Paint device returned engine == 0, type: 2 
10:04:58.888 WARN  unknown                          "sni-qt/13956" WARN  10:04:58.888 void StatusNotifierItemFactory::connectToSnw() Invalid interface to SNW_SERVICE  
10:04:58.888 WARN  unknown                          No systemtrayicon available 
[...]

PR #6154, opened 2018-09-30
Only use system tray if it's available on the system

jonaski:
This should fix issue #6041 among others

### DIFF
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -1075,7 +1075,9 @@ MainWindow::~MainWindow() {
 
 void MainWindow::ReloadSettings() {
 #ifndef Q_OS_DARWIN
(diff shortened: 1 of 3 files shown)


ralmond:
I'm still getting what looks to be this problem running Clemintine 1.4rc and 1.4.1-16-gfa35e229f under Pop OS 22.04 (downstream of Ubuntu 22.04) using gnome-shell and Wayland.  

Launching clementine gives me the menu, but not the main window.  Editing the config file so that `showtray=false` as mentioned in #6041 fixes the problem.

ralmond:
BTW, issue is only happening under Wayland, not a problem under Xorg.  I am gettting the dreaded message:
```Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway. ```
regardless of the value of `QT_QPA_PLATFORM`.

PR #6161, opened 2018-10-04
Allow ignoring prefixes when sorting playlist

### LINKED ISSUE #6143: Artist Sort in Playlists Does Not Ignore "The" in Name When alphabetizing
### Before posting

### System information

- Operating System: Manjaro Linux
- Clementine version: 1.3.1-18

Expected: Sorting by Artist in playlists should ignore "The" in front of artist names ("sort name" should not include 'the')
Actual: "The" is not ignored when sorting by artists in playlists
Create Playlist with mixture of artists, including "The" preceding some artist names.  Sort; artists with "The" preceding their name will be alphabetized with T names.

amhndu:
This introduces new configuration options which allow you to ignore
prefixes while sorting the playlist on album, artist and titles.
Prefixes are configurable, default are "a" and "the".

Closes #6143

### DIFF
--- a/src/playlist/playlist.cpp
+++ b/src/playlist/playlist.cpp
@@ -89,12 +89,25 @@ const char* Playlist::kSettingsGroup = "Playlist";
 
 const char* Playlist::kPathType = "path_type";
 const char* Playlist::kWriteMetadata = "write_metadata";
+const char* Playlist::kSortIgnorePrefix = "sort_ignore_prefix";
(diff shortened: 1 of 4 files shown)


amhndu:
Would it be better to preprocess and remove all the prefixes instead of doing it again and again in the predicate ?
I saw no noticeable performance issues on a playlist with 1500 songs with a modest laptop.

PR #6162, opened 2018-10-05
Convert uses of QtAlgorithms to std:: algorithms

amhndu:
Qt5 has deprecated QtAlgorithms and recommends using the standard algorithms.
This patch converts all uses of Qt algorithms to the corresponding std:: function.


hatstand:
We prefer patches to apply to master

amhndu:
Moved to master. Not sure why the build-ci didn't run here.

PR #6165, opened 2018-10-07
Remove spacers in non-LargeSideBar modes for FancyTabWidget

### LINKED ISSUE #6163: New FancyTabWidget has spacers in all modes
### Before posting

- [x] I built the qt5 branch

Before the change from the Nokia widget, the tabs were all fixed and on the large side bar mode, there would be a spacer separating the last two tabs, but not in any other mode.
Now, the spacer is present in all modes.
Now that the tabs are movable by the user, it makes little sense to even have a spacer in the large side bar mode.

amhndu:
To maintain behavior with the older FancyTabWidget, spacers are now
removed and stored when switching to a non-LargeSideBar mode.
And subsequently restored when switching back to LargeSideBar mode.

Fixes #6163


amhndu:
Thanks, I wasn't really aware. I thought qt5 is the main development branch

amhndu:
@jonaski A different PR of mine was merged to qt5 similarly, should I make a new PR cherry-picking that onto master ? (and similarly for this)

jonaski:
I'm not the project owner or maintainer here. But I suggest that yes.
Whats done until now is that we create PRs mainly on master and then we merge master into qt 5.
As long as they want to still keep Qt 4 support and have that in master.

mrnhmath:
I wonder if there's a reason to keep master on Qt 4.

PR #6166, opened 2018-10-07
Port 'Allow ignoring prefixes when sorting playlist' to master

amhndu:
Ports #6161 onto master as suggested by @jonaski


hatstand:
This doesn't compile with qt4

PR #6177, opened 2018-10-19
Add keyboard shortcut to delete files from disk, closes #6164

### LINKED ISSUE #6164: DELETE key should perform 'Delete from Disk'
Right now:
- Backspace key performs 'Remove from playlist'
- Delete key performs 'Remove from playlist'

Suggested behavior:
- Backspace key performs 'Remove from playlist' (no change)
**- Delete key should perform 'Delete from Disk'** (changed!)

Additionally, when the 'Delete from disk' action is started, the prompt that appears has two buttons: **Cancel** and **Yes**. I believe **Cancel** should be changed to **No** and a temporary key binding during this prompt should allow a **Y** or **N** key to be pressed to perform either of these actions. This would make deleting files from disk entirely possible with keyboard actions alone -- not requiring a mouse.

wheresvic:
Uses `ctrl+x`

### DIFF
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -722,6 +722,7 @@ MainWindow::MainWindow(Application* app, SystemTrayIcon* tray_icon, OSD* osd,
   playlist_delete_ = playlist_menu_->addAction(
       IconLoader::Load("edit-delete", IconLoader::Base),
       tr("Delete from disk..."), this, SLOT(PlaylistDelete()));
+  playlist_delete_->setShortcut(QKeySequence("Ctrl+X"));
   playlist_open_in_browser_ = playlist_menu_->addAction(
       IconLoader::Load("document-open-folder", IconLoader::Base),
       tr("Show in file browser..."), this, SLOT(PlaylistOpenInBrowser()));


wheresvic:
@hatstand aha ok fair enough, I will close the PR !

G-9999:
@wheresvic is control+x hard coded or can it be changed? Personally, I'd  love to be able to customize it to a single key, like the DELETE key !

wheresvic:
@G-9999 I had just hardcoded it as you can see from the change `playlist_delete_->setShortcut(QKeySequence("Ctrl+X"));` but you can also simply try using `DELETE ` and see if that works!

G-9999:
@wheresvic love to... but the latest build is
clementine-1.3.1-500-g9eb92ee.dmg                  16-Oct-2018 00:50            39092592
...not sure the frequency of these builds or if they are even automatic?

amhndu:
Can we instead have it as a configurable shortcut but disabled by default ?
If a user explicitly sets a shortcut by themselves, surely they can't blame the player.

G-9999:
Please make this as @amhndu has suggested which is inline with my original request in https://github.com/clementine-player/Clementine/issues/6164 ... being able to bind the delete action to a single key. Having it not set after install also solves the "user stupidity" issue as has been mentioned. I'm personally not happy with Control+X at all 😭

G-9999:
@wheresvic could you perhaps offer some guidance on perhaps re-opening this ticket so that it can track the status of this feature being implemented as a user-configurable key? Myself and @amhndu  would love to have this power and control !!

G-9999:
😢

PR #6181, opened 2018-10-22
Use qt5 version of system-wide libraries

eclipseo:
These changes are necessary for proper Fedora builds with unbundled dependencies. Otherwise it pulls Qt4 deps which are linked into the resulting binary and thus creating conflict with Qt5 libraries.


eclipseo:
Tested and approved in https://copr.fedorainfracloud.org/coprs/eclipseo/clementine-qt5/package/clementine/

PR #6186, opened 2018-11-02
Drop USE_SYSTEM_QXT, libqxt upstream has declared it unmaintained

a17r:
Quote upstream repository:

"LIBQXT IS NO LONGER MAINTAINED
Qxt will likely not work with newer Qt versions due to usage of internal api.
We recommend that you pick out the parts you want instead of using the entire
libqxt."

Fedora and Mageia seem to be the only ones packaging some libqxt-qt5, but no one else.
Gentoo dropped it half a year ago, turns out Debian did very recently:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=906650
See also:
https://repology.org/metapackage/libqxt/versions
https://repology.org/metapackage/libqxt-qt5/versions
With no one there to accept Clementine patches, the option is obsolete.

Long-term, these files may be either moved out of 3rdparty or the code be
ported away from Qxt.


EGD:
I use this patch with [qxtglobalshortcut](https://github.com/hluk/qxtglobalshortcut) library and not see any problem with it.
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 820b13851..2a90ef899 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -352,11 +352,10 @@ endif(NOT QTIOCOMPRESSOR_INCLUDE_DIRS OR
 # used to link to system installed qxt instead.
 option(USE_SYSTEM_QXT "Don't set this option unless your system Qxt library has been compiled with the Clementine patches in 3rdparty" OFF)
 if (USE_SYSTEM_QXT)
-  find_path(QXTCORE_INCLUDE_DIRS qxtglobal.h PATH_SUFFIXES qt5/QxtCore)
-  find_path(QXTGUI_INCLUDE_DIRS qxtglobalshortcut.h PATH_SUFFIXES qt5/QxtWidgets)
-  set(QXT_INCLUDE_DIRS ${QXTCORE_INCLUDE_DIRS} ${QXTGUI_INCLUDE_DIRS})
+  find_path(QXTGLOBALSHORTCUT_INCLUDE_DIRS qxtglobalshortcut.h PATH_SUFFIXES qxtglobalshortcut)
+  set(QXT_INCLUDE_DIRS ${QXTGLOBALSHORTCUT_INCLUDE_DIRS})
   # We only need its header. We don't need to link to QxtCore.
-  find_library(QXT_LIBRARIES QxtWidgets-qt5)
+  find_library(QXT_LIBRARIES qxtglobalshortcut)
 else (USE_SYSTEM_QXT)
   add_definitions(-DQXT_STATIC -DBUILD_QXT_GUI -DBUILD_QXT_CORE)
   set(QXT_INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/qxt)
diff --git a/src/core/qxtglobalshortcutbackend.cpp b/src/core/qxtglobalshortcutbackend.cpp
index c04e41d4a..4bbaa82c8 100644
--- a/src/core/qxtglobalshortcutbackend.cpp
+++ b/src/core/qxtglobalshortcutbackend.cpp
@@ -41,7 +41,7 @@ bool QxtGlobalShortcutBackend::DoRegister() {
 void QxtGlobalShortcutBackend::AddShortcut(QAction* action) {
   if (action->shortcut().isEmpty()) return;
   QxtGlobalShortcut* shortcut = new QxtGlobalShortcut(action->shortcut(), this);
-  connect(shortcut, SIGNAL(activated()), action, SLOT(trigger()));
+  connect(shortcut, &QxtGlobalShortcut::activated, action, &QAction::trigger);
   shortcuts_ << shortcut;
 }
```

a17r:
I'm all for using system libs, but is this project by any means a self-conscious upstream (in part-)successor to libqxt? It does not have a release yet while no commit since Aug 1, and their cmake is a bit cringy.

EGD:
> is this project by any means a self-conscious upstream (in part-)successor to libqxt?

I don't know anything about it

> It does not have a release yet while no commit since Aug 1

it just works

a17r:
Many things 'just work', does not mean they get packaged

jonaski:
This is a really bad idea and will piss off Fedora packagers, who a lot of other distros base their packaging on. The comment referred was put up in 2014, it's not news. At least give them the option to package the library as long as it's used in Clementine.
It would make more sense to move away from qxt and now that using shortcuts through x11 breaks on a lot of distros something needs to be done with that.

a17r:
I don't think so. They would be pissed off if this was still a library in active development, but since it is not, really clementine is new upstream for that. Now if https://github.com/hluk/qxtglobalshortcut is a credible alternative and picking up the phone on bug reports, that would make sense to keep the system option then.

PR #6203, opened 2018-11-16
Display podcast episode information.

jbroadus:
Add an EpisodeInfoWidget with title, author, duration, date, and description fields. Include this in the PodcastInfoDialog. If exactly one episode is selected, then show both the podcast and episode widgets and display the episode's URL. Otherwise, hide the episode widget and follow the existing behavior. Note that the desription field for the EpisodeInfoWidget uses the QLabel and does not currently download embedded images.
Add an always_show_hours option to the PrettyTime methods to include hours in formatting even if the field is zero. This is less ambiguious in some cases where duration is displayed.


jbroadus:
"make format" still didn't work, so I manually applied the patch. I'd like to understand how to properly run against the format checker.

PR #6220, opened 2018-11-29
Convert NULL to nullptr

jonaski:


hatstand:
FYI, this broke the build on Jessie & Trusty: https://buildbot.clementine-player.org/builders/Ubuntu%20Trusty%2032-bit/builds/10/steps/compile/logs/stdio

PR #6242, opened 2018-12-26
Handle MTP connection failures more gracefully.

jbroadus:
Currently, the failure to connect to an MTP device results in the display of an
open device that appears empty.
- Modify ConnectedDevice::Init to return a boolean to indicate success.
- Add an Init to MtpLoader to create the MtpConnection and check the result.
- Call MtpLoader::Init from MtpDevice::Init. This does cause the MTP device to
  be opened on the main thread rather than the MtpDevice's loader thread.
- Handle the result in the device manager.

### DIFF
--- a/src/devices/cddadevice.cpp
+++ b/src/devices/cddadevice.cpp
@@ -40,9 +40,10 @@ CddaDevice::CddaDevice(const QUrl& url, DeviceLister* lister,
 
 CddaDevice::~CddaDevice() {}
 
-void CddaDevice::Init() {
+bool CddaDevice::Init() {
   song_count_ = 0;  // Reset song count, in case it was already set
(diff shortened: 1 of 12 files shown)


hatstand:
Can you change this to not do device connection on the main thread?

jbroadus:
> Can you change this to not do device connection on the main thread?

Yeah, we could have the device report the failure asynchronously. My motivation behind allowing Init to fail is that it would eventually allow the device manager to try a different URL if the first chosen failed. But I think that could be accomplished without blocking the main thread with a bit of rework.

hatstand:
Yeah, given device connection stuff is probably the most likely thing to hang I think we need it to be off the main thread.

jbroadus:
Okay. I'll close this request and look at alternatives.

PR #6246, opened 2018-12-30
Fix Qt 5 build

jonaski:
Commit 453270c broke building.
Because the link line does not include dbus or x11 if found.
Also, visualizations are optional and newer macOS don't have OpenGL so it shouldn't be forced to link.


seanbruno:
This fixes the FreeBSD for Clementine's QT5 branch.

PR #6247, opened 2018-12-31
Fixes for compiling qt5 on ubuntu

camellan:


hatstand:
Some of these fixes are in other PRs, there are some unrelated changes. Additionally, we are deliberately building both qt5 and qt4 on travis.

PR #6259, opened 2019-01-18
[qxt] Compil fix for qt version > 5.6

aliallaoui:
qxt required QPlatformNativeInterface which is not available in qt > 5.6 (I compiled with 5.9 under ubuntu 18.04), so let's use QX11Info instead

### DIFF
--- a/3rdparty/qxt/qxtglobalshortcut_x11.cpp
+++ b/3rdparty/qxt/qxtglobalshortcut_x11.cpp
@@ -29,8 +29,11 @@
 ** <http://libqxt.org>  <foundation@libqxt.org>
 *****************************************************************************/
 
-#if QT_VERSION < QT_VERSION_CHECK(5,0,0)
+#if QT_VERSION < QT_VERSION_CHECK(5,0,0) || QT_VERSION > QT_VERSION_CHECK(5,6,0) 
 #   include <QX11Info>
+#if QT_VERSION > QT_VERSION_CHECK(5,6,0)
+#   include <xcb/xcb.h>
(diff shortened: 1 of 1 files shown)


3rdparty/qxt/qxtglobalshortcut_x11.cpp
```diff
@@ -29,8 +29,11 @@
 ** <http://libqxt.org>  <foundation@libqxt.org>
 *****************************************************************************/
 
-#if QT_VERSION < QT_VERSION_CHECK(5,0,0)
+#if QT_VERSION < QT_VERSION_CHECK(5,0,0) || QT_VERSION > QT_VERSION_CHECK(5,6,0)
```
jonaski: Need need to do this check at all, since Qt 5.6 is required. So just use QX11Info

3rdparty/qxt/qxtglobalshortcut_x11.cpp
```diff
@@ -92,7 +95,7 @@ class QxtX11Data {
 public:
     QxtX11Data()
     {
-#if QT_VERSION < QT_VERSION_CHECK(5,0,0)
+#if QT_VERSION < QT_VERSION_CHECK(5,0,0) || QT_VERSION > QT_VERSION_CHECK(5,6,0)
         m_display = QX11Info::display();
 #else
         QPlatformNativeInterface *native = qApp->platformNativeInterface();
```
jonaski: Remove platformNativeInterface, and use QX11Info, it's available as of 5.1.

PR #6274, opened 2019-01-27
Fix this properly

jonaski:


EGD:
This is a wrong fix.
Please, see
https://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals

EGD:
@jonaski, no, need like this:

```status_text = tr("%n song(s)", "", count);```

and update translations
ie, two from for EN (singular — %n song, plural — %n songs) and three for RU (singular, dual, plural)

PR #6285, opened 2019-02-14
Vorbis comment functors

smithjd15:
Convert the Vorbis comment helper functions into in-line functors.


hatstand:
I don't really understand the advantages of inlining these as lambdas?

smithjd15:
These are similar in purpose and design to the ape tag functors, and are candidates for inlining. Having these also as functors improves the code consistency.

PR #6286, opened 2019-02-15
More bits for APE files (Finish APE file support)

smithjd15:
I *think* this is now all that was needed for APE files.


smithjd15:
Please pull before #6285, that pr fixes the clang-formatter check reported in this pr.

PR #6295, opened 2019-02-22
Avoid making copies when iterate on container

ShFil119:


src/internet/podcasts/gpoddersearchpage.cpp
```diff
@@ -60,7 +60,7 @@ void GPodderSearchPage::SearchFinished(mygpo::PodcastListPtr list) {
 
   model()->clear();
 
-  for (mygpo::PodcastPtr gpo_podcast : list->list()) {
+  for (const mygpo::PodcastPtr& gpo_podcast : list->list()) {
```
hatstand: These pointer types should be treated as values, i.e., we always want to copy them.

PR #6297, opened 2019-02-22
Initialize variables (not covered by ctor)

ShFil119:


src/core/mpris.h
```diff
@@ -39,7 +39,7 @@ class Mpris : public QObject {
   void RaiseMainWindow();
 
  private:
-  Mpris1* mpris1_;
+  Mpris1* mpris1_ = nullptr;
```
jonaski: This variable is unused and can be removed

PR #6311, opened 2019-03-23
Free decoder bin if error occurs during setup.

jbroadus:
In the case that an error occurs in ReplaceDecodeBin before the bin is added to
the pipeline, unreference the object to allow cleanup. This change also separates
CreateDecodeBinFromUrl from ReplaceDecodeBin, following the pattern of
CreateDecodeBinFromString.

### DIFF
--- a/src/engines/gstenginepipeline.cpp
+++ b/src/engines/gstenginepipeline.cpp
@@ -153,17 +153,28 @@ bool GstEnginePipeline::ReplaceDecodeBin(GstElement* new_bin) {
 }
 
 bool GstEnginePipeline::ReplaceDecodeBin(const QUrl& url) {
-  GstElement* new_bin = nullptr;
+  GstElement* new_bin = CreateDecodeBinFromUrl(url);
+  return ReplaceDecodeBin(new_bin);
+}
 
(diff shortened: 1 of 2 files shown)


jbroadus:
Looks like a failed docker pull caused the build failure.

src/engines/gstenginepipeline.cpp
```diff
@@ -153,17 +153,28 @@ bool GstEnginePipeline::ReplaceDecodeBin(GstElement* new_bin) {
...
+  GstElement* new_bin = CreateDecodeBinFromUrl(url);
+  return ReplaceDecodeBin(new_bin);
+}
 
+GstElement* GstEnginePipeline::CreateDecodeBinFromUrl(const QUrl& url) {
+  GstElement* new_bin = nullptr;
 #ifdef HAVE_SPOTIFY
   if (url.scheme() == "spotify") {
     new_bin = gst_bin_new("spotify_bin");
+    if (!new_bin) return NULL;
```
hatstand: Use `nullptr` instead of `NULL` everywhere.

jbroadus: Fixed

PR #6323, opened 2019-04-07
Add error handling path for async song loading.

jbroadus:
Async song loading can fail without user feedback. This change adds return codes
to these async load functions. It will now produce an error dialog in simple
scenarios (test case is user selecting a file that is not readable). Other cases,
such as directories and playlists, aren't yet covered.


jbroadus:
clang-formatter is returning a 503.

hatstand:
I'll fix it, but for now it's just the domain that changed: https://clang-in-the-cloud.clementine-player.org/github/clementine-player/Clementine/6323

PR #6335, opened 2019-04-22
setFirstSectionMovable() in playlistview.cpp

aerusso:
This makes the first column in the playlist viewer draggable.


hatstand:
This function only exists from Qt 5.11. I think you'll need to `#ifdef` it.

aerusso:
@hatstand Sorry for missing that! I've surrounded that in a QT_VERSION check.

Issue #6333, opened 2019-04-19
Complete tags automatically causes crash


alongdecline:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [x] I checked the issue tracker for similar issues
- [x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [x] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Pop!_OS 19.4
- Clementine version: 1.3.1

### Expected behaviour / actual behaviour
The ability to fetch tags is a great aspect of Clementine, but since updated my operating system, Clementine crashes and shuts down immediately when I try to "Complete tags automatically..."

This happens every screen; playlist, library, track editing.

gianlucapettinello:
Same issue for me. Is there a way we can bisect the problem?

gianlucapettinello:
Jonaski patch fixes the issue! Great man

PR #6338, opened 2019-04-26
Fix gst_buffer_unref assertion with chromaprinter

jonaski:
Automatic tag fetching is broken as of at least gstreamer 1.5.90 and newer. With the official released gstreamer 1.6.0 this leaves the feature broken.
There is an incorrect call to gst_buffer_unref that causes an assertion in gstreamer followed by a crash.
My understanding is that gst_buffer_unref should only be done when gst_buffer_ref is called, see:
https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gstreamer/html/GstSample.html#gst-sample-get-buffer
This fix adds a missing gst_sample_unref which is enough to free the buffer too according to the documentation.

Fixes https://github.com/clementine-player/Clementine/issues/6333

### DIFF
--- a/src/musicbrainz/chromaprinter.cpp
+++ b/src/musicbrainz/chromaprinter.cpp
@@ -204,12 +204,16 @@ GstFlowReturn Chromaprinter::NewBufferCallback(GstAppSink* app_sink,
   Chromaprinter* me = reinterpret_cast<Chromaprinter*>(self);
 
   GstSample* sample = gst_app_sink_pull_sample(app_sink);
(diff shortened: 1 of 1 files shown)


gianlucapettinello:
I confirm the patch fixes the issue. You are great!

Turpster:
Cherry-picking this commit to the qt branch and merging...

bxabi:
It's great that this is fixed, but since there is no new release containig the fix, the auto tagging still crashes Clementine for example on Arch Linux.
Could this be released?

PR #6343, opened 2019-05-11
update fr translation

Cormoran96:
Add some translation


hatstand:
Translations are automatically imported via https://www.transifex.com/

PR #6351, opened 2019-05-31
Fix blank window when restoring from systray

### LINKED ISSUE #6111: Clementine does not restore correctly from KDE system tray.
If I open the application and close it to the system tray, if I click to restore it, the application doesn't show anything. No controls. Just a blank for. The application seems to function normally though (I can control Qbittorent from web interface). I also noticed that the issue occurs only when closing the application from the "x". If I bring up from the tray by clicking the icon it opens just fine. Once I close it to tray, using the "x", it wont restore normally.

This is the change that causes the issue:

http://code.qt.io/cgit/qt/qtbase.git/commit/?id=e0b5ff4a

### System information

O.S.: Arch Linux
Version: Qt5-git r565.gd20c2244a

lordlethis:
Fix #6111 by ignoring instead of accepting the main window's closeEvent.

This applies the fix qBittorrent used for this same issue (https://github.com/qbittorrent/qBittorrent/issues/9240) to clementine, so credit goes to those guys.

### DIFF
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -1441,7 +1441,7 @@ void MainWindow::closeEvent(QCloseEvent* event) {
     keep_running = s.value("keeprunning", tray_icon_->IsVisible()).toBool();
 
   if (keep_running && event->spontaneous()) {
-    event->accept();
+	event->ignore();
     SetHiddenInTray(true);
   } else {
     Exit();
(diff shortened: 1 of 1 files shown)


src/ui/mainwindow.cpp
```diff
@@ -1441,7 +1441,7 @@ void MainWindow::closeEvent(QCloseEvent* event) {
     keep_running = s.value("keeprunning", tray_icon_->IsVisible()).toBool();
 
   if (keep_running && event->spontaneous()) {
-    event->accept();
+	event->ignore();
```
a17r: Please fix the whitespace error.

Issue #3711, opened 2013-12-14
'-1' added to ID3 year tag when changing other tag(s)


Clementine-Issue-Importer:
_From [rowanpar...@gmail.com](https://code.google.com/u/105847107567242214105/) on May 31, 2013 23:23:36_

Hello, got a bit of a niggle. Could just be my OCD but I don't think it should happen. I did try to search but searching for '-1' isn't the easiest thing to do!
1. Check the file has a blank year tag (I used EasyTAG).
2. Add it to a playlist.
3. Change the album (possibly others) tag.
4. EasyTAG now gives year tag as '-1'.

I expect that the year tag should be unchanged. It causes VLC to give the year as '1'. Also, could it be possible that it is changing an unset track number to '-1' also?

I am using the nightlies on Kubuntu 12.04.

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=3711_

Issue #3528, opened 2013-12-13
Can't delete track information


Clementine-Issue-Importer:
_From [talscudi@gmail.com](https://code.google.com/u/talscudi@gmail.com/) on February 28, 2013 17:26:50_

What steps will reproduce the problem? 1. Open up 'Edit Track Information'.
2. Delete a field's value.
3. Move to a different field.

ALTERNATE (produces same error)
1. Edit directly in playlist.
2. Delete a field's value.
3. Finish editing OR move to a new field to edit. What is the expected output? What do you see instead? Expected:  Empty field to stay that way.
Actual:  Field is re-filled with whatever was in the field prior to clearing it. What version of the product are you using? On what operating system? Clementine: 1.0.1
OS: Debian Wheezy 3.2.0-4-amd64 Please provide any additional information below. Surely people can't be required to keep all of this information attached to each song?

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=3528_

Lukastos:
I have the same (lubuntu 16.04).

I can edit tags but not delete them totally.

The dirty workaround is to put a space instead of the tag...

Issue #1835, opened 2013-12-09
Tag editor should not write other fields than the edited ones


Clementine-Issue-Importer:
_From [goetzchrist](https://code.google.com/u/101757420644949299224/) on May 01, 2011 01:34:26_

What steps will reproduce the problem? 1. Edit with Clementine any tag of a file.
2. Open the file with some other tag editor, e.g. kid3.
3. It is possible to see that there are other new fields with some values. What is the expected output? What do you see instead? Clementine should not write in any field, only in the edited fields by the user. What version of the product are you using? On what operating system? r3170 Kubuntu 10.10 Please provide any additional information below. The new fields with their respective values I see are (those are ID3v2.4):
Date = -1
Track Number = -1
Unknown = 0   (for what is this?)

Also, if I have a file with only tags in ID3v2.x, Clementine should not write the edited tags to ID3v1.x too (or this could be configurable).

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=1835_

PR #6354, opened 2019-06-04
Fix a number of zero-value tag issues (New PR)

smithjd15:
Some fixes to prevent the writing of zero values and also addressing the inability to clear some tag fields in the tag editor.

- Fix compilation and disc field zero-value checks
- Check year, track and statistics fields for zero-values

Fixes (or improves): #1835 #3711 #3528

### DIFF
--- a/ext/libclementine-tagreader/tagreader.cpp
+++ b/ext/libclementine-tagreader/tagreader.cpp
@@ -38,7 +38,6 @@
 #include <mp4file.h>
 #include <mp4tag.h>
 #include <mpcfile.h>
-#include <apefile.h>
 #include <mpegfile.h>
 #include <oggfile.h>
 #ifdef TAGLIB_HAS_OPUS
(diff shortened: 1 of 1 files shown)


ext/libclementine-tagreader/tagreader.cpp
```diff
@@ -783,20 +787,20 @@ bool TagReader::SaveFile(const QString& filename,
   fileref->tag()->setAlbum(StdStringToTaglibString(song.album()));
   fileref->tag()->setGenre(StdStringToTaglibString(song.genre()));
   fileref->tag()->setComment(StdStringToTaglibString(song.comment()));
-  fileref->tag()->setYear(song.year());
-  fileref->tag()->setTrack(song.track());
+  fileref->tag()->setYear(song.year() <= 0 - 1 ? 0 : song.year());
```
hatstand: I'm not sure what the point of the `0 - 1` is?

smithjd15: Matching other instances of the representation of '-1'.

hatstand: Fair enough.

PR #6371, opened 2019-07-17
Introduce limit for number of tag client processes.

areading:
Previously, the number of processes spawned was always
QThread::idealThreadCount() (returning the number of logical CPU
cores). On new systems with many cores, however, this can result
in 12, 16, 24, or ... processes being spawned, which is a bit
excessive.

This establishes a new config variable,
'max_numprocs_tagclients' within the Settings group, in order
to limit the maximum number of tag client processes that get
spawned. It also adds a means of setting this via the Behavior
page in Settings. It can be set to any integer in the interval
[1, QThread::idealThreadCount()]; it defaults to the maximal value
so as to emulate the old behavior.

### DIFF
--- a/src/core/tagreaderclient.cpp
+++ b/src/core/tagreaderclient.cpp
@@ -18,6 +18,7 @@
    along with Clementine.  If not, see <http://www.gnu.org/licenses/>.
 */
 
+#include "player.h"
 #include "tagreaderclient.h"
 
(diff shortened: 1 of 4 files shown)


areading:
Thanks for the feedback. I modified the commit as appropriate and backed out the changes to the header file; ready for re-review.

Issue #5777, opened 2017-07-04
Clementine crashes many times a day


Ra0EL:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [x] I checked the issue tracker for similar issues
- [x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [x] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System:   OpenSuSE Tumbleweed VERSION_ID="20170617"
- Clementine version:    1.3.1-5.3 x64

### Description
Hi,

I am facing the same problem for years already with Clementine. Today I am fed up with it and have decided to post something.
I have read a lot in many forum about the same issue, and different guidances in order to solve the problem.
But nothing has proved to be of some help.

In few words, problems are :
 - when I edit a file, the save process is never ending. If I try to edit another song, nothing will come. Or the edit window, but most of fields are grey. Or Clementine crashes.
 - when I start Clementine, update scan seems to run and end as expected. But if I start another scan, it will be stuck at 1%
 - when playing dynamic "Smart playlist", Clementine is crashing now and then - see capture below :

This an abstract of message in --verbose mode :
```
16:40:28.642 DEBUG GstEnginePipeline:519            2 sync bus message state-changed 
16:40:28.642 DEBUG GstEnginePipeline:519            2 sync bus message state-changed 
16:40:28.642 DEBUG GstEnginePipeline:519            2 sync bus message async-done 
16:40:28.642 DEBUG GstEnginePipeline:902            2 event latency 
16:40:28.643 DEBUG GstEnginePipeline:902            2 event latency 
16:40:28.643 DEBUG GstEnginePipeline:519            2 sync bus message state-changed 
16:40:28.643 DEBUG GstEnginePipeline:519            2 sync bus message state-changed 
16:40:28.643 DEBUG GstEnginePipeline:519            2 sync bus message state-changed 
16:40:28.643 DEBUG GstEnginePipeline:519            2 sync bus message state-changed 
16:40:28.643 DEBUG GstEnginePipeline:519            2 sync bus message state-changed 
16:40:28.643 DEBUG GstEnginePipeline:519            2 sync bus message state-changed 
16:40:28.643 DEBUG GstEnginePipeline:519            2 sync bus message state-changed 
[...]

Issue #6317, opened 2019-03-29
Clementine-qt5-git crashes very regularly


trougnouf:
[gdb.txt](https://github.com/clementine-player/Clementine/files/3024056/gdb.txt)

Using clementine-qt5-git on Arch Linux, it crashes all the time which decreases my quality of life significantly because I rely on Clementine's scoring and ratings-to-file features and life sparse music isn't as good.

I installed the qt5 debug symbols and attached a log.

I made a bug report in the distant past and it seemed to have gotten better for a while, perhaps with some patches from strawberry or some other reason, so I closed it, but it's now worse than ever.

TimonAlex:
Same thing on FreeBSD with qt5. Player crashes after a few minutes of playing files. I build clementine with -fsanitize=address and got this:

```21:01:07.749 DEBUG MessageReply<pb::tagreader::Message>:91 Messages: request 0x 0x6120008e0b60 , reply 0x 0x6120008e0bf0
=================================================================                                                                                                                             
==47241==ERROR: AddressSanitizer: heap-use-after-free on address 0x6120008e0c78 at pc 0x000001753aaf bp 0x7fffdeff3cc0 sp 0x7fffdeff3cb8                                                      
READ of size 4 at 0x6120008e0c78 thread T9                                                                                                                                                    
    #0 0x1753aae in pb::tagreader::Message::id(void) const /usr/ports/audio/clementine-player/work/.build/ext/libclementine-tagreader/tagreadermessages.pb.h:5504:10                          
    #1 0x1753f28 in MessageReply<pb::tagreader::Message>::SetReply(pb::tagreader::Message const&) /usr/ports/audio/clementine-player/work/Clementine-1.3.1-661-g4ff370ce1/ext/libclementine-co
mmon/core/messagereply.h:92:57
<...>
```
MessageReply object deleted in another thread:
```
0x6120008e0c78 is located 312 bytes inside of 320-byte region [0x6120008e0b40,0x6120008e0c80)
freed by thread T0 here:
    #0 0x16051f2 in operator delete(void*) /usr/src/contrib/compiler-rt/lib/asan/asan_new_delete.cc:149:3                                                                                    
    #1 0x175a841 in MessageReply<pb::tagreader::Message>::~MessageReply(void) /usr/ports/audio/clementine-player/work/Clementine-1.3.1-661-g4ff370ce1/ext/libclementine-common/core/messagerep
ly.h:59:7
[...]

PR #6372, opened 2019-07-20
Fix crash in messagereply

jonaski:
The messagereply gets deleted too early, I tried different approaches to solving it including using invokeMethod with Qt::QueuedConnection on deleteLater() which had no affect, but found that simply moving the signal works.
It seem like the messagereply gets deleted in the main thread before the semaphore is released.
I've set up a small test program that does tags on files in a directory in a loop. Before this it always crashed, typically around something like 300 tagwrites. Now I'm testing 10.000 writes in a loop without any crash.

Fixes https://github.com/clementine-player/Clementine/issues/5777 and https://github.com/clementine-player/Clementine/issues/6317

### DIFF
--- a/ext/libclementine-common/core/messagereply.h
+++ b/ext/libclementine-common/core/messagereply.h
@@ -89,7 +89,6 @@ void MessageReply<MessageType>::SetReply(const MessageType& message) {
   qLog(Debug) << "Releasing ID" << id() << "(finished)";
   semaphore_.release();
   emit Finished(success_);
-
 }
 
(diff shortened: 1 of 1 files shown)


plater:
I've added the patch to Tumbleweed's qt5 branch snapshot and to Leaps 1.3.1 builds. I can reproduce and test the bug against Leap:15.0 and 1 with clementine-1.3.1 but I rely on the reporter to test the Tumbleweed bug because I can't reproduce it.

PR #6375, opened 2019-07-25
Fix memory leaks in database

jonaski:
Credit goes to Jiri Slaby (SUSE) for pointing out the leak in Database::BackupFile()

### DIFF
--- a/src/core/database.cpp
+++ b/src/core/database.cpp
@@ -262,7 +262,7 @@ QSqlDatabase Database::Connect() {
   }
 
   // Find Sqlite3 functions in the Qt plugin.
-  StaticInit();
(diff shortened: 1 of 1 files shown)


jonaski:
Only nullptr is passed to BOOST_SCOPE_EXIT because both OpenDatabase's is done after BOOST_SCOPE_EXIT, memory is never free'ed, so we pass by reference instead, which resolves the leak. There is a leak on each backup otherwise.

sFTSTokenizer is allocated on each thread because each thread has a unique db connection, even in concurrent runs, it is never freed. Whats the purpose when the reference is passed to bindValue, won't that always point to the latest one anyway?

PR #6376, opened 2019-07-25
Fix memory leaks

jonaski:


ext/libclementine-tagreader/tagreader.cpp
```diff
@@ -125,6 +125,11 @@ TagReader::TagReader()
       network_(new QNetworkAccessManager),
       kEmbeddedCover("(embedded)") {}
 
+TagReader::~TagReader() {
+  delete network_;
```
hatstand: Can we use `unique_ptr` here instead?

ext/libclementine-tagreader/tagreader.h
```diff
@@ -120,7 +120,6 @@ class TagReader {
       TagLib::ID3v2::Tag* tag);
 
   FileRefFactory* factory_;
```
hatstand: Can't this be a `unique_ptr` too?

Issue #6209, opened 2018-11-22
[regression] Sidebar tabs position is no longer persisted


ronjouch:
### Before posting

- [x] I checked the issue tracker for similar issues
- [x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [x] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

- Operating System: Ubuntu Linux 18.04.1 LTS x64
- Clementine version: `master` as of November 22: 1.3.1-517-g 3bc6f7059, built with default options (`cd bin && cmake .. && make -j -l4`).

### Steps to reproduce & expected/actual behaviour

1. Right-click on sidebar tabs (Library, Song info, Search, Files, ...) and pick a non-default position, e.g. `Tabs on top`.
2. Quit Clementine.
3. Start Clementine.

**Expected**: Clementine should honor my setting and start with `Tabs on top`:

![screenshot_clementine-tabsontop_2018-11-22-10 20 05](https://user-images.githubusercontent.com/522085/48911252-55d2dc00-ee40-11e8-8686-6a1b960256d2.png)

**Actual**: Clementine starts with `Plain sidebar`, forgetting my previously-set preference:

![screenshot_clementine-plainsidebar_2018-11-22-10 20 36](https://user-images.githubusercontent.com/522085/48911273-63886180-ee40-11e8-910d-e81039c2b809.png)

Problem is a regression that I'd say happened somewhere in November 2018.

jonaski:
I'm aware of this. It's the same with the window positions and the columns, I'll submit a pr for it as soon as I get the time, probably next week.

ronjouch:
@jonaski I confirm https://github.com/clementine-player/Clementine/pull/6217 / 289ea31f8cf43e11921e525e87323158f831796c fixes the issue. Closing. Thanks!

PR #6398, opened 2019-09-07
Periodically save settings

aerusso:
This patchset consists of two patches, representing the development timeline. The first patch introduces an efficient framework to periodically save modified settings, rather than doing so immediately. This is the convention for events that may happen many times per second (such as flipping through tabs or resizing a window) in, e.g., KDE (see `kxmlgui`'s `kmainwindow.cpp`).  Instead, the setting is marked as dirty, and a timer is set (to 1 second), and saved then. If the program is closed before then, the dirty data is also saved via an `aboutToQuit` signal connection.

By coordinating the saving, the number of write locks that are required can be reduced to a maximum of one per second, tremendously improving user experience on systems with slow locks, reducing power consumption, and significantly reducing wear on devices (e.g., SSD drives).

The first patch implements this for geometry changes (both for `MainWindow` and  `PlaylistView`) and for `PlaylistView` stretched settings. It also cleans up some inconsistent uses of `QSettings`, possibly further reducing the need for read locks.

Compare to #6057 (6a312e7), which only ever intended to save on program exit (and may have failed to do so for those using Nvidia drivers), causing problems for some people (see #6217 and #6209, these are probably nvidia users whose shutdown sequence was abruptly terminated by 5237c005c940, inhibiting the intended save logic). At this point the workaround for that ancient Nvidia bug has been reverted (see 783dada13, #6334), so there is no reason to expect users of "recent" (release in the last ~7 years) drivers to experience any shutdown issues. Furthermore, this would only ever result in at most 1 second of lost settings.

The second commit uses the framework to also save the currently selected tab, which at the time is only saved at program exit. Along the way, it can happily simplify the logic of applying `SetApplication`.

### DIFF
--- a/src/core/application.cpp
+++ b/src/core/application.cpp
@@ -20,6 +20,9 @@
    along with Clementine.  If not, see <http://www.gnu.org/licenses/>.
 */
 
+#include <QSettings>
+#include <QTimer>
+
 #include "application.h"
 
 #include "config.h"
@@ -67,7 +70,8 @@ bool Application::kIsPortable = false;
 class ApplicationImpl {
  public:
   ApplicationImpl(Application* app)
-      : tag_reader_client_([=]() {
+      : settings_timer_(app),
+        tag_reader_client_([=]() {
           TagReaderClient* client = new TagReaderClient(app);
           app->MoveToNewThread(client);
           client->Start();
@@ -150,6 +154,9 @@ class ApplicationImpl {
         }) {
   }
 
+  QTimer settings_timer_;
+  QSettings settings_;
+
   Lazy<TagReaderClient> tag_reader_client_;
   Lazy<Database> database_;
   Lazy<AlbumCoverLoader> album_cover_loader_;
@@ -188,6 +195,10 @@ Application::Application(QObject* parent)
 
   // TODO(John Maguire): Make this not a weird singleton.
   tag_reader_client();
+
+  p_->settings_timer_.setInterval(1000);
+  p_->settings_timer_.setSingleShot(true);
+  connect(&(p_->settings_timer_), SIGNAL(timeout()), SLOT(SaveSettings_()));
 }
 
 Application::~Application() {
@@ -228,6 +239,8 @@ QString Application::language_without_region() const {
   }
   return language_name_;
 }
+
+void Application::SaveSettings_() { emit SaveSettings(&(p_->settings_)); }
 
 void Application::ReloadSettings() { emit SettingsChanged(); }
 
@@ -326,3 +339,5 @@ TagReaderClient* Application::tag_reader_client() const {
 TaskManager* Application::task_manager() const {
   return p_->task_manager_.get();
 }
+
+void Application::DirtySettings() { p_->settings_timer_.start(); }
--- a/src/core/application.h
+++ b/src/core/application.h
@@ -28,6 +28,8 @@
 
 #include "ui/settingsdialog.h"
 
+class QSettings;
+
 class AlbumCoverLoader;
 class Appearance;
 class ApplicationImpl;
@@ -98,6 +100,8 @@ class Application : public QObject {
   TagReaderClient* tag_reader_client() const;
   TaskManager* task_manager() const;
 
+  void DirtySettings();
+
   void MoveToNewThread(QObject* object);
   void MoveToThread(QObject* object, QThread* thread);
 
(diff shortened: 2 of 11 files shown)


src/core/application.cpp
```diff
@@ -177,7 +180,9 @@ class ApplicationImpl {
 };
 
 Application::Application(QObject* parent)
-    : QObject(parent), p_(new ApplicationImpl(this)) {
+    : QObject(parent),
+      p_(new ApplicationImpl(this)),
+      settings_timer_(new QTimer(this)) {
```
aerusso: Is the motivation for putting most of the implementation details into the `p_` object to reduce the complexity of the (widely included) `src/core/application.h` header? I did not understand the convention, and the `QSettings` class needs to be declared at this global level so that it can be used as an argument.

Should I hide `settings_timer_` inside `ApplicationImpl`?

src/widgets/fancytabwidget.h
```diff
@@ -43,8 +44,8 @@ class FancyTabWidget : public QTabWidget {
         void setBackgroundPixmap(const QPixmap& pixmap);
         void addSpacer();
 
-        void loadSettings(const char *);
-        void saveSettings(const char *);
+        void loadSettings(QSettings&);
```
hatstand: Do not use non-const references, you probably just want to pass-by-value here as with most `QObjects`

src/core/application.h
```diff
@@ -109,12 +113,18 @@ class Application : public QObject {
...
   void SettingsDialogRequested(SettingsDialog::Page page);
 
+ private slots:
+  void SaveSettings_();
+
  private:
   QString language_name_;
   std::unique_ptr<ApplicationImpl> p_;
   QList<QThread*> threads_;
+  QTimer* settings_timer_;
```
hatstand: This can just be on the stack, i.e., `QTimer settings_timer_;`

src/core/application.cpp
hatstand: It's the "pimpl" idiom and yes, ideally, `settings_timer_` should be inside `ApplicationImpl` like everything else.

src/widgets/fancytabwidget.cpp
```diff
@@ -290,33 +290,27 @@ FancyTabWidget::FancyTabWidget(QWidget* parent) : QTabWidget(parent),
     connect(tabBar, SIGNAL(currentChanged(int)), this, SLOT(currentTabChanged(int)));
 }
 
-void FancyTabWidget::loadSettings(const char *kSettingsGroup) {
-    QSettings settings;
-    settings.beginGroup(kSettingsGroup);
+void FancyTabWidget::loadSettings(QSettings& settings) {
+  for (int i = 0; i < count(); i++) {
+    int originalIndex = tabBar()->tabData(i).toInt();
+    std::string k = "tab_index_" + std::to_string(originalIndex);
```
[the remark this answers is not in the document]
aerusso: Yep! Since this is mostly unrelated to the original patch, I'll factor it out into another commit, but include it in this MR.

src/widgets/fancytabwidget.h
aerusso: Just to be clear, you're talking about only this change (to `loadSettings`, not about, e.g., `saveSettings` right beneath it)?

hatstand: Both

aerusso: Sorry, then I'm confused. `saveSettings` changes the `QSettings` object (in set). It can't be passed const. What am I missing?

hatstand: Ah, I see what you're doing here now. It should be:
```c++
void loadSettings(const QSettings&);
void saveSettings(QSettings*);
```

aerusso: Gotcha! I'll probably be able to get these resolved tonight. Thanks for your comments!

aerusso:
I've split out the changes to `FancyTabWidget::loadSettings`, and the changes to the string handling into two other commits.  Also, this version hides most of the changes to Application inside ApplicationImpl.

I also structured the changes to the original two patches in a (somewhat) [sane way in this branch](https://github.com/aerusso/Clementine/tree/support/settings-autosave-timer).  I actually think it winds up being harder to parse that branch, because it includes some refactoring and whitespace changes that really muck things up.

aerusso:
@hatstand Is there anything else I can do to make this patch easier to merge?

PR #6429, opened 2019-11-01
Remove trailing slash for LastFM auth URLs to avoid a redirect.

uqs:
Also add better error logging while here. This will at least spit out
SSLHandshake errors and avoids parsing an empty reply if the request
fails.

This is fallout from issue #5993


uqs:
This is now ready to be merged, anyone?

src/internet/lastfm/lastfmservice.cpp
```diff
@@ -174,20 +175,28 @@ void LastFMService::Authenticate() {
     NewClosure(reply, SIGNAL(finished()), this, SLOT(AuthenticateReplyFinished(QNetworkReply*)), reply);
   });
 
+  qLog(Debug) << "auth URL:" << url.toString();
```
hatstand: Can you remove the logging for non-errors here and below?

uqs: I found these very helpful when trying to debug my authentication problem. Note that In case openUrl fails the URL is also being logged. This is also happening just once during the auth phase, so it's not adding undue log volume or anything.

As I said, it was helpful to see what my browser was returning for these URLs and they might help someone to not have to read the code.

Could you expand why you think they should be removed?

hatstand: We should avoid logging potentially sensitive data, e.g., tokens, oauth codes, etc. as people often post full logs here.

It's easy enough to add this back in later when developing.

PR #6437, opened 2019-11-09
Replace QString::null with QString()

jonaski:


JulianVolodia:
Hi, why clang-formatter made X instead of V on 0da490a commit?

Details are not accessible there. https://clang-in-the-cloud.clementine-player.org/github/clementine-player/Clementine/6437
SSO with Github not longer working.

PR #6451, opened 2019-11-13
Fix setting mtime/btime in tagreader

### LINKED ISSUE #6423: Wrong handling of QT's created() time broke FreeBSD / NFS
Hi there, this is my investigation of issue #5187 and issue #4454. The fix is also straightforward, except that I think that "ctime" should never have been introduced in Clementine, as it's not what people think on POSIX filesystems and more importantly the actual creation time (birthtime) of a file is totally irrelevant to the library handling, IMHO. Not sure what people use it for.

Anyway, QTs "created() function is badly named and implemented. So badly named, that as of 5.10 it has been deprecated and split into proper ctime and btime methods, but they are called in a straightforward way:

/usr/local/include/qt5/QtCore/qfileinfo.h:137:5: note: 'created' has been explicitly marked deprecated here
    QT_DEPRECATED_X("Use either birthTime() or metadataChangeTime()")

Historically, Clementine worked fine on FreeBSD with my files sitting on an NFSv3 mount (also FreeBSD). I haven't used Clementine in probably the last 5 years or so, so I'm assuming that QT5 broke this, as now "created" returns birthtime and no longer ctime. Either that, or the birthtime+NFS support in FreeBSD is just half-implemented.

Long story short: stat(2) on a NFSv3 mount returns -1 for birthtime, whereas on the server locally it returns an actual birthtime.

The fix thus is simple in that on seeing a -1 being returned, we should swap in the real ctime or mtime instead.

ctime doesn't make any sense to be honest. I don't think a change in permissions or ownership is of importance to the library maintenance.

I'll send a pull request for this soon. First need to check whether depending on QT5.10 is ok, or whether older QT5 versions must also still be supported.

Oh, the tagreader also uses toTime_t(), which is no longer documented as of QT5.8. As of Qt5.8 there is toSecsSinceEpoch, but that returns a qint64, not a uint(32)

Some info on the lowest supported QT version would help.
Cheers

jonaski:
Because the version is in hex, 0x051000 is not the same as Qt 5.10. So this code will never be used.
Even if it was, it won't compile because it's called birthTime() not birthtime().

Fixes #6423

### DIFF
--- a/ext/libclementine-tagreader/tagreader.cpp
+++ b/ext/libclementine-tagreader/tagreader.cpp
@@ -128,7 +128,10 @@ void TagReader::ReadFile(const QString& filename,
 
 #if QT_VERSION >= QT_VERSION_CHECK(5, 10, 0)
   qint64 mtime = info.lastModified().toSecsSinceEpoch();
-  qint64 btime = info.birthTime().toSecsSinceEpoch();
+  qint64 btime = mtime;
+  if (info.birthTime().isValid()) {
+    btime = info.birthTime().toSecsSinceEpoch();
+  }
 #elif QT_VERSION >= QT_VERSION_CHECK(5, 8, 0)
   qint64 mtime = info.lastModified().toSecsSinceEpoch();
   qint64 btime = info.created().toSecsSinceEpoch();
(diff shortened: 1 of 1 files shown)


uqs:
Thanks! I'm confused though, see https://doc.qt.io/qt-5/qtglobal.html#QT_VERSION which makes me believe the macros where fine.

I also tested and ran this code, so I'm no longer sure *what* the hell I was testing. Macros are evil, eh?

uqs:
Erm, 0x10 vs 0x0a, nevermind. Got it now.

PR #6462, opened 2019-11-16
Fix and update snap

jonaski:
- Upgrade to core18
- Update dependencies to Qt 5
- Add proper alsa support
- Fix launch icon
- Fix missing plugs ie.: network-manager-observe
- Add plugs for clementine-tagreader
- Remove useless comments


hatstand:
Can we add a circleci job for this?

jonaski:
Maybe. Add builder with ubuntu 18.04 and just run snapcraft in the source directory should be sufficient.

Issue #6363, opened 2019-06-29
Resume Playback on Start Doesn't Work In Ubuntu 18.04


linux4me:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [X] I checked the issue tracker for similar issues
- [X] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [X] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Ubuntu 18.04.2
- Clementine version: 1.3.1

### Expected behaviour / actual behaviour
Expected behavior: If I check Tools -> Preferences -> Behavior -> Resume playback on start, playback should resume when Clementine is started when launched manually or as a Startup Application with the command line, "clementine."

Although it worked once, it no longer does, and I did a complete uninstall and reinstall and still can't get resume playback on start to work. 

### Steps to reproduce the problem (only for bugs)

1. Do a fresh install of Clementine on Ubuntu 18.04.2 after deleting or renaming the folder .config/Clementine.
2. In Tools -> Music Library, set the path to your music library.
3. Check Tools -> Preferences -> Behavior -> Resume playback on start.
4. Load a playlist and begin playing a track.
5. Quit Clementine.
6. Restart Clementine. 

The desired behavior is for Clementine to resume playback. For me, it does not.

linux4me:
It's not a fix, but I did find a workaround. I've been starting Clementine via Startup Applications with the command line, "clementine -v 100," and I also tried "clementine -v 100 -p," which didn't resume playback on start. With "clementine -v 100 -l /home/username/Playlists/My-Playlist.xspf," it would resume playback on start, but would always start playing the first song in the playlist regardless of where it was when it stopped. Once it finished the first track, it would begin shuffling as it should.

The workaround was to record a three-second silence in an Ogg using Audacity, and save it in a folder and file that would always be listed first in the playlist. I used EasyTag to tag the song so it would show up in Clementine. In my case, I just named the artist "001 Placeholder." Now, when Startup Applications launches Clementine, it loads my playlist and plays that three seconds of silence, then begins shuffling songs the way I want it to.

gadfsgagfaq435:
[...]

PR #6475, opened 2019-12-14
Fix resuming playback position on startup

rohtie:
Resuming the playback position on startup has not worked for a while. See issues like #6363 

The reason it did not work is because the song to be resumed does not have a length yet when the seek is requested. My solution is to add a slot that waits until the song has a length so that the seek will work.


Sadi58:
@rohtie I doubt that your fix will work for me. I didn't have this issue until recently (not after any OS or Clementine upgrade) and playback doesn't resume even when the song has already played for more than two minutes (and still is not halfway). This is a very odd behavior. :-(

PR #6494, opened 2020-01-03
Gpodder server

jbroadus:
These changes add an option to specify an alternate or development gpodder server.


jbroadus:
I'll close this and do two separate requests.

PR #6495, opened 2020-01-03
Update to libmygpo-qt 1.1.0.

jbroadus:
The updates include:
- The ability to specify a custom gpodder service.
- Use https instead of http for gpodder.net.
- A fix that prevents encoded URLs from being re-encoded.


jbroadus:
Should formatting be changed for third party libraries?

PR #6525, opened 2020-01-11
Use qt5 version of liblastfm dev package for Debian and Ubuntu builds.

jbroadus:


jbroadus:
Doesn't look like that's working.

jbroadus:
Okay, looks better now.

PR #6530, opened 2020-01-14
Fix LibraryModel async query crash.

jbroadus:
A LibraryBackend may be deleted while an associated LibraryModel object is using
it. An example is an async query running while a connected device is removed.
To prevent this, use a share pointer for the LibraryBackend.

This fixes one case where LibraryBackend is used after deletion. However, the
raw pointer is still passed around in several other places. These should be
evaluated on a case-by-case basis to insure that circular depencencies aren't
introduced.

### DIFF
--- a/src/devices/connecteddevice.cpp
+++ b/src/devices/connecteddevice.cpp
@@ -39,16 +39,15 @@ ConnectedDevice::ConnectedDevice(const QUrl& url, DeviceLister* lister,
       unique_id_(unique_id),
       database_id_(database_id),
       manager_(manager),
-      backend_(nullptr),
       model_(nullptr),
       song_count_(0) {
   qLog(Info) << "connected" << url << unique_id << first_time;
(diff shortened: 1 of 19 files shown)


jbroadus:
This leaves an aesthetically unpleasing mix of shared_ptr and QSharedPointers in the device classes. I can look at converting the shared_ptr usage there to QSharedPointers as well unless there's a preference for the former. I assume that this was implemented before qt had sharedFromThis.

hatstand:
I'd have a slight preference to using `std` over Qt where both exist, especially where we're already using `std` with `shared_ptr` for example.

jbroadus:
Yeah, grepping through the code, I see that QSharedPointer is mainly used with third party libraries. In the future, I imagine that things like connect() might directly support QSharedPointer, but it wouldn't be a heavy lift to switch back at that point.

PR #6533, opened 2020-01-17
Inotify errors

jbroadus:


src/devices/filesystemdevice.cpp
```diff
@@ -77,3 +78,7 @@ FilesystemDevice::~FilesystemDevice() {
   watcher_thread_->exit();
   watcher_thread_->wait();
 }
+
+void FilesystemDevice::WatcherError(const QString& error) {
```
hatstand: Can these just be lambdas in the `connect()` call?

jbroadus: > Can these just be lambdas in the `connect()` call?

Yep

jbroadus: Actually, can be done without a lambda.

PR #6534, opened 2020-01-18
Support compiling with Boost 1.69

Chaz6:
This change allows compilation on RHEL/CentOS 7.


Chaz6:
Sorry, made a mistake on the patch.

PR #6535, opened 2020-01-18
Support compiling with Boost 1.69

Chaz6:
This allows compilation on RHEL/CentOS 7 on which the latest available Boost version is 1.69.


hatstand:
Can you add a CentOS builder to the CircleCI config?

Chaz6:
I can try!

hatstand:
That's what I like to hear :-)

PR #6536, opened 2020-01-19
Fix hung progress indicator on device disconnect.

jbroadus:
When a device is removed during a scan, the scan is cancelled, but the task is
not set to finished.


jbroadus:
I also noticed that there is a TaskManager::ScopedTask that doesn't currently have any users. That could be a good fit for this case.

jbroadus:
Looks like build_buster_64 might have failed due to a build system error.

PR #6540, opened 2020-01-22
Pass a Directory reference to ScanTransaction instead of a directory id.

jbroadus:
This is can be done without protecting the directory reference since the method
that removes directories from the watch list is only called on the same thread
as the scan, and never during the life of the ScanTransaction object.


jbroadus:
The main reason for this change is to provide flexibility for and groundwork for a couple future changes.
- In the near future, allowing a scan of a specific directory to be cancelled when the directory is removed.
- Further in the future, it might be possible to allow a device to store relative subdirectories and get rid of the ConnectedDevice::InitBackendDirectory hack.

PR #6543, opened 2020-01-22
Add ability to cancel scan per directory.

jbroadus:
Store a boolean along with watched directories to indicate active. Use this flag
to halt scans on a per-directory basis.

### DIFF
--- a/src/library/librarywatcher.cpp
+++ b/src/library/librarywatcher.cpp
@@ -27,13 +27,13 @@
 
 #include <QDateTime>
 #include <QDirIterator>
-#include <QtDebug>
-#include <QThread>
-#include <QDateTime>
 #include <QHash>
+#include <QMutexLocker>
 #include <QSet>
 #include <QSettings>
+#include <QThread>
 #include <QTimer>
+#include <QtDebug>
 
 #include <fileref.h>
 #include <tag.h>
@@ -59,7 +59,6 @@ LibraryWatcher::LibraryWatcher(QObject* parent)
       backend_(nullptr),
       task_manager_(nullptr),
       fs_watcher_(FileSystemWatcherInterface::Create(this)),
-      stop_requested_(false),
       scan_on_startup_(true),
       monitor_(true),
       rescan_timer_(new QTimer(this)),
@@ -85,10 +84,9 @@ LibraryWatcher::LibraryWatcher(QObject* parent)
 // is only created on a stack and the removal of a directory from the watch
 // list only occurs as a result of a signal and happens on the  watcher's
 // thread. So the Directory object will not be deleted out from under us.
-LibraryWatcher::ScanTransaction::ScanTransaction(LibraryWatcher* watcher,
-                                                 const Directory& dir,
-                                                 bool incremental,
-                                                 bool ignores_mtime)
+LibraryWatcher::ScanTransaction::ScanTransaction(
+    LibraryWatcher* watcher, const LibraryWatcher::WatchedDir& dir,
+    bool incremental, bool ignores_mtime)
     : progress_(0),
       progress_max_(0),
       dir_(dir),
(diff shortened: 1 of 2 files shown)


src/library/librarywatcher.h
```diff
@@ -200,11 +217,27 @@ class LibraryWatcher : public QObject {
...
-  bool stop_requested_;
   bool scan_on_startup_;
   bool monitor_;
 
-  QMap<int, Directory> watched_dirs_;
+  typedef QMap<int, WatchedDir> WatchListBase;
+  // All methods of QMap are reentrant We should only need to worry about
+  // syncronizing methods that remove directories on the watcher thread and
+  // methods that modify directories called from other threads.
+  class WatchList : public WatchListBase {
```
hatstand: Not really a fan of inheriting from collection classes either. This is much more readable as an object containing a map and a mutex. The typedef is unnecessary too.

src/library/librarywatcher.cpp
```diff
@@ -764,14 +780,15 @@ QString LibraryWatcher::PickBestImage(const QStringList& images) {
 }
 
 QString LibraryWatcher::ImageForSong(const QString& path,
-                                     QMap<QString, QStringList>& album_art) {
+                                     QMap<QString, QStringList>& album_art,
```
hatstand: This should really be a pointer while you're here.

src/library/librarywatcher.h
```diff
@@ -77,6 +78,18 @@ class LibraryWatcher : public QObject {
   void SetRescanPaused(bool pause);
 
  private:
+  class WatchedDir : public Directory {
+   public:
+    WatchedDir() { active_ = true; }
+
+    void operator=(const Directory& other) {
```
hatstand: I think this is better with an explicit constructor instead.

src/library/librarywatcher.h
jbroadus: Updated.

src/library/librarywatcher.h
jbroadus: Fixed.

src/library/librarywatcher.cpp
jbroadus: Fixed.

src/library/librarywatcher.cpp
```diff
@@ -204,12 +202,33 @@ SubdirectoryList LibraryWatcher::ScanTransaction::GetAllSubdirs() {
...
+}
+
 void LibraryWatcher::AddDirectory(const Directory& dir,
                                   const SubdirectoryList& subdirs) {
-  watched_dirs_[dir.id] = dir;
-  // Use a reference to our copy of the directory, not the reference to the
-  // caller's instance.
-  Directory& new_dir = watched_dirs_[dir.id];
+  watched_dirs_.Add(dir);
+  WatchedDir& new_dir = watched_dirs_.list_[dir.id];
```
hatstand: No non-const refs please.

src/library/librarywatcher.cpp
```diff
@@ -682,21 +701,22 @@ void LibraryWatcher::DirectoryChanged(const QString& subdir) {
...
-    if (stop_requested_) return;
-
-    if (!watched_dirs_.contains(dir)) {
-      qLog(Warning) << "Rescan id" << dir << "not in watch list.";
+  for (int id : rescan_queue_.keys()) {
+    if (!watched_dirs_.list_.contains(id)) {
+      qLog(Warning) << "Rescan id" << id << "not in watch list.";
       continue;
     }
+    WatchedDir& dir = watched_dirs_.list_[id];
```
hatstand: Ditto

src/library/librarywatcher.h
```diff
@@ -77,6 +78,14 @@ class LibraryWatcher : public QObject {
   void SetRescanPaused(bool pause);
 
  private:
+  class WatchedDir : public Directory {
+   public:
+    WatchedDir() { active_ = true; }
```
hatstand: Can you use an initialiser list instead of assignment in both these constructors?

src/library/librarywatcher.cpp
jbroadus: Fixed

src/library/librarywatcher.cpp
jbroadus: fixed

src/library/librarywatcher.h
jbroadus: done

PR #6551, opened 2020-01-26
Early detect device removal

jbroadus:
These two changes fix a remaining timing hole that exists when a device is removed. They also create a mechanism to provide some device information to the watcher, library backend, etc.


jbroadus:
This will probably conflict with my other open PR since the two modify adjacent lines in a header file.

jbroadus:
Thinking more about this, I'm not sure it's quite the right approach. The paradigm fits Unix-like systems, but not windows.

My goal was to share some data from top level library or device object that could provide information such as:
- whether the device is removable
- mount point that may not be the same for each use (Mostly for removable devices)
- if the paths should be relative or absolute (consolidate checks for Application::kIsPortable and Utilities::UrlOnSameDriveAsClementine)

On a windows system, this approach doesn't work for the last point breaks since it would vary from directory to directory, so maybe providing a common interface that can be a Directory factory instead of this MountInfo would be more flexible.

Thoughts?

Issue #6531, opened 2020-01-14
1.4.0rc1 not loading UNC paths; works in 1.3.1


sjuxax:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [x] I checked the issue tracker for similar issues
- [x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [x] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Windows 10.0.18363
- Clementine version: 1.4.0rc1, 1.4.0rc1-63-g456314948

### Expected behaviour / actual behaviour

Expected: plays UNC paths.

Actual: skips track as unavailable. Right-click -> "Show in file browser" opens the location, the library seems to know about it, but when you try to actually play, it skips. Local files appear to work.

### Steps to reproduce the problem (only for bugs)

Load a track with a UNC path (e.g. "`\\my-nas\music\CoolSongs\Track1.mp3`"), observe that track doesn't play.

jbroadus:
Looks like the leading slashes are being lost in the URL conversion.

sjuxax:
Confirmed working for me now with Version 1.4.0rc1-101-gd45a9f5c5. Thanks @jbroadus!

PR #6554, opened 2020-01-29
Fix UNC paths.

jbroadus:
The fix-up for URLs for files that that begin with // no longer works since the
QUrl class determines that these modifications are invalid, resulting in an
empty string when converted. Instead of attempting to modify the QUrl, add a
utility function that makes the correction on the encoded byte array at time of
usage.


jbroadus:
This should fix #6531

PR #6561, opened 2020-01-31
Add a factory class for Directory objects.

jbroadus:
Create a DirectoryFactory class that is owned by the Library or Device objects
and provided to methods that need to create Directory objects. This will allow
information to be provided through the Directory object. For example:
- If the directory is on a removable device.
- A mount point that may change on future uses.
- Whether paths should be relative or absolute.


jbroadus:
Consider this an RFC. This is a pattern that isn't really used anywhere else in the code base, as far as I can tell. An alternative would be to pass around a non-factory object (DirectoryManager?) that can be queried for this information when a Directory is created.

jbroadus:
Also, if you're trying to stabilize a 1.4 release, this is probably 1.4.1.

jbroadus:
Renamed DirectoryFactory -> DirectoryManager to be more general.

src/devices/connecteddevice.cpp
```diff
@@ -61,12 +62,12 @@ ConnectedDevice::ConnectedDevice(const QUrl& url, DeviceLister* lister,
   model_ = new LibraryModel(backend_, app_, this);
 }
 
-ConnectedDevice::~ConnectedDevice() {}
+ConnectedDevice::~ConnectedDevice() { delete directory_manager_; }
```
kentsangkm, line 65: why not to use the unique_ptr?

PR #6564, opened 2020-02-02
Remove embedded qsqlite [rfc/test/wip]

aerusso:
Removes an embedded copy of qsqlite.  According to a comment in CMakeLists.txt (removed by this patch):

> Build our copy of QSqlLiteDriver.
We do this because we can't guarantee that the driver shipped with Qt exposes the
raw sqlite3_ functions required for FTS support. This way we know that those symbols
exist at compile-time and that our code links to the same sqlite library as the
Qt driver.

It's not clear (to me) if this is still the case (comment is from 2013).  Using qt5's qsqlite builds and runs fine on Debian Linux.

This PR was prompted by @jbroadus's comment in #6560. @hatstand Can you comment on this? If the embedded qsqlite is still needed on Mac and Windows, I can wrap it in `ifdef`'s so at least Linux users don't have to deal with the embedded copy of qsqlite.  Some comments in the code suggested that it was only Windows and Mac that needed the embedded qsqlite, but it didn't seem like there were any such conditionals.


hatstand:
The reasoning in the comment is still valid AFAIK. We need SQLite to be compiled with FTS3 support and we don't know if this is the case everywhere.

aerusso:
Understood. I think this patch should be applied at the distribution level (i.e., I'll submit it to Debian) where I can be confident in the compile-time features of all the software.

jbroadus:
@hatstand @aerusso The sqlite library itself was removed in 4c23072be, so Clementine has been depending on the distributed library since that time. I think the thing we might still depend on is ClementineSqlCachedResult.

aerusso:
@jbroadus Sorry, I'm confused now. I assumed @hatstand was referring to qsqlite, which is what this patch removes. If I understand correctly, some distributions may not enable FTS3 at compile time---Debian does, and so my testing does not demonstrate any problems. Other distributions may not, and users of those would benefit from having the embedded qsqlite (because it is built with FTS3 enabled).

Because Debian does build qsqlite with FTS3 enabled, I should expect no complications by building with this patch. Regarding `ClementineSqlCachedResult` (which I did not notice!), it doesn't appear to be referenced by anything outside of the qsqlite package. Given that Clementine seems to run fine without it (in particular, it's not a compile time dependency), I'm guessing that it must be either dead code, or "for the convenience of other QT classes" (as described in the `.h`) which may have something to do with allowing the transplanted qsqlite library to integrate into Clementine (this is pure conjecture).

Is this an accurate description of the situation?

jbroadus:
@aerusso Prior to 4c23072, sqlite3.c was included in 3rdparty/qsqlite/ and a static libsqlite library was built from that using FTS3 options. The current qsqlite driver links against the installed shared library.

But the original comment does refer to qsqlite itself, so it's likely something that I don't understand about how the drivers work.

jbroadus:
Revisiting this because there is a build breakage in the 3rd party library and it looks like FTS3 is enabled by default in qt5:
https://github.com/qt/qtbase/commit/ee69c935c0f7b4ad3747e369fc18373d31e2521b

We could leave the 3rd party code in for now and enable it with a config option.

PR #6574, opened 2020-02-13
Convert remaining QJsonValue::isUndefined usage.

jbroadus:
Noted previously, using the [] operator on a non-const QJsonObject causes the
creation of the key and does not work for checking existence. Convert the
remaining isUndefined call sites to use QJsonObect::contains.


jbroadus:
Full disclosure: this change hasn't been thoroughly tested since I don't have a Digitally Imported account and SongKicks is currently broken.

jbroadus:
Working on a fix for songkicks, so I was able to test that one.

PR #6575, opened 2020-02-13
Add a logging category to log all network requests.

jbroadus:
New logging class NetworkRequests disabled by default, but can be enabled with
the command line option.


jbroadus:
I know there was a recent concern about logging network requests, but I didn't know if this would be an issue if the option had to be manually enabled.

hatstand:
Fine if it's not by default - the concern is users attaching log files containing their credentials.

PR #6577, opened 2020-02-14
Fix dropbox

jbroadus:


src/internet/dropbox/dropboxservice.cpp
```diff
@@ -267,6 +277,29 @@ QUrl DropboxService::GetStreamingUrlFromSongId(const QUrl& url) {
...
 
-  QJsonObject json_response = QJsonDocument::fromJson(reply->readAll()).object();
+  QJsonDocument document;
+  if (!ParseReply(reply, &document)) return QUrl();
+
+  QJsonObject json_response = document.object();
   return QUrl::fromEncoded(json_response["link"].toVariant().toByteArray());
 }
+
+// Note that this would look nicer if it returned a QJsonDocument, but move
```
hatstand: I don't think we care about the cost of a copy TBH.

jbroadus: Changed the interface and also moved to the base class for future reuse.

PR #6578, opened 2020-02-16
Scrub debug urls

jbroadus:
While debugging an issue, I came across a couple places where auth tokens were being logged in url query strings.

### DIFF
--- a/src/core/utilities.cpp
+++ b/src/core/utilities.cpp
@@ -760,6 +760,13 @@ QByteArray GetUriForGstreamer(const QUrl& url) {
   return url.toEncoded();
 }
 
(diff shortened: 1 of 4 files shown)


src/core/utilities.cpp
```diff
@@ -760,6 +760,12 @@ QByteArray GetUriForGstreamer(const QUrl& url) {
   return url.toEncoded();
 }
 
+QString ScrubUrlQueries(QString& str) {
```
hatstand: I think this should be a `const &`?

jbroadus: Yep. Fixed.

PR #6580, opened 2020-02-17
Gdrive get file

jbroadus:
This fixes GetStreamingUrlFromSongId for Google Drive. 

With this change, playback is still broken due to an API change:
https://cloud.google.com/blog/products/application-development/upcoming-changes-to-the-google-drive-api-and-google-picker-api

I'm looking at solutions for that.

### DIFF
--- a/src/internet/googledrive/googledriveclient.cpp
+++ b/src/internet/googledrive/googledriveclient.cpp
@@ -72,7 +72,7 @@ QStringList File::parent_ids() const {
 ConnectResponse::ConnectResponse(QObject* parent) : QObject(parent) {}
 
 GetFileResponse::GetFileResponse(const QString& file_id, QObject* parent)
-    : QObject(parent), file_id_(file_id) {}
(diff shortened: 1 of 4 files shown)


hatstand:
We do, or used to pass auth headers via the URI fragment and then fish it out in `GstEngine` somehow.

hatstand:
https://github.com/clementine-player/Clementine/commit/0beb6d3c599b22eb926ee7dfee18a0da80559e86#diff-a1f8da8007e51dcea858c84fe1eedba2

jbroadus:
> [0beb6d3#diff-a1f8da8007e51dcea858c84fe1eedba2](https://github.com/clementine-player/Clementine/commit/0beb6d3c599b22eb926ee7dfee18a0da80559e86#diff-a1f8da8007e51dcea858c84fe1eedba2)

We could do that, but it might be nice to avoid service specific logic in the engine. Since we already pass a few pieces of data to SetNextUrl, I thought that we might be able to consolidate the playback info into a single object that could also contain things like auth strings.

PR #6581, opened 2020-02-18
RFC: Add MediaPlaybackData class.

jbroadus:
Add a class to wrap the URL in the playback engines. In the future, this will
contain authentication information for the specified URL. It can also include
the start and end time as well as other data that is currently specified along with
the URL.

If you're not comfortable with this approach, we can use the URL fragment mentioned in the other pull request. I'm also happy to take suggestions on a better name than MediaPlaybackData.

### DIFF
--- a/src/core/player.cpp
+++ b/src/core/player.cpp
@@ -146,9 +146,10 @@ void Player::HandleLoadResult(const UrlHandler::LoadResult& result) {
         item->SetTemporaryMetadata(song);
         app_->playlist_manager()->active()->InformOfCurrentSongChange();
       }
-      engine_->Play(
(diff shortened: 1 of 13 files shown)


hatstand:
General idea LGTM but it'll be quite a bit of plumbing. How about `MediaPlaybackRequest` to mirror `QNetworkRequest`?

jbroadus:
Mirroring QNetworkRequest is a good idea.

Really, it's not as much plumbing as I had initially thought.

jbroadus:
These changes now fix the google drive auth problem. Still need some clean-up. I'll probably look at share pointers for the media request, especially since the gstreamer source setup callback, which handles the headers, is on a different thread.

Issue #6583, opened 2020-02-18
/usr/bin/data usage + library scan errors


jnvandermeer:
Dear developers,

I have installed Clementine 1.3.1 on Arch. Specifically version 1.3.1.r831.664c5a31f-3

Is it normal that the /usr/bin/data directory is used to store the databases jamendo.db and clementine.db? The location is highly unusual and I would expect more of a ~/.clementine folder. But this might be a packaging issue on Arch side.

Also, when I update my Music folder, it will miss each file. This is because the folders in the database miss one 'level' of folder.

Example: When a music file is in:
/data/Music/a/b/c/file.mp3

The database reads:
/data/Music/b/c/file.mp3

The name of the 'a' folder doesn't matter - it will be removed, so it cannot find anything. Is this something you came across?

Thanks for your help
Best wishes from Australia
Johan

### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [ x] I checked the issue tracker for similar issues
- [ x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [ ] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System:
- Clementine version:

jbroadus:
There was a portable feature added in 2013 that checks if a "data" directory exists in same location as the clementine executable. I'm not sure what the use case was for this or if it's still valid. I assume the data directory already existed?

See commit acb18cdcde.

jbroadus:
Oh, and those paths in the database should be fixed at e7b066719

jnvandermeer:
Thanks for your quick reply. 

I cannot remember what made the data directory in /usr/bin, but apparently Spotify and one other program used that. I'll try rebuilding in a few days.

jbroadus:
The spotify files might be from Clementine integration, but a Google search shows a few packages that do use this directory, though none should (I assume the Clementine case was intended for a USB stick or something). Other people are probably hitting this as well, so maybe we should consider something more unique like clementine-data

PR #6584, opened 2020-02-19
Prevent accidental usage of /usr/bin/data

jbroadus:
Clementine detects a data directory in the same directory as the executable to
determine portable configuration. But there are some packages that create
/usr/bin/data, causing Clementine to run in portable mode.

Use a more unique data directory name, clementine-data, as the portable data
directory. For backwards compatibility, use the legacy data directory if the
already exists there.

This is a response to #6583. I don't see the portable mode documented in the readme or the wiki.

### DIFF
--- a/src/core/application.cpp
+++ b/src/core/application.cpp
@@ -66,6 +66,9 @@
 #endif
 
 bool Application::kIsPortable = false;
+const char* Application::kLegacyPortableDataDir = "data";
+const char* Application::kDefaultPortableDataDir = "clementine-data";
+const char* Application::kPortableDataDir = nullptr;
 
 class ApplicationImpl {
  public:
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -197,15 +197,33 @@ void ParseAProto() {
 }
 
 void CheckPortable() {
-  QFile f(QApplication::applicationDirPath() + QDir::separator() + "data");
-  if (f.exists()) {
+  QDir appDir(QApplication::applicationDirPath());
+  // First look for legacy data location.
+  QDir d(appDir.filePath(Application::kLegacyPortableDataDir));
+  // Key off of database file since config name may vary depending on platform.
+  if (d.exists("clementine.db")) {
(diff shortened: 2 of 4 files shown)


jbroadus:
Looks like builds are offline.

src/core/#urlhandler.h#
```diff
@@ -0,0 +1,91 @@
+/* This file is part of Clementine.
```
hatstand: Not sure what happened to this file?

jbroadus: My clumsiness.

src/core/application.cpp
```diff
@@ -66,6 +66,7 @@
 #endif
 
 bool Application::kIsPortable = false;
+QString Application::kPortableDataDir;
```
hatstand: We generally avoid non-POD static data. I think a `const char[]` will work here fine?

src/main.cpp
```diff
@@ -100,6 +100,9 @@ const QDBusArgument& operator>>(const QDBusArgument& arg, QImage& image);
 #include <QtPlugin>
 Q_IMPORT_PLUGIN(QSQLiteDriverPlugin)
 
+const char* legacy_portable_data_location = "data";
```
hatstand: These can be inside the anonymous namespace and should be named like `kLegacyPortableDataLocation`

src/core/application.cpp
jbroadus: Don't think that would work (dynamic sizing of static data), but I can use a const char* that points to one of the other static pointers.

src/core/application.cpp
```diff
@@ -66,6 +66,9 @@
 #endif
 
 bool Application::kIsPortable = false;
+const char* Application::kLegacyPortableDataDir = "data";
+const char* Application::kDefaultPortableDataDir = "clementine-data";
+const char* Application::kPortableDataDir = NULL;
```
hatstand: `nullptr`

jbroadus: Fixed

Issue #6587, opened 2020-02-22
Clementine segfaults when QT_QPA_PLATFORM=wayland


JacobHenner:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [x] I checked the issue tracker for similar issues
- [x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [x] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Arch Linux
- Clementine version: 1.4.0rc1-128-ge1c872666 

### Expected behaviour / actual behaviour
I expect Clementine to run under wayland. However, when I run clementine with QT_QPA_PLATFORM=wayland under GNOME, it immediately crashes with a segfault (no other output).

I've also tried on Weston - same issue.

I use two monitors with fractional scaling enabled. The issue occurs regardless of how many monitors are connected.

### Steps to reproduce the problem (only for bugs)

Run `QT_QPA_PLATFORM=wayland clementine`

```
Starting program: /usr/bin/clementine                                                                                                                                               
[Thread debugging using libthread_db enabled]                                                                                                                                       
Using host libthread_db library "/usr/lib/libthread_db.so.1".                                                                                                                       
__logging_message__20:44:14.549 DEBUG unknown                          Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.   
[New Thread 0x7fffec13d700 (LWP 58603)]                                                                                                                                             
[New Thread 0x7fffeb93c700 (LWP 58604)]                                                                                                                                             
[New Thread 0x7fffeb0ec700 (LWP 58605)]                                                                                                                                             
[New Thread 0x7fffea8eb700 (LWP 58606)]                                                                                                                                             
[New Thread 0x7fffea0d6700 (LWP 58607)]                                                                                                                                             
[New Thread 0x7fffe9715700 (LWP 58608)]                                                                                                                                             
[New Thread 0x7fffe8f14700 (LWP 58609)]                                                                                                                                             
[New Thread 0x7fffd3fff700 (LWP 58610)]                                                                                                                                             
[Detaching after fork from child process 58611]                                                                                                                                     
[Detaching after fork from child process 58612]                                                                                                                                     
[Detaching after fork from child process 58613]                                                                                                                                     
[...]

PR #6589, opened 2020-02-23
Short-term fix for shortcut crash

jbroadus:
The 3rd party qxt library assumes X11 on Linux systems. Don't register
QxtGlobalShortcutBackend when using Wayland.


jbroadus:
Immediate fix for #6587.

Issue #6591, opened 2020-02-23
Playlist search causes segmentation fault


TheBlackKoala:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [X ] I checked the issue tracker for similar issues
- [X] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [x] I tried the latest Clementine build from [here](https://builds.clementine-player.org/) - i always have the latest build from AUR.

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: 5.5.5-arch1-1
- Clementine version: Clementine 1.3.1-831-g664c5a31f

### Expected behaviour / actual behaviour
Search in playlists shows the resulting songs from resulting playlists (it seem to be a new feature and i love it!) - after the first search clementine is liable to clash with a segmentation fault

### Steps to reproduce the problem (only for bugs)
Search in the input-field in the playlist tab.
Search for something else.
Try a few times if it doesn't crash at first.

My longest playlist is a little more than 5k items and i have playlists sorted into folders (4 folders, 36 playlists in total).

TheBlackKoala:
To clearify - the search that causes the segmentation fault is in the same place that you would search your library in the library tab but for playlists.

jbroadus:
I was able to repro after adding folders.

TheBlackKoala:
@jbroadus so without folders it does not occur? Might go without those until it has been resolved if that is the case.

jbroadus:
Yeah, I wasn't able to repro without folders. Let me know if your experience is different. I'll try to dig a little more later.

jbroadus:
The crash occurs while iterating over the QList expandList in PlaylistListFilterProxyModel.
```
  void refreshExpanded(QTreeView *tree) {
    tree->collapseAll();
    for(QModelIndex sourceIndex : expandList ) {
      QModelIndex mappedIndex = mapFromSource( sourceIndex );
      tree->setExpanded( mappedIndex, true );
    }
  }
```
The setExpanded eventually causes an overloaded function in the class to append expandList. QList is implemented as an array, not a linked list, so I think that when it grows beyond its preallocated size and reallocates the array, the iterators become garbage. The good news is that the data can be accessed efficiently using indexes, so the iterators aren't necessary.

jbroadus:
@TheBlackKoala This fixes the issue for me. If you're able to test the latest, let me know if it works for you.

[...]

PR #6593, opened 2020-02-24
Fix crash when searching playlist.

jbroadus:
While iterating over expandList in refreshExpanded, calls to setExpanded cause
the list to be appended. Since a QList uses and array implementation that must
resize when reserved space is exhausted, iterators are unsafe for this case.
Use indexes, which are O(1) in QLists, instead of iterators for this case.


jbroadus:
This is a fix for #6591

PR #6640, opened 2020-04-16
Fix tranlations test.

planglois:
Hello there :-)

When packaging clementine for guix, we run the unittests. I noticed the pre-release was failing this test because of an out-dated translation so here's a fix!

Thanks,
Pierre


fitojb:
💜

planglois:
Thanks for the merge! Looking forward to the 1.4 release :-).

PR #6653, opened 2020-05-09
Compiler & library modernisation

Crystalix007:
I was previously unable to compile Clementine with the combination of old headers and historic C++ compiler flags specified.

This patch makes the compiler build with standard `C++11` instead of the former `gnu++98` standard, which required fixing the reference comparison errors.

`GLEW`'s headers are the latest release available from its SourceForge website.


hatstand:
This broke the mac build: https://app.circleci.com/pipelines/github/clementine-player/Clementine/478/workflows/3ccf2270-0b74-444c-bc8f-221e2cc13dca/jobs/5196

Crystalix007:
Apologies for that. Mac build is using clang, which was being more stringent about applying C++11 rules. I've attempted to fix on Crystalix007/clementine@52bb29f with clang-tidy, and it seems to build with clang (on Linux). I don't have a way to test on OSX , unfortunately.

hatstand:
I started adding a Github Actions workflow that includes a mac builder - it should be possible to use that branch to test OSX: https://github.com/clementine-player/Clementine/pull/6661

Crystalix007:
So, I rebased the branch (minus the reverts) and the compile seems to work [fine](https://github.com/Crystalix007/Clementine/runs/674410429).

I assume you don't rebase master, so I need to open another PR on top, right?

hatstand:
Yep, you'll need to open a new PR.

PR #6680, opened 2020-05-21
Compressed PNGs

dmdmdm:
I used advpng to compress all the PNGs.  The resulting clementine executable is 132K smaller (on Linux).
The exact command was advpng -z -4 *.png


hatstand:
This seems mixed in with other commits and conflicts with your earlier PR.

PR #6681, opened 2020-05-21
Guess title and artist from filename

dmdmdm:
I have many files named like Led_Zeppelin_-_Dazed_and_Confused.mp3 with no id3 tags.
This change makes the playlist look much nicer by grabbing the artist from before the dash
and the title from after.  Handles spaces or underscores. Additionally when you click on Edit Track Information the filename artist and title are there as default so you can just click [Save] to update the id3 tag - saves typing.

### DIFF
--- a/src/core/song.cpp
+++ b/src/core/song.cpp
@@ -41,6 +41,7 @@
 #include <QTime>
 #include <QVariant>
 #include <QtConcurrentRun>
+#include <QRegExp>
 
 #ifdef HAVE_LIBLASTFM
 #include "internet/lastfm/fixlastfm.h"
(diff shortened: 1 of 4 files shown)


dmdmdm:
I see your point.  Its really a hack no matter where is done ;)
I'll move it to the tag reader.

As a more deluxe enhancement...
It would be nice if the Edit Track Information dialog
showed the artist and title in bold (ie changed) if they came from this guessing instead of
a proper id3 tag.   That way the user would know that pressing [Save] would update the
id3 tag with the guess and [Cancel] would mean the id3 tag would remain empty when read
in another player.

PR #6683, opened 2020-05-22
Guess artist and title from filename (second attempt)

dmdmdm:
Moved to tagreader.

Will extract artist and title from a filename like Led_Zeppelin_-_Dazed_and_Confused.mp3
if there are no id3 tags.

### DIFF
--- a/ext/libclementine-tagreader/tagreader.cpp
+++ b/ext/libclementine-tagreader/tagreader.cpp
@@ -130,6 +130,51 @@ const char* kASF_OriginalDate_ID = "WM/OriginalReleaseTime";
 const char* kASF_OriginalYear_ID = "WM/OriginalReleaseYear";
 }  // namespace
 
+void TagReader::RemoveExtension(std::string* s) const {
+  if (s->empty()) return;
+  const int i = s->find_last_of('.');
+  if (i == std::string::npos) return;
+  s->assign(s->substr(0, i));
+}
+
+void TagReader::ChangeUnderscores(std::string* s) const {
+  for (int i = 0; i < s->size(); i++) {
+    if (s->at(i) == '_') s->at(i) = ' ';
+  }
+}
+
+std::string& TagReader::ltrim(std::string& str, const std::string& chars) const {
+    str.erase(0, str.find_first_not_of(chars));
+    return str;
+}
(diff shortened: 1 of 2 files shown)


ext/libclementine-tagreader/tagreader.cpp
```diff
@@ -130,6 +130,51 @@ const char* kASF_OriginalDate_ID = "WM/OriginalReleaseTime";
...
+  s->assign(s->substr(0, i));
+}
+
+void TagReader::ChangeUnderscores(std::string* s) const {
+  for (int i = 0; i < s->size(); i++) {
+    if (s->at(i) == '_') s->at(i) = ' ';
+  }
+}
+
+std::string& TagReader::ltrim(std::string& str, const std::string& chars) const {
```
hatstand, line 146: We only use pointers, const references and values to keep things readable: https://google.github.io/styleguide/cppguide.html

ext/libclementine-tagreader/tagreader.cpp
```diff
@@ -130,6 +130,51 @@ const char* kASF_OriginalDate_ID = "WM/OriginalReleaseTime";
...
+    return str;
+}
+
+std::string& TagReader::trim(std::string& str, const std::string& chars) const {
+    return ltrim(rtrim(str, chars), chars);
+}
+
+void TagReader::GuessArtistAndTitle(pb::tagreader::SongMetadata *song) const {
+  const std::string bn = song->basefilename();
+  std::string *artist = song->mutable_artist();
```
hatstand, line 162: Calling `mutable` means that `has_artist()` will no longer return `false` even if you leave it empty.

ext/libclementine-tagreader/tagreader.h
```diff
@@ -95,6 +95,14 @@ class TagReader {
       TagLib::Ogg::XiphComment* vorbis_comments,
       const pb::tagreader::SongMetadata& song) const;
 
+  // Helpers for GuessArtistAndTitle()
+  void RemoveExtension(std::string*) const;
```
hatstand, line 99: These helpers don't need to be in the header, they can be only in the `.cpp` inside an anonymous namespace.

PR #6684, opened 2020-05-22
Guess artist and title from filename (third attempt)

dmdmdm:
Made requested changes.

### DIFF
--- a/ext/libclementine-tagreader/tagreader.cpp
+++ b/ext/libclementine-tagreader/tagreader.cpp
@@ -19,6 +19,7 @@
 
 #include <memory>
 #include <algorithm>
+#include <regex>
 
 #include <QCoreApplication>
 #include <QDateTime>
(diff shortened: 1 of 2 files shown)


ext/libclementine-tagreader/tagreader.cpp
```diff
@@ -130,6 +131,52 @@ const char* kASF_OriginalDate_ID = "WM/OriginalReleaseTime";
...
+
+void TagReader::GuessArtistAndTitle(pb::tagreader::SongMetadata *song) const {
+  std::string artist = song->artist();
+  std::string title = song->title();
+  const std::string bn = song->basefilename();
+  if (!artist.empty() || !title.empty()) return;
+  if (bn.empty()) return;
+  int i = bn.find(" - ");
+  if (i == std::string::npos) i = bn.find("_-_");
+  if (i == std::string::npos) return;
```
jbroadus, line 168: Have you thought about using std::regex? You could match both separators and remove the extension in a single expression.

dmdmdm: ok will do

dmdmdm:
Changed to use std::regex

PR #6688, opened 2020-05-24
Adding sq translation

ujdhesa:


hatstand:
Translations live in Transifex and are automatically synced: https://www.transifex.com/davidsansome/clementine/

ujdhesa:
> Translations live in Transifex and are automatically synced: https://www.transifex.com/davidsansome/clementine/
Thank you! 
I think I tried to aply there. Knowing nothing about the current state of the project, I thought a PR might do the job as well…

PR #6705, opened 2020-05-28
Change text for Original Year

dmdmdm:
When editing a smart playlist "Original Year" comes before "Year" alphabetically (of course).
But the user might want Year.  So I changed the text to "Year - original" so the user
sees both kinds of year beside each other in the list and can make a better choice.

### DIFF
--- a/src/playlist/playlist.cpp
+++ b/src/playlist/playlist.cpp
@@ -1388,7 +1388,7 @@ QString Playlist::column_name(Column column) {
     case Column_Year:
       return tr("Year");
     case Column_OriginalYear:
-      return tr("Original year");
+      return tr("Year - original"); // Done like this so sorts beside "Year" when editing smart playlists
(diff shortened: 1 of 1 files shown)


src/playlist/playlist.cpp
```diff
@@ -1388,7 +1388,7 @@ QString Playlist::column_name(Column column) {
     case Column_Year:
       return tr("Year");
     case Column_OriginalYear:
-      return tr("Original year");
+      return tr("Year - original"); // Done like this so sorts beside "Year" when editing smart playlists
```
hatstand, line 1391: Can you make this a translator comment so translators know to try to do this for their language too? https://doc.qt.io/qt-5/i18n-source-translation.html#translator-comments

dmdmdm: Right, done

PR #6710, opened 2020-06-01
Fix cd playback

jbroadus:
This is a follow-on to change 335bc89. That changed fixed cdda URLs without paths but broke the schemes on URLs with paths.

### DIFF
--- a/src/devices/cddadevice.h
+++ b/src/devices/cddadevice.h
@@ -45,6 +45,17 @@ class CddaDevice : public ConnectedDevice {
 
   static QStringList url_schemes() { return QStringList() << "cdda"; }
 
+  // QUrl interprets a single number as an ip address, so the QString cdda://1
+  // would become the QUrl cdda://0.0.0.1. For this reason, we append a single
+  // character to strings when converting to URLs and strip that character
+  // when doing the reverse conversion.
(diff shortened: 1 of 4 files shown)


jbroadus:
I guess I'll need to move the helper functions elsewhere. Maybe `core/utilities.h`, but it feels too specific to be there.

hatstand:
I think this needs some `ifdef` for when `libcdio` is missing.

jbroadus:
> I think this needs some `ifdef` for when `libcdio` is missing.

Yeah, I guess it makes sense to #ifdef the code that strips the extra character if the code that adds it isn't built.

hatstand:
It's the `#include` in `gstenginepipeline.cpp` that breaks the build. You can see the build error here: https://github.com/clementine-player/Clementine/pull/6710/checks?check_run_id=726353929

jbroadus:
> It's the `#include` in `gstenginepipeline.cpp` that breaks the build. You can see the build error here: https://github.com/clementine-player/Clementine/pull/6710/checks?check_run_id=726353929

Right. I was going to move the helpers to utilities or another place that doesn't include cdio. But I think it makes sense to add ifdefs in gstenginepipeline since the cdda code paths shouldn't be used if HAVE_AUDIOCD isn't set.

PR #6713, opened 2020-06-04
Use builtin taglib on Windows and set -DTAGLIB_STATIC

jonaski:


hatstand:
Isn't this doing the opposite of the description? It's using the _system_ taglib on windows if it's there.

jonaski:
No. It will use the system version by default for windows too if the verison is higher than 1.11.1, otherwise use the static one. It was failing previously because it didn't set TAGLIB_STATIC, that's why I did the exception for Windows in the first place since the static taglib failed to build.

PR #6762, opened 2020-07-26
Share the love

dmdmdm:
Previously the Love button only worked for Lastfm play.
I modified it to work for local files too.
When the user clicks, they get a small dialog with the name of the playing track and a rating control.
Clicking on the rating (eg 4 stars of out 5) will:
- Save it in the mp3 file 
- Update the display in the current playlist (if has rating column)
- Close the dialog

### DIFF
--- a/src/ui/lovedialog.cpp
+++ b/src/ui/lovedialog.cpp
@@ -68,12 +68,11 @@ void LoveDialog::showEvent(QShowEvent* e) {
 
 void LoveDialog::SongRated(float rating) {
   qLog(Debug) << "LoveDialog::SongRated: rating=" << rating;
-  Song* song = &song_;
-  qLog(Debug) << "LoveDialog::SongRated: isValid=" << song->is_valid() << " id=" << song->id();
-  if (!song->is_valid() || song->id() == -1) return;
+  qLog(Debug) << "LoveDialog::SongRated: isValid=" << song_.is_valid() << " id=" << song_.id();
+  if (!song_.is_valid() || song_.id() == -1) return;
 
   qLog(Debug) << "LoveDialog::SongRated: song is valid and id is good";
-  song->set_rating(rating);
-  app_->library_backend()->UpdateSongRatingAsync(song->id(), rating);
+  song_.set_rating(rating);
+  app_->library_backend()->UpdateSongRatingAsync(song_.id(), rating);
   QDialog::accept();
 }
--- a/src/ui/lovedialog.cpp
+++ b/src/ui/lovedialog.cpp
@@ -1,0 +1,79 @@
+/* This file is part of Clementine.
+   Copyright 2020, Dave McKellar
+
+   Clementine is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation, either version 3 of the License, or
+   (at your option) any later version.
+
+   Clementine is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with Clementine.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "ui/lovedialog.h"
+#include "ui_lovedialog.h"
+
+#include <QLabel>
+#include <QPushButton>
+
+#include "core/application.h"
+#include "core/logging.h"
+#include "core/utilities.h"
+#include "library/librarybackend.h"
+
+const char* LoveDialog::kSettingsGroup = "LoveDialog";
+
+LoveDialog::LoveDialog(Application* app, QWidget* parent)
+    : QDialog(parent),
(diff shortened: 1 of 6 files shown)


src/ui/lovedialog.cpp
```diff
@@ -0,0 +1,79 @@
...
+  // Restore the tab that was current last time.
+  QSettings s;
+  s.beginGroup(kSettingsGroup);
+
+  QDialog::showEvent(e);
+}
+
+void LoveDialog::SongRated(float rating) {
+  qLog(Debug) << "LoveDialog::SongRated: rating=" << rating;
+  Song* song = &song_;
```
hatstand, line 71: I don't think you need this variable, just use `song_` directly.

src/ui/lovedialog.h
```diff
@@ -0,0 +1,59 @@
...
+   along with Clementine.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef LOVE_DIALOG_H
+#define LOVE_DIALOG_H
+
+#include <QDialog>
+
+#include "core/song.h"
+//#include "widgets/lineedit.h"
```
hatstand, line 24: Remove

src/ui/mainwindow.cpp
```diff
@@ -1582,13 +1586,28 @@ void MainWindow::ScrobbledRadioStream() {
   ui_->action_love->setEnabled(true);
   if (tray_icon_) tray_icon_->LastFMButtonLoveStateChanged(true);
 }
+#endif
 
 void MainWindow::Love() {
-  app_->scrobbler()->Love();
-  ui_->action_love->setEnabled(false);
-  if (tray_icon_) tray_icon_->LastFMButtonLoveStateChanged(false);
+  Playlist* activePlaylist = app_->playlist_manager()->active();
```
hatstand, line 1592: `active_playlist`

src/ui/mainwindow.cpp
```diff
@@ -1582,13 +1586,28 @@ void MainWindow::ScrobbledRadioStream() {
...
+    qLog(Warning) << "Love: nothing playing so can't love it";
+    return;
+  }
+
+  if (item->IsLocalLibraryItem()) {
+    Song song = item->Metadata();
+    love_dialog_->SetSong(song);
+    love_dialog_->show();
+  }
+  else {
```
hatstand, line 1605: Same line as above

src/ui/mainwindow.cpp
```diff
@@ -1582,13 +1586,28 @@ void MainWindow::ScrobbledRadioStream() {
...
+  Playlist* activePlaylist = app_->playlist_manager()->active();
+  PlaylistItemPtr item = activePlaylist->current_item();
+  if (!item) {
+    // Don't make a big deal about it
+    qLog(Warning) << "Love: nothing playing so can't love it";
+    return;
+  }
+
+  if (item->IsLocalLibraryItem()) {
+    Song song = item->Metadata();
```
hatstand, line 1601: `const Song& song`

src/ui/lovedialog.cpp
```diff
@@ -0,0 +1,78 @@
...
+  QSettings s;
+  s.beginGroup(kSettingsGroup);
+
+  QDialog::showEvent(e);
+}
+
+void LoveDialog::SongRated(float rating) {
+  qLog(Debug) << "LoveDialog::SongRated: rating=" << rating;
+  qLog(Debug) << "LoveDialog::SongRated: isValid=" << song_.is_valid() << " id=" << song_.id();
+  if (!song_.is_valid() || song_.id() == -1) return;
```
jbroadus, line 72: This is done elsewhere, but I think it's actually broken. From what I saw, a non-library song from a playlist will be given the id from the playlist table. So, the rating would be set for the wrong item.

That said, this is no worse than what the tag editing dialog is currently doing. I was testing a changeset that set a library flag in the song metadata. I'll dig that out and see if it's ready for review.

dmdmdm: In mainwindow.cpp I do `if (item->IsLocalLibraryItem()) { ...` before sending the user to LoveDialog.  So it should be a library song in LoveDialog::SongRated() or that's what I think ;).  I look forward to hearing how your testing went.  I want this to work!

jbroadus: Yep.  IsLocalLibraryItem() should insure that you're using the correct backend.

jonaski:
I'm surprised this wasn't stopped. It's now impossible to love songs that are part of the library, for what purpose? The ratings are easily accessible from the playlist, if there should be an additional dialog for rating it should be a separate button with it's own setting, at least not replace loving songs when both last.fm and the love button is enabled in settings.

dmdmdm:
I'm at a place with really limited internet. The reason for the change:
rating control isn't in the playlist by default, as a beginner user I
clicked on the heart for a local track and got an error about last.fm -
very confusing and unfriendly. My intention was to make local rating easy
without breaking last.fm love. But sounds like it broke a bit. I can
improve it next week. Thanks for the feedback

Echocs7:
Can I understand why `MainWindow::Love()` is outside `#ifdef HAVE_LIBLASTFM` while it is still inside `#ifdef HAVE_LIBLASTFM` in mainwindow.h ?

PR #6769, opened 2020-08-05
Replaced the heart icon with a star for local files

dmdmdm:
Since pressing the :heart: now takes the user to a dialog to rate local library items
it made sense to change it to a :star: for local items.

I made the star blue rather than gold/yellow because I only want to tell the user what pressing the button will do.  I don't want to mislead the user into thinking that its a favourite as a bright colour might suggest.  Of course, you can use the blue star button to give a song a good or bad rating.  After you have rated it stays enabled so you can re-rate if you change your mind.


src/ui/mainwindow.cpp
```diff
@@ -1163,17 +1163,29 @@ void MainWindow::MediaPlaying() {
...
-  bool can_seek = !(app_->player()->GetCurrentItem()->options() &
-                    PlaylistItem::SeekDisabled);
+  bool can_seek = !(item->options() & PlaylistItem::SeekDisabled);
   ui_->track_slider->SetCanSeek(can_seek);
 
   // We now always enable Love when playing since it works for local files
   ui_->action_love->setEnabled(true);
 
+  // Set the rate/love icon
+  if (item) {
```
jbroadus, line 1178: We've already dereferenced `item` by this point. Maybe better to check for nullptr, log an error, and return at line 1167?

dmdmdm: Good point, will do

PR #6796, opened 2020-09-16
Fix logout cancelling

### LINKED ISSUE #6401: Logout/Reboot Canceled by Clementine in KDE5
### Before posting

- [ x] I checked the issue tracker for similar issues
- [ x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [ x] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

- Operating System:
Arch Linux with KDE Plasma
- Clementine version: 
1.3.1.r771.75f18dab2-1

I just click reboot and even with music stopped, if clementine is in the background it prevents me from rebooting. Just annoying

Only happens in KDE5 as far as I know.
1.) Open clementine
2.) Click reboot/logout
3.) The reboot process is stopped with a popup saying clementine is responsible

Korvox:
Closes #6401

Never heard anyone on Win / OSX having a problem with this so might as well merge it finally.

### DIFF
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -1459,8 +1459,10 @@ void MainWindow::StopAfterCurrent() {
 }
 
 void MainWindow::closeEvent(QCloseEvent* event) {
-  if (!tray_icon_ || !settings_.value("keeprunning", tray_icon_->IsVisible()).toBool())
-        Exit();
+  if (!tray_icon_ ||
+      !settings_.value("keeprunning", tray_icon_->IsVisible()).toBool())
+    Exit();
+
   QMainWindow::closeEvent(event);
 }
 
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -1459,7 +1459,7 @@ void MainWindow::StopAfterCurrent() {
 }
 
 void MainWindow::closeEvent(QCloseEvent* event) {
-  if (!tray_icon_ || !tray_icon_->IsVisible())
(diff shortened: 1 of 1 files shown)


jonaski:
This minimizes Clementine to the system tray instead of exiting when exiting with [X] even when the keep running setting is off.
You just removed the entire code, how does that fix anything?

Korvox:
It fixes the bug by not ignoring close events when the tray is active. Before system close events were being ignored and thus Clementine was preventing logout.

src/ui/mainwindow.cpp
```diff
@@ -1465,17 +1465,9 @@ void MainWindow::StopAfterCurrent() {
...
-    keep_running =
-        settings_.value("keeprunning", tray_icon_->IsVisible()).toBool();
-
-  if (keep_running && event->spontaneous()) {
-    event->ignore();
-    SetHiddenInTray(true);
-  } else {
-    Exit();
-  }
+  if (!tray_icon_ || !settings_.value("keeprunning", tray_icon_->IsVisible()).toBool())
```
jonaski, line 1468: This will now crash on systems without a system tray icon.

Korvox: No it won't, if tray_icon_ is null it short circuits to Exit.

Unless there is some circumstance where tray_icon_ is non-null but tray_icon_->IsVisible is an invalid op? But that would have crashed the previous code.

jonaski: Yes your right, my bad

hatstand:
If you rebase, CI might pass.

PR #6809, opened 2020-10-03
add CLI play-playlist option, to play given playlist name.

NicolasToussaint:
Adding a command line option to start playing the playlist given as argument.

I reused the command line argument playlist_name_ from the create_playlist option - I had to add it to the [De]Serialize functions to make it work, meaning -unless I missed something- that the create playlist could not be working before.
And this "create playlist" is still not working for me, but I did not dig further.

### DIFF
--- a/src/core/player.cpp
+++ b/src/core/player.cpp
@@ -224,6 +224,34 @@ void Player::NextItem(Engine::TrackChangeFlags change) {
   PlayAt(i, change, false);
 }
 
+void Player::PlayPlaylist(const QString& playlistName) {
+  PlayPlaylistInternal(Engine::Manual, playlistName);
+}
+
+void Player::PlayPlaylistInternal(Engine::TrackChangeFlags change,
+                                  const QString& playlistName) {
+  Playlist* playlist = NULL;
+  for (Playlist* p : app_->playlist_manager()->GetAllPlaylists()) {
+    if (playlistName == app_->playlist_manager()->GetPlaylistName(p->id())) {
+      playlist = p;
+      break;
+    }
+  }
+
+  if (playlist != NULL) {
+    app_->playlist_manager()->SetActivePlaylist(playlist->id());
+    if (playlist->rowCount() == 0) return;
+
+    int i = app_->playlist_manager()->active()->current_row();
+    if (i == -1) i = app_->playlist_manager()->active()->last_played_row();
+    if (i == -1) i = 0;
+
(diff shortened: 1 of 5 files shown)


src/core/player.cpp
```diff
@@ -224,6 +224,34 @@ void Player::NextItem(Engine::TrackChangeFlags change) {
   PlayAt(i, change, false);
 }
 
+void Player::PlayPlaylist(const QString& playlistName) {
+  PlayPlaylistInternal(Engine::Manual, playlistName);
+}
+
+void Player::PlayPlaylistInternal(Engine::TrackChangeFlags change,
+                                  const QString& playlistName) {
+  Playlist* playlist = NULL;
```
hatstand, line 233: Use `nullptr` instead of `NULL`.

src/core/player.cpp
```diff
@@ -224,6 +224,34 @@ void Player::NextItem(Engine::TrackChangeFlags change) {
...
+  for (Playlist* p : app_->playlist_manager()->GetAllPlaylists()) {
+    if (playlistName == app_->playlist_manager()->GetPlaylistName(p->id())) {
+      playlist = p;
+      break;
+    }
+  }
```
hatstand, line 239: You can probably write this `for` loop as a one liner with `std::find`

src/core/player.cpp
```diff
@@ -224,6 +224,34 @@ void Player::NextItem(Engine::TrackChangeFlags change) {
...
+
+  if (playlist != NULL) {
```
hatstand, line 241: Invert this logic so it returns early if no playlist was found.

src/ui/mainwindow.cpp
```diff
@@ -2270,6 +2271,14 @@ void MainWindow::CommandlineOptionsReceived(const CommandlineOptions& options) {
     case CommandlineOptions::Player_Next:
       app_->player()->Next();
       break;
+    case CommandlineOptions::Player_PlayPlaylist:
+      if (options.playlist_name().isNull() ||
+          options.playlist_name().isEmpty()) {
```
hatstand, line 2276: `isEmpty()` is sufficient.

src/core/player.cpp
NicolasToussaint: OK

src/core/player.cpp
NicolasToussaint: sorry, that's reaching out of my C++ skills :-/

PR #6826, opened 2020-11-02
 Allow Network Remote to browse the FileSystem from a defined root folder and then enqueue tracks from there

mbruel:
Hi,

I'm developing [ClementineRemote](https://github.com/mbruel/Clementine/).
You can find built version for Linux and MacOS [here](https://github.com/mbruel/ClementineRemote/releases).

The main goal for me was to add the possibility to browse the FileSystem remotely and thus to add Files to Playlists from there. (I don't use the Library and don't want to as many of my music has no metadata)
Another plus is that it will be available on any OS (working great on iPhone and laptops ;))

So as the Android Remote didn't provide that feature due to "security reasons", I've tried to make it secured...
I've added 2 new settings in the Network Remote. [cf snapshot here](https://github.com/mbruel/ClementineRemote/blob/main/pics/Clementine_Network_Remote_Settings.png).
So on the server we define a `files_root_folder` where the clients will be chrooted (cf the condition `rootDir.relativeFilePath(fiFolder.absoluteFilePath()).startsWith("../")`).
There is also `music_extensions` with a default list in `const char* Application::kDefaultMusicExtensionsAllowedRemotely = "aac,alac,flac,m3u,m4a,mp3,ogg,wav,wmv"` to limit the visibility on the clients.
And obviously the FileSystem is not stored remotely in any kind, it is requested on the fly, subfolder by subfolder.

To implement it and make it easier, I've created
 - 3 new protobuf types: `REQUEST_FILES`, `APPEND_FILES` and `LIST_FILES` 
 - with their corresponding messages: `RequestListFiles`, `ResponseListFiles` and `RequestAppendFiles`

Then their management is pretty straight forward in `IncomingDataParser` and `OutgoingDataCreator`
I reused the principle of `FileViewList::MimeDataFromSelection` to emit (forward) an `AddToPlaylistSignal(QMimeData* data)` signal connected to `MainWindow::AddToPlaylist(QMimeData* data)`

Let me know what you think about the changes.

I may have more changes to do for ClementineRemote, especially I'd like to have a Radio view (Internet -> Your radio streams) so we can change the radio from the Remote.

Could you please contact me to have an idea of when you're thinking to release the v1.4 and what you wish to include in it.
Here is my email: Matthieu.Bruel@gmail.com

Cheers! :)

Edit: I've also removed the use of `QBuffer` in `RemoteClient` as a `QByteArray` is more than enough

### DIFF
--- a/ext/libclementine-remote/remotecontrolmessages.proto
+++ b/ext/libclementine-remote/remotecontrolmessages.proto
@@ -87,23 +87,38 @@ enum MsgType {
 
 }
 
-message RequestFiles {
-  optional string relativePath = 1;
+message RequestListFiles {
+  optional string relative_path = 1;
 }
 
 message FileMetadata {
     optional string filename  = 1;
-    optional bool   isDir = 2;
+    optional bool   is_dir = 2;
 }
 
-// used by both Client (APPEND_FILES) and server (LIST_FILES)
-message ResponseFiles {
-  optional string relativePath = 1;
+// reply from server to LIST_FILES
+message ResponseListFiles {
+  enum Error {
+    NONE = 0;
+    ROOT_DIR_NOT_SET = 1;
+    DIR_NOT_ACCESSIBLE = 2;
+    DIR_NOT_EXIST = 3;
+    UNKNOWN = 4;
+  }
+  optional string relative_path = 1;
   repeated FileMetadata files  = 2;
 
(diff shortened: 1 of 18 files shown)


mbruel:
How can I have more detail on this `clang-format` failure?
I've checked the log but I don't see anything about the issue...
any help?

PR #6828, opened 2020-11-03
Allow Network Remote to browse the FileSystem from a defined root folder and then enqueue tracks from there

mbruel:
Hi,

I'm developing [ClementineRemote](https://github.com/mbruel/Clementine/).
You can find built version for Linux, MacOS and Windows 64bit [here](https://github.com/mbruel/ClementineRemote/releases).

The main goal for me was to add the possibility to browse the FileSystem remotely and thus to add Files to Playlists from there. (I don't use the Library and don't want to as many of my music has no metadata)
Another plus is that it will be available on any OS (working great on iPhone and laptops ;))

So as the Android Remote didn't provide that feature due to "security reasons", I've tried to make it secured...
I've added 2 new settings in the Network Remote. [cf snapshot here](https://github.com/mbruel/ClementineRemote/blob/main/pics/Clementine_Network_Remote_Settings.png).
So on the server we define a `files_root_folder` where the clients will be chrooted (cf the condition `rootDir.relativeFilePath(fiFolder.absoluteFilePath()).startsWith("../")`).
There is also `music_extensions` with a default list in `const char* Application::kDefaultMusicExtensionsAllowedRemotely = "aac,alac,flac,m3u,m4a,mp3,ogg,wav,wmv"` to limit the visibility on the clients.
And obviously the FileSystem is not stored remotely in any kind, it is requested on the fly, subfolder by subfolder.

To implement it and make it easier, I've created
 - 3 new protobuf types: `REQUEST_FILES`, `APPEND_FILES` and `LIST_FILES` 
 - with their corresponding messages: `RequestListFiles`, `ResponseListFiles` and `RequestAppendFiles`

Then their management is pretty straight forward in `IncomingDataParser` and `OutgoingDataCreator`
I reused the principle of `FileViewList::MimeDataFromSelection` to emit (forward) an `AddToPlaylistSignal(QMimeData* data)` signal connected to `MainWindow::AddToPlaylist(QMimeData* data)`

Let me know what you think about the changes.

I may have more changes to do for ClementineRemote, especially I'd like to have a Radio view (Internet -> Your radio streams) so we can change the radio from the Remote.

Could you please contact me to have an idea of when you're thinking to release the v1.4 and what you wish to include in it.
Here is my email: Matthieu.Bruel@gmail.com

Cheers! :)

Edit: I've also removed the use of `QBuffer` in `RemoteClient` as a `QByteArray` is more than enough

### DIFF
--- a/ext/libclementine-remote/remotecontrolmessages.proto
+++ b/ext/libclementine-remote/remotecontrolmessages.proto
@@ -205,9 +205,10 @@ message Shuffle {
 message ResponseClementineInfo {
   optional string version = 1;
   optional EngineState state = 2;
+  optional bool allow_downloads = 3;
 
   // allowed extensions for REQUEST_FILES and LIST_FILES
-  repeated string files_music_extensions = 3;
+  repeated string files_music_extensions = 4;
 }
 
 // The current song played
--- a/ext/libclementine-remote/remotecontrolmessages.proto
+++ b/ext/libclementine-remote/remotecontrolmessages.proto
@@ -87,23 +87,38 @@ enum MsgType {
 
 }
 
-message RequestFiles {
-  optional string relativePath = 1;
+message RequestListFiles {
+  optional string relative_path = 1;
 }
 
 message FileMetadata {
     optional string filename  = 1;
-    optional bool   isDir = 2;
+    optional bool   is_dir = 2;
 }
 
-// used by both Client (APPEND_FILES) and server (LIST_FILES)
-message ResponseFiles {
-  optional string relativePath = 1;
+// reply from server to LIST_FILES
+message ResponseListFiles {
+  enum Error {
+    NONE = 0;
+    ROOT_DIR_NOT_SET = 1;
+    DIR_NOT_ACCESSIBLE = 2;
+    DIR_NOT_EXIST = 3;
+    UNKNOWN = 4;
+  }
+  optional string relative_path = 1;
   repeated FileMetadata files  = 2;
 
-  optional string error = 3;
-  optional string musicExtentions = 4;
+  optional Error error = 3;
+  optional string music_extentions = 4;
 }
+
+// request to append files (APPEND_FILES)
+message RequestAppendFiles {
+  optional string relative_path = 1;
+  optional string new_playlist_name = 2;
+  repeated string files  = 3;
+}
+
 
 // Valid Engine states
 enum EngineState {
@@ -409,7 +424,8 @@ message Message {
   optional RequestDownloadSongs request_download_songs = 31;
   optional RequestRateSong request_rate_song = 35;
   optional RequestGlobalSearch request_global_search = 37;
-  optional RequestFiles request_files = 50;
+  optional RequestListFiles request_list_files = 50;
+  optional RequestAppendFiles request_append_files = 51;
 
   optional Repeat repeat = 13;
   optional Shuffle shuffle = 14;
@@ -430,5 +446,5 @@ message Message {
   optional ResponseGlobalSearch response_global_search = 38;
   optional ResponseTranscoderStatus response_transcoder_status = 39;
   optional ResponseGlobalSearchStatus response_global_search_status = 40;
-  optional ResponseFiles response_files = 51;
+  optional ResponseListFiles response_list_files = 52;
 }
--- a/ext/libclementine-remote/remotecontrolmessages.proto
+++ b/ext/libclementine-remote/remotecontrolmessages.proto
@@ -314,6 +314,13 @@ message RequestDownloadSongs {
   optional DownloadItem download_item = 1;
   optional int32 playlist_id = 2;
   repeated string urls = 3;
+
+  // within a Playlist, download only requested songs
+  repeated int32 songs_ids = 4;
+
+  // download from the FileSystem remotely
+  // using the defined root directory and the urls (filenames)
+  optional string relative_path = 5;
 }
 
 message ResponseSongFileChunk {
--- a/ext/libclementine-remote/remotecontrolmessages.proto
+++ b/ext/libclementine-remote/remotecontrolmessages.proto
@@ -394,7 +394,7 @@ message ResponseListFiles {
   repeated FileMetadata files = 2;
 
   optional Error error = 3;
-  optional string music_extentions = 4;  // set on server
+  optional string music_extensions = 4;  // set on server
 }
 
 message RequestAppendFiles {
--- a/ext/libclementine-remote/remotecontrolmessages.proto
+++ b/ext/libclementine-remote/remotecontrolmessages.proto
@@ -87,38 +87,6 @@ enum MsgType {
 
 }
 
-message RequestListFiles {
-  optional string relative_path = 1;
-}
-
-message FileMetadata {
-    optional string filename  = 1;
-    optional bool   is_dir = 2;
-}
-
-// reply from server to LIST_FILES
-message ResponseListFiles {
-  enum Error {
-    NONE = 0;
-    ROOT_DIR_NOT_SET = 1;
-    DIR_NOT_ACCESSIBLE = 2;
-    DIR_NOT_EXIST = 3;
-    UNKNOWN = 4;
-  }
-  optional string relative_path = 1;
-  repeated FileMetadata files  = 2;
-
-  optional Error error = 3;
-  optional string music_extentions = 4;
-}
-
-// request to append files (APPEND_FILES)
-message RequestAppendFiles {
-  optional string relative_path = 1;
-  optional string new_playlist_name = 2;
-  repeated string files  = 3;
-}
-
 
 // Valid Engine states
 enum EngineState {
@@ -405,6 +373,41 @@ message ResponseGlobalSearchStatus {
   optional string query = 2;
   optional GlobalSearchStatus status = 3;
 }
+
+
+// access the FileSystem remotely from a defined root directory
+message RequestListFiles {
+  optional string relative_path = 1;
+}
+
+message FileMetadata {
+    optional string filename  = 1;
+    optional bool   is_dir = 2;
+}
+
+message ResponseListFiles {
+  enum Error {
+    NONE = 0;
+    ROOT_DIR_NOT_SET = 1;
+    DIR_NOT_ACCESSIBLE = 2;
+    DIR_NOT_EXIST = 3;
+    UNKNOWN = 4;
+  }
+  optional string relative_path = 1;
+  repeated FileMetadata files  = 2;
+
+  optional Error error = 3;
+  optional string music_extentions = 4; // set on server
+}
+
+message RequestAppendFiles {
+  optional string relative_path = 1;
+  optional string new_playlist_name = 2;
+  repeated string files  = 3;
+  optional bool play_now = 4;
+  optional bool clear_first = 5;
+}
+
 
 // The message itself
 message Message {
--- a/ext/libclementine-remote/remotecontrolmessages.proto
+++ b/ext/libclementine-remote/remotecontrolmessages.proto
@@ -46,6 +46,9 @@ enum MsgType {
   GET_LIBRARY = 18;
   RATE_SONG = 19;
   GLOBAL_SEARCH = 100;
+  // access Files from remote control
+  REQUEST_FILES = 200;
+  APPEND_FILES = 201;
 
   // Messages send by both
   DISCONNECT = 2;
@@ -79,6 +82,27 @@ enum MsgType {
   GLOBAL_SEARCH_RESULT = 54;
   TRANSCODING_FILES = 55;
   GLOBAL_SEARCH_STATUS = 56;
+  // access Files from remote control
+  LIST_FILES = 202;
+
+}
+
+message RequestFiles {
+  optional string relativePath = 1;
+}
+
+message FileMetadata {
+    optional string filename  = 1;
+    optional bool   isDir = 2;
+}
+
+// used by both Client (APPEND_FILES) and server (LIST_FILES)
+message ResponseFiles {
+  optional string relativePath = 1;
+  repeated FileMetadata files  = 2;
+
+  optional string error = 3;
+  optional string musicExtentions = 4;
 }
 
 // Valid Engine states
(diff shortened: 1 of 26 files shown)


ext/libclementine-remote/remotecontrolmessages.proto
```diff
@@ -367,10 +372,44 @@ message ResponseGlobalSearchStatus {
...
+    ROOT_DIR_NOT_SET = 1;
+    DIR_NOT_ACCESSIBLE = 2;
+    DIR_NOT_EXIST = 3;
+    UNKNOWN = 4;
+  }
+  optional string relative_path = 1;
+  repeated FileMetadata files = 2;
+
+  optional Error error = 3;
+  optional string music_extentions = 4;  // set on server
```
hatstand, line 397: `music_extensions`

src/networkremote/incomingdataparser.cpp
```diff
@@ -370,3 +380,57 @@ Song IncomingDataParser::CreateSongFromProtobuf(
...
   return song;
 }
+
+void IncomingDataParser::AppendFilesToPlaylist(const pb::remote::Message& msg) {
+  if (files_root_folder_.isEmpty()) {  // should never happen...
+    qLog(Warning) << "Remote root dir is not set although receiving "
+                     "APPEND_FILES request...";
+    return;
+  }
+  QDir rootDir(files_root_folder_);
```
hatstand, line 390: `clang-format` should have caught this but it hasn't, all non-static variable names should be in snake_case. If in doubt, check the Google C++ style guide.

ext/libclementine-remote/remotecontrolmessages.proto
mbruel: what is the problem here?

src/networkremote/incomingdataparser.cpp
mbruel: hum ok, I'll correct that... and have  a look at Google C++ style guide.
since when Google dictate a standard for C++ style (and I guess other languages...) when I learned, the trend was camelCase...

src/networkremote/incomingdataparser.cpp
```diff
@@ -370,3 +380,58 @@ Song IncomingDataParser::CreateSongFromProtobuf(
...
+    qLog(Warning) << "Remote relative path " << relative_path
+                  << " is not a directory...";
+  else if (root_dir.relativeFilePath(fi_folder.absoluteFilePath())
+               .startsWith("../"))
+    qLog(Warning) << "Remote relative path " << relative_path
+                  << " should not be accessed...";
+  else {
+    QList<QUrl> urls;
+    QDir dir(fi_folder.absoluteFilePath());
+    for (auto it = req_append.files().cbegin(),
```
hatstand, line 414: I think you can use `for (const auto& file : req_append.files())`

src/networkremote/networkremote.cpp
```diff
@@ -213,6 +219,11 @@ void NetworkRemote::CreateRemoteClient(QTcpSocket* client_socket) {
     RemoteClient* client = new RemoteClient(app_, client_socket);
     clients_.push_back(client);
 
+    // Update the Remote Root Files for the latest Client
+    outgoing_data_creator_->SetMusicEextensions(client->music_extensions());
```
hatstand, line 223: `SetMusicExtensions`

src/networkremote/outgoingdatacreator.cpp
```diff
@@ -744,3 +745,59 @@ void OutgoingDataCreator::SearchFinished(int id) {
...
+}
+
+void OutgoingDataCreator::SendListFiles(QString relative_path) {
+  pb::remote::Message msg;
+  msg.set_type(pb::remote::LIST_FILES);
+
+  pb::remote::ResponseListFiles* files = msg.mutable_response_list_files();
+
+  if (files_root_folder_.isEmpty())
+    files->set_error(pb::remote::ResponseListFiles::ROOT_DIR_NOT_SET);
```
hatstand, line 765: You should return early for these error states to make the following code more readable and less indented.

src/networkremote/outgoingdatacreator.h
```diff
@@ -83,6 +87,8 @@ class OutgoingDataCreator : public QObject {
   void ResultsAvailable(int id, const SearchProvider::ResultList& results);
   void SearchFinished(int id);
 
+  void SendListFiles(QString relative_path);
```
hatstand, line 90: `const QString&`

src/networkremote/remoteclient.cpp
```diff
@@ -43,8 +40,11 @@ RemoteClient::RemoteClient(Application* app, QTcpSocket* client)
   use_auth_code_ = s.value("use_auth_code", false).toBool();
   auth_code_ = s.value("auth_code", 0).toInt();
   allow_downloads_ = s.value("allow_downloads", false).toBool();
-
+  files_root_folder_ = s.value("files_root_folder", "").toString();
+  music_extensions_ = s.value("music_extensions", "").toString();
```
hatstand, line 44: If this is really a `QStringList` I think you can store that directly in `QSettings`.

ext/libclementine-remote/remotecontrolmessages.proto
```diff
@@ -367,10 +372,44 @@ message ResponseGlobalSearchStatus {
...
+    ROOT_DIR_NOT_SET = 1;
+    DIR_NOT_ACCESSIBLE = 2;
+    DIR_NOT_EXIST = 3;
+    UNKNOWN = 4;
+  }
+  optional string relative_path = 1;
+  repeated FileMetadata files = 2;
+
+  optional Error error = 3;
+  optional string music_extensions = 4;  // set on server
```
hatstand, line 397: If this is really a list it should be `repeated string` rather than encoding that via commas.

src/ui/filechooserwidget.cpp
```diff
@@ -0,0 +1,107 @@
...
+      pathEdit_(new QLineEdit(this)),
+      mode_(mode),
+      filter_(""),
+      openDirPath_("") {
+  QLabel* lbl = new QLabel(label, this);
+  layout_->addWidget(lbl);
+  _init(initialPath);
+}
+
+void FileChooserWidget::setFileFilter(const QString& filter) {
```
hatstand, line 44: The style guide specifies Pascal case for method names.

src/ui/filechooserwidget.cpp
```diff
@@ -0,0 +1,107 @@
...
+  if (!fileInfo.exists()) return "";
+  if (mode_ == Mode::File) {
+    if (!fileInfo.isFile()) return "";
+  } else {
+    if (!fileInfo.isDir()) return "";
+  }
+  return path;
+}
+
+void FileChooserWidget::_init(const QString& initialPath) {
```
hatstand, line 68: `Init()`

src/ui/filechooserwidget.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include <QLineEdit>
+#include <QPushButton>
+FileChooserWidget::FileChooserWidget(QWidget* parent)
+    : QWidget(parent),
+      layout_(new QHBoxLayout(this)),
+      pathEdit_(new QLineEdit(this)),
+      mode_(Mode::Directory),
+      filter_(""),
+      openDirPath_("") {
+  _init("");
```
hatstand, line 16: `_init()` does real work here so should not be called from the constructor. If you really want to call it from the constructor you can probably use constructor chaining instead.

src/networkremote/incomingdataparser.cpp
mbruel: hum true, I gonna give it a try

src/networkremote/networkremote.cpp
mbruel: arf... it was late when I finished this part. sorry...

src/networkremote/outgoingdatacreator.cpp
mbruel: alright. I know some people that hate returning before the end but I kind of prefer it too

src/networkremote/outgoingdatacreator.h
mbruel: Indeed, the IncomingDataParser and the OutgoingDataCreator are living in the thread.
but the signal just above wasn't using cont ref.
```
    connect(incoming_data_parser_.get(),
            SIGNAL(DoGlobalSearch(QString, RemoteClient*)),
            outgoing_data_creator_.get(),
            SLOT(DoGlobalSearch(QString, RemoteClient*)));

    connect(incoming_data_parser_.get(), SIGNAL(SendListFiles(QString)),
            outgoing_data_creator_.get(), SLOT(SendListFiles(QString)));
```
I guess I'll update both ;)

src/networkremote/remoteclient.cpp
mbruel: it's easier indeed to have it as a QStringList at least at the bottom end for this test:
` if (fi.isDir() || music_extensions_.contains(fi.suffix()))`

I don't see the real benefit to store it as a QStringList in QSettings. In the GUI (network remote settings page) the User WILL define it as a coma separated list.. it sounds easier to me to just break it down in a List when needed.
i.e: in `OutgoingDataCreator::SetMusicEextensions(const QString& music_extensions)`
what you think?
I'll even add a new member: OutgoingDataCreator::music_extensions_str_ which will be an std::string to avoid doing this conversion: `files->set_music_extensions(music_extensions_.join(",").toStdString());`

ext/libclementine-remote/remotecontrolmessages.proto
mbruel: hum... I've given my opinion above. this field is not really important. I find it more compact to have a coma separated list.
faster to set server side. easy to break down client side.

src/ui/filechooserwidget.cpp
mbruel: I didn't finish reading the guide... bloody long. I strongly hate having Pascal case method names, but alright for this project.
I forgot about this FileChooserWidget. It's reused from another project of mine. I'll update the snake_case too

src/ui/filechooserwidget.cpp
mbruel: ok

src/ui/filechooserwidget.cpp
mbruel: `constructor chaining`, first time I hear that... I gonna have a look.
It's not real work, it's just there to factorize code of the 3 different constructors.
It is not really needed for this project but I don't see the benefit to narrow it down.

mbruel: following the remark from [this thread](https://stackoverflow.com/a/33275207) 

> However, there is a severe limitation that a constructor that calls another constructor is not allowed to initialize any other members. So you cannot do the following with a delegating constructor:

I find that constructor chaining is not a good idea. It is slowing down object default initialization.
If you don't mind, I prefer to keep an Init method (old way).

src/ui/filechooserwidget.cpp
hatstand: Thanks. It's not about personal preference, it's about consistency.

ext/libclementine-remote/remotecontrolmessages.proto
hatstand: The disadvantage there is that now every piece of code that touches it has to understand how it's encoded (is it comma-separated? is it null-separated, pipe-separated?, etc.).

mbruel: what about a comment in the proto declaration?
if a client would use it, it is normal to check the server encoding choice no?

src/ui/filechooserwidget.cpp
mbruel: sure, no pb, I completely agree

ext/libclementine-remote/remotecontrolmessages.proto
hatstand: If it's specified as `repeated string` then there's nothing for the client to check really. Either it successfully decodes as `String[]` or it doesn't.

mbruel: alright, then if you don't mind I'm going to move from `ResponseListFiles` to `ResponseClementineInfo`
there is no need to repeat this information.
Or would you prefer another type like `FILES_EXTENSIONS` with a dedicated response message `ResponseFilesExtensions` ?

src/ui/filechooserwidget.cpp
mbruel: happy with keeping an Init method?

ext/libclementine-remote/remotecontrolmessages.proto
hatstand: I think it makes sense in `ResponseClementineInfo`.

src/ui/filechooserwidget.cpp
hatstand: Sure

ext/libclementine-remote/remotecontrolmessages.proto
mbruel: ok cool, I may then rename it in `files_music_extensions`as it is only concerning the browsing of the FileSystem and not the Library for example.

src/networkremote/outgoingdatacreator.h
mbruel: hum in fact I'd rather keep it as a copy as I may remove the first slash
`if (relative_path.startsWith("/")) relative_path.remove(0, 1);`

src/networkremote/incomingdataparser.cpp
mbruel: done

src/networkremote/outgoingdatacreator.cpp
mbruel: done

src/networkremote/outgoingdatacreator.h
mbruel: so I've let it

mbruel:
we should be good no?
anything about a date for the new release or is it taboo?
I'd like to know if I should hurry and do all the addition I need in Clementine or if I can take my time and work on the remote...

mbruel:
I've made a little change that was needed for the Remote (as Clementine current tab is not in sync with the Remote one)

PR #6834, opened 2020-11-09
inform remote in ResponseClementineInfo if downloads are allowed

mbruel:
We could just inform the Remote when it connects if downloads are allowed or not.
This avoid to be disconnected if they're not allowed.
Instead we can just warn the user with a popup or disable the download buttons...


mbruel:
this has been merged in #6828 as it is also modifying ResponseClementineInfo

Issue #6855, opened 2020-12-05
clementine: QSortFilterProxyModelPrivate::proxy_to_source(): clementine killed by SIGSEGV


eclipseo:
Reported on Fedora: 

Description of problem:
- I am downloading a podcast episode.
- I right click on it while it is downloading and want to click on "Append to current playlist"
- If the download finishes before I manage to click on that context menu item (Append to current playlist), the podcast episodes tree colapses
- I then click on Append to current playlist and Clementine crashes

I tried this multiple times, always the same result. I assume when the pocasts episodes subtree collapses, the context menu doesn't have a correct reference to an episode anymore.

Version-Release number of selected component:
clementine-1.4.0-3.rc1.20200617gitedb8c3b.fc33.1

Additional info:
reporter:       libreport-2.14.0
backtrace_rating: 4
cmdline:        /usr/bin/clementine
crash_function: QSortFilterProxyModelPrivate::proxy_to_source
executable:     /usr/bin/clementine
journald_cursor: s=5f3ec159e24c47f4b50bac5b211274cb;i=6456a;b=077231dd8d324703a83d7a677491c22d;m=259c5fdf78;t=5b57c50331f9c;x=eaa08d6b10ba0c3c
kernel:         5.9.10-200.fc33.x86_64
rootdir:        /
runlevel:       N 5
type:           CCpp
uid:            1000

Truncated backtrace:
Thread no. 0 (10 frames)
 #0 QSortFilterProxyModelPrivate::proxy_to_source at itemmodels/qsortfilterproxymodel.cpp:546
 #1 QSortFilterProxyModel::mapToSource at itemmodels/qsortfilterproxymodel.cpp:3122
 #2 QSortFilterProxyModel::mimeData at itemmodels/qsortfilterproxymodel.cpp:2319
 #3 MergedProxyModel::mimeData at /usr/src/debug/clementine-1.4.0-3.rc1.20200617gitedb8c3b.fc33.1.x86_64/src/core/mergedproxymodel.cpp:432
 #4 InternetService::AddItemsToPlaylist at /usr/src/debug/clementine-1.4.0-3.rc1.20200617gitedb8c3b.fc33.1.x86_64/src/internet/core/internetservice.cpp:116
 #5 doActivate<false> at kernel/qobject.cpp:3898
 #6 QAction::triggered at .moc/moc_qaction.cpp:376
 #7 QAction::activate at kernel/qaction.cpp:1161
[...]

PR #6863, opened 2020-12-15
Clear cached indexes in InternetModel when rows are removed

jbroadus:
When opening a context menu on an internet item, the selected items are stored
in the InternetModel instance. In cases when the items are removed, certain menu
options can cause a crash. A specific case is downloading a podcast when the
user has chosen to limit the number of visible episodes. The subtree for the
podcast is rebuilt after the download completes, so if a context menu was opened
during the download time, selecting the append to playlist option will attempt
to operate on bad indexes.

This fix uses the rowsAboutToBeRemoved signal to remove these stored indexes.

There are likely another rare cases where the indexes can become invalid. For
example, sibling items within a subtree may be removed, causing the stored
indexes to become incorrect or out of range.

This is a fix for #6855

### DIFF
--- a/src/internet/core/internetmodel.cpp
+++ b/src/internet/core/internetmodel.cpp
@@ -81,6 +81,8 @@ InternetModel::InternetModel(Application* app, QObject* parent)
   Q_ASSERT(sServices->isEmpty());
 
   merged_model_->setSourceModel(this);
+  connect(merged_model_, SIGNAL(rowsAboutToBeRemoved(QModelIndex, int, int)),
+          SLOT(RowsAboutToBeRemoved(QModelIndex, int, int)));
 
   AddService(new ClassicalRadioService(app, this));
   AddService(new DigitallyImportedService(app, this));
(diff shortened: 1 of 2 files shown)


jbroadus:
#6864 should fix the build issue.

src/internet/core/internetmodel.cpp
```diff
@@ -234,6 +236,48 @@ int InternetModel::rowCount(const QModelIndex& parent) const {
...
+      qLog(Debug) << "Removing selected item";
+      iter = selected_indexes_.erase(iter);
+    } else {
+      iter++;
+    }
+  }
+}
+
+bool InternetModel::IsInLineage(QModelIndex d, const QModelIndex& a) {
+  while (!a.isValid()) {
```
hatstand, line 268: I think this can be an `if`?

jbroadus: Yeah. Copy/paste error.

jbroadus: Fixed

PR #6864, opened 2020-12-15
build: Remove eoan usage

jbroadus:
Ubuntu Eoan (19.10) is EOL and the repositories are no longer accessible. Bump
the image used for the source tarball build to LTS ubuntu:focal.


hatstand:
Thanks for  this.

PR #6869, opened 2020-12-17
Add FC33 build

hatstand:


jbroadus:
I was about to submit a PR for the same thing. There are a couple spec file changes that are needed. I can submit my request or just a request for the prerequisites.

jbroadus:
You've also discovered that the qt metapackages disappeared. These are the changed that I used to get it working:
https://github.com/clementine-player/Clementine/compare/master...jbroadus:add-fedora-33

PR #6878, opened 2020-12-29
Fix debug console instance leak

jbroadus:
Each time the debug console is launched, a new instance is created, but never
deleted. To fix, create one instance, if the option is enabled, and show that
one each time the menu option is selected.

### DIFF
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -172,6 +172,7 @@ MainWindow::MainWindow(Application* app, SystemTrayIcon* tray_icon, OSD* osd,
       edit_tag_dialog_(std::bind(&MainWindow::CreateEditTagDialog, this)),
       love_dialog_(std::bind(&MainWindow::CreateLoveDialog, this)),
       stream_discoverer_(std::bind(&MainWindow::CreateStreamDiscoverer, this)),
+      debug_console_(std::bind(&MainWindow::CreateDebugConsole, this)),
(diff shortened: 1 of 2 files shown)


src/ui/mainwindow.cpp
```diff
@@ -927,9 +928,7 @@ MainWindow::MainWindow(Application* app, SystemTrayIcon* tray_icon, OSD* osd,
           SLOT(EnableKittens(bool)));
   connect(ui_->action_kittens, SIGNAL(toggled(bool)), app_->network_remote(),
           SLOT(EnableKittens(bool)));
-  QString showConsole =
-      QProcessEnvironment::systemEnvironment().value(kShowDebugConsoleKey, "0");
-  if (showConsole == "1")
+  if (debug_console_.get())
```
hatstand, line 931: I think you want to keep the logic here as it was and allow `debug_console_->show()` inside `ShowConsole()` to implicitly create the `Console`.

jbroadus: Yeah. My ulterior motive for extending the lifetime of the Console instance is to support an interface where other components (specifically GetEngine) can add their own pages. If this occurred on initialization, the Lazy wrapper would just provide a mechanism to aid in initialization dependency resolution.

Maybe the better option is to emit a signal when a Console is actually created. That would still allow the Lazy construction to remain lazy.

jbroadus: Fixed.

Issue #6860, opened 2020-12-12
"Next track" and "Previous track" should work in "Repeat track" mode


Symbian9:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [x] I checked the issue tracker for similar issues
- [x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [ ] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Debian 10
- Clementine version: `1.4rc1`

### Expected behaviour / actual behaviour

If in *"Repeat mode > Repeat track"* user click <kbd>Next track</kbd>/<kbd>Previous</kbd> button (or shortcut) Clementine **should** switch to next/previous track in active playlist and play it in "Repeat track" mode.

Actually clicking on <kbd>Next track</kbd>/<kbd>Previous track</kbd> buttons Clementine just restart to play actual track.

maarten256:
I am not able to reproduce the exact effect described in this issue. However I did find a few use cases that produce what look to be incorrect behavior.

**Assuming that an album is present in the playlist:**
1. Start Clementine
2. Verify in "Repeat Track" mode
2. Select "Play" - selected track starts playing
3. Select "Next Track" twice (third track on the album is selected and playing)
4. Quit Clementine
5. Start Clementine (third track in the album is highlighted)
6. Select "Play"
7. First track on the album starts playing
8. Select "Next Track" twice (fourth track on the album is selected and playing)

**A playlist with random tracks**
1. Start Clementine
2. Create a playlist with a random selection of tracks from the library from different albums
3. Clementine will play the first track added to the playlist
4. Select "Next Track"
5. A seemingly random track on the list starts playing, not the next track in the playlist

PR #6887, opened 2021-01-10
Adding a "play next album" function, in addition to "play next track". Closes #2505.

### LINKED ISSUE #2505: Add a "Next Album" button when in Shuffle Album mode
_From [rkeevil@gmail.com](https://code.google.com/u/rkeevil@gmail.com/) on December 23, 2011 06:43:18_

Following on from issue #1130 and issue #1152 , it would be nice to have an extra button (perhaps only visible/enabled when Shuffle Album is selected) that skips to the next random album.

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=2505_

maarten256:
I've written a small change to clementine to allow the user to skip to the next album in a playlist. Similar functionality to that used to play the next track, but it will continue to skip forward until the album title changes.

### DIFF
--- a/src/core/player.cpp
+++ b/src/core/player.cpp
@@ -170,7 +170,9 @@ void Player::HandleLoadResult(const UrlHandler::LoadResult& result) {
 
 void Player::Next() { NextInternal(Engine::Manual); }
 
-void Player::NextInternal(Engine::TrackChangeFlags change) {
+void Player::NextAlbum() { NextInternal(Engine::Manual, true); }
+
+void Player::NextInternal(Engine::TrackChangeFlags change, bool next_album) {
   if (HandleStopAfter()) return;
 
   if (app_->playlist_manager()->active()->current_item()) {
@@ -186,10 +188,10 @@ void Player::NextInternal(Engine::TrackChangeFlags change) {
     }
   }
 
-  NextItem(change);
+  NextItem(change, next_album);
 }
 
-void Player::NextItem(Engine::TrackChangeFlags change) {
+void Player::NextItem(Engine::TrackChangeFlags change, bool next_album) {
   Playlist* active_playlist = app_->playlist_manager()->active();
 
   // If we received too many errors in auto change, with repeat enabled, we stop
@@ -213,7 +215,25 @@ void Player::NextItem(Engine::TrackChangeFlags change) {
   // Manual track changes override "Repeat track"
   const bool ignore_repeat_track = change & Engine::Manual;
 
-  int i = active_playlist->next_row(ignore_repeat_track);
+  int i = active_playlist->current_row();
+  if (next_album && i != -1) {
+    if (active_playlist->sequence()->repeat_mode() !=
+        PlaylistSequence::Repeat_Track) {
+      int original_index = i;
+      QString album = active_playlist->current_item_metadata().album();
+
+      i = active_playlist->next_row(ignore_repeat_track);
+      while (i != -1 &&
+             active_playlist->item_at(i)->Metadata().album() == album &&
+             i != original_index) {
+        active_playlist->set_current_row(i, true);
+        i = active_playlist->next_row(ignore_repeat_track);
+      }
+    }
+  } else {
+    i = active_playlist->next_row(ignore_repeat_track);
+  }
+
   if (i == -1) {
     app_->playlist_manager()->active()->set_current_row(i);
     emit PlaylistFinished();
--- a/src/core/player.cpp
+++ b/src/core/player.cpp
@@ -170,9 +170,12 @@ void Player::HandleLoadResult(const UrlHandler::LoadResult& result) {
 
 void Player::Next() { NextInternal(Engine::Manual); }
 
-void Player::NextAlbum() { NextInternal(Engine::Manual, true); }
+void Player::NextAlbum() {
+  NextInternal(Engine::Manual, NextTrackOrAlbumSelected::SelectNextAlbum);
+}
 
-void Player::NextInternal(Engine::TrackChangeFlags change, bool next_album) {
+void Player::NextInternal(Engine::TrackChangeFlags change,
+                          NextTrackOrAlbumSelected NextTrackOrAlbum) {
   if (HandleStopAfter()) return;
 
   if (app_->playlist_manager()->active()->current_item()) {
@@ -188,10 +191,11 @@ void Player::NextInternal(Engine::TrackChangeFlags change, bool next_album) {
     }
   }
 
-  NextItem(change, next_album);
+  NextItem(change, NextTrackOrAlbum);
 }
 
-void Player::NextItem(Engine::TrackChangeFlags change, bool next_album) {
+void Player::NextItem(Engine::TrackChangeFlags change,
+                      NextTrackOrAlbumSelected NextTrackOrAlbum) {
   Playlist* active_playlist = app_->playlist_manager()->active();
 
   // If we received too many errors in auto change, with repeat enabled, we stop
@@ -216,7 +220,7 @@ void Player::NextItem(Engine::TrackChangeFlags change, bool next_album) {
   const bool ignore_repeat_track = change & Engine::Manual;
 
   int i = active_playlist->current_row();
-  if (next_album && i != -1) {
+  if (NextTrackOrAlbum == SelectNextAlbum && i != -1) {
     if (active_playlist->sequence()->repeat_mode() !=
         PlaylistSequence::Repeat_Track) {
       int original_index = i;
--- a/src/core/player.h
+++ b/src/core/player.h
@@ -128,6 +128,8 @@ class Player : public PlayerInterface {
     PreviousBehaviour_Restart = 2
   };
 
+  enum NextTrackOrAlbumSelected { SelectNextTrack, SelectNextAlbum };
+
   void Init();
 
   EngineBase* engine() const { return engine_.get(); }
@@ -179,10 +181,13 @@ class Player : public PlayerInterface {
   void TrackEnded();
   // Play the next item on the playlist - disregarding radio stations like
   // last.fm that might have more tracks.
-  void NextItem(Engine::TrackChangeFlags change, bool next_album = false);
+  void NextItem(Engine::TrackChangeFlags change,
+                NextTrackOrAlbumSelected NextTrackOrAlbum = SelectNextTrack);
   void PreviousItem(Engine::TrackChangeFlags change);
 
-  void NextInternal(Engine::TrackChangeFlags, bool next_album = false);
+  void NextInternal(
+      Engine::TrackChangeFlags,
+      NextTrackOrAlbumSelected NextTrackOrAlbum = SelectNextTrack);
   void PlayPlaylistInternal(Engine::TrackChangeFlags,
                             const QString& playlistName);
 
--- a/src/core/player.h
+++ b/src/core/player.h
@@ -72,6 +72,7 @@ class PlayerInterface : public QObject {
 
   // Skips this track.  Might load more of the current radio station.
   virtual void Next() = 0;
+  virtual void NextAlbum() = 0;
 
   virtual void Previous() = 0;
   virtual void SetVolume(int value) = 0;
@@ -150,6 +151,7 @@ class Player : public PlayerInterface {
   void PlayPause();
   void RestartOrPrevious();
   void Next();
+  void NextAlbum();
   void Previous();
   void PlayPlaylist(const QString& playlistName);
   void SetVolume(int value);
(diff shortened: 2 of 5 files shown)


src/core/player.cpp
```diff
@@ -213,7 +215,19 @@ void Player::NextItem(Engine::TrackChangeFlags change) {
...
 
-  int i = active_playlist->next_row(ignore_repeat_track);
+  int i = active_playlist->current_row();
+  if (next_album && i != -1) {
+      QString album = active_playlist->current_item_metadata().album();
+
+      i = active_playlist->next_row(ignore_repeat_track);
+      while (active_playlist->item_at(i)->Metadata().album() == album) {
+          active_playlist->set_current_row(i,true);
+          i = active_playlist->next_row(ignore_repeat_track);
```
jbroadus: If you have a single album in your playlist, and you have the playlist to repeat, will this loop forever?

maarten256: Good test case - I tried it and it causes clementine to crash. I'll have to think of a fix for this.

jbroadus: Maybe Playlist could have a find_next_row (or FindNextRow.. there's a lot of inconsistency in method naming in there) that takes a functor that returns true if a match is found. Then you could just do something like:

```i = active_playlist->find_next_row([](PlaylistItemPtr& item) -> bool { return item->Metadata().album() != album});```

The method would return -1 if it looped through without finding a match.

Comment (maarten256) [reply src/core/player.cpp]: I found a shortcut to achieve the same behavior, "next_row()" return -1 when it tries to access the row beyond the last one on the list. I made that part of the while condition and tested it out. See my next commit.

File: src/core/player.h [hunk 2]
```diff
@@ -177,10 +179,10 @@ class Player : public PlayerInterface {
   void TrackEnded();
   // Play the next item on the playlist - disregarding radio stations like
   // last.fm that might have more tracks.
-  void NextItem(Engine::TrackChangeFlags change);
+  void NextItem(Engine::TrackChangeFlags change, bool next_album = false);
```
hatstand, line 182: Can you make `next_album` an enum for readability?

src/core/player.cpp
jbroadus: But I think it only returns -1 if you're not in playllist repeat mode. Otherwise, it loops (default case)
```
  if (next_virtual_index >= virtual_items_.count()) {
    // We've gone off the end of the playlist.

    switch (playlist_sequence_->repeat_mode()) {
      case PlaylistSequence::Repeat_Off:
      case PlaylistSequence::Repeat_Intro:
        return -1;
      case PlaylistSequence::Repeat_Track:
        next_virtual_index = current_virtual_index_;
        break;

      default:
        next_virtual_index = NextVirtualIndex(-1, ignore_repeat_track);
        break;
    }
```

Also note that NextVirtualIndex() check or album repeat mode.

maarten256: Thanks Jim. I'm going to rework the code per your original suggestion and create a new function to search for the next track rather than fiddling with the existing function.

src/core/player.h
maarten256: Per my conversation with Jim I'm going to create a new function that looks for the next album/track to play so this may become OBE.

maarten256:
I fixed the main issue with my original code (it would hang or crash depending on the setting of the repeat mode). I'm still considering to create a new nextAlbum function that independently searches for the next track to play.

src/core/player.cpp
```diff
@@ -213,7 +215,25 @@ void Player::NextItem(Engine::TrackChangeFlags change) {
   // Manual track changes override "Repeat track"
   const bool ignore_repeat_track = change & Engine::Manual;
 
-  int i = active_playlist->next_row(ignore_repeat_track);
+  int i = active_playlist->current_row();
+  int original_index = i;
+  if (next_album && i != -1) {
+    if (active_playlist->sequence()->repeat_mode() !=
+        PlaylistSequence::Repeat_Track) {
```
jbroadus, line 222: If you decide that it should be a no-op, then we should probably grey out the menu item.

There's also a Repeat_Album mode. I think it would loop back to your starting point there as well. I haven't looked at the behavior of the other repeat modes.

src/ui/mainwindow.ui
```diff
@@ -997,6 +1017,7 @@
  </customwidgets>
  <resources>
   <include location="../../data/data.qrc"/>
+  <include location="../../data/data.qrc"/>
```
jbroadus, line 1020: Not sure why it added a duplicate here.

jbroadus:
There's this semi-related bug that I noticed the other day: https://github.com/clementine-player/Clementine/issues/6860

src/core/player.cpp
```diff
@@ -213,7 +215,25 @@ void Player::NextItem(Engine::TrackChangeFlags change) {
   // Manual track changes override "Repeat track"
   const bool ignore_repeat_track = change & Engine::Manual;
 
-  int i = active_playlist->next_row(ignore_repeat_track);
+  int i = active_playlist->current_row();
+  int original_index = i;
```
jbroadus, line 219: You could move this into the scope of the second `if`.

src/core/player.cpp
maarten256: Here's the list of modes and their respective behaviors when there is only one album in the playlist: 
- Don't Repeat: Stops when it finds there are no further items to play;
- Repeat Track: Replays the same track;
- Repeat Album: Replays the same track;
- Repeat Playlist: Replays the same track;
- Stop after each Track: Stops immediately (same behavior as next track in this case)
- Intro Tracks: Stops when it finds there are no further items to play.

I'm not 100% sure this is exactly the expected behavior, but then again doing a "next album" on a playlist with only one album is a bit of an edge case.

As far as greying out the menu item is concerned, I'd have to look at tying that in with the selection of the "Repeat Track" option.

src/ui/mainwindow.ui
maarten256: To be fixed in the next commit.

src/core/player.cpp
maarten256: To be fixed in the next commit.

maarten256:
With respect to #6860, I can't reproduce the problem - I can't see how my code changes would've made a difference in that particular behavior. I'd have to test that out on a clean branch.

src/core/player.cpp
jbroadus: In the painfully long `MainWindow::PlaylistRightClick` method, we have access to the current playlist and we already call `setEnabled` on quite a few items, so that should be pretty straight forward.

jbroadus: I think Repeat Album would give you the same behavior if you had multiple albums in the playlist.

maarten256: Examining the code it looks like the selection of the Repeat Mode is handled in the PlaylistSequence object. It is not clear to me how this action can be tied back to changing the state of a component on the MainWindow. Do you have any suggestions?

And yes you're right - when Repeat Mode is Repeat Album, the behavior is the also same when there are multiple albums in the playlist.

jbroadus: In PlaylistRightClick, you should be able to access  `app_->playlist_manager()->current()->sequence()`. It looks like this is the same object as ui_->playlist_sequence, but it's probably best not to make that assumption. In fact, it would be nice if Playlist had a GetRepeatMode() accessor.

maarten256: I understand how to get to the PlaylistSequence from MainWindow, however what I am struggling with is that I'd need access to the MainWindow from PlaylistSequence and I don't have a good solution for that.

Changing the repeat mode triggers the following in PlaylistSequence:

void PlaylistSequence::RepeatActionTriggered(QAction* action) {}

I am assuming that a change to the repeat mode, where repeat mode becomes Repeat_Track, an update to the MainWindow needs to be made to deselect the next album option. Maybe I'm not looking at this right but I don't see or haven't found an entry point in MainWindow for a change of repeat mode?

maarten256: Ignore my previous post. I figured out how to handle the "repeat album" signal in MainWindow.cpp and have added a commit that disables the "next album" option when repeat mode is set (or changed) to "repeat track". I wonder if I shoud expand that to include "repeat album" as well?

jbroadus: You shouldn't need to handle the signal. In other cases, we enable/disable items when we show the menu. Again, take a look at PlaylistRightClick.

jbroadus: Regarding the repeat album mode, I think you would want to disable the menu item for that as well. But when/if we address that other issue that I mentioned, we probably want to follow the same behavior here. Both would require some plumbing in the playlist, though.

maarten256: > You shouldn't need to handle the signal. In other cases, we enable/disable items when we show the menu. Again, take a look at PlaylistRightClick.

I was of the impression that PlaylistRightClick shows a number of good examples of how to enable/disable items on the UI given specific circumstances. I tried to recreate that behavior in the new "SetNextAlbumEnabled" method. What is not clear to me is how I find out the Repeat Mode menu was accessed other than through a specific signal? This would then force a change of the availability of the Next Album function.

maarten256: > Regarding the repeat album mode, I think you would want to disable the menu item for that as well. But when/if we address that other issue that I mentioned, we probably want to follow the same behavior here. Both would require some plumbing in the playlist, though.

I get what you're saying - in Repeat Album mode, the Next Album action should still cause the next album to be selected. Agreed this would involve some more plumbing work, maybe for a future pull request? I tried to recreate #6860 without success so at this point I'm not sure what to do about that one...

jbroadus: > I get what you're saying - in Repeat Album mode, the Next Album action should still cause the next album to be selected. Agreed this would involve some more plumbing work, maybe for a future pull request? I tried to recreate #6860 without success so at this point I'm not sure what to do about that one...

Yeah, I wasn't suggesting that it should be fixed with this change.

jbroadus: > I was of the impression that PlaylistRightClick shows a number of good examples of how to enable/disable items on the UI given specific circumstances. I tried to recreate that behavior in the new "SetNextAlbumEnabled" method. What is not clear to me is how I find out the Repeat Mode menu was accessed other than through a specific signal? This would then force a change of the availability of the Next Album function.

It looks like we do set menu items in slots for the main controls. But in those cases, we also need to update main controls that are always visible.

A potential issue that I see with your new approach are that it assumes that the repeat behavior is global. In the future, if you broke that assumption, you would also have to handle the update when you changed current playlists. It could work, it just seems more complex.

jbroadus: Not sure I understand the question about the repeat mode menu. We're talking about the playlist menu, right?

maarten256: Now I think I understand where the confusion lies. I changed the behavior of the Next Track button to have a pulldown menu (like the Stop button):

Screenshot_2021-01-19_09-06-00

There is a submenu from that button:

screen_grab

So the action is not tied to the right-click menu but a new menu on the UI - hence the need to tie the functionality to a specific signal,

jbroadus: Oh, I see. Sorry, I missed that. That makes more sense. I think you could also use the `QMenu::aboutToShow` signal if you wanted to update on show.

src/ui/mainwindow.cpp
```diff
@@ -3096,3 +3117,15 @@ void MainWindow::keyPressEvent(QKeyEvent* event) {
     QMainWindow::keyPressEvent(event);
   }
 }
+
+// Change the state of action_next_album based when the repeat mode is
+// changed.
+void MainWindow::SetNextAlbumEnabled(PlaylistSequence::RepeatMode mode) {
+  if (mode == PlaylistSequence::Repeat_Track) {
+    if (ui_->action_next_album->isEnabled())
```
jbroadus, line 3125: Don't think you need to do this check.

maarten256: I put that in so the change to the action is only applied if action_next_album is enabled... So when the same selection is made again the action is not disabled again when it's already disabled.

src/core/player.cpp
maarten256: The RepeatModeChanged signal was already used in MainWindow.cpp so I tied the behavior to that.

jbroadus: Looks good to me. You probably want to squash some of the commits.

maarten256: I thought I squashed the commits so all the changes for this would be on a single commit but somehow it appears to have pulled in a bunch of other changes. It didn't change the number of files affected by this Pull Request though, so I guess it's ok? (Still a bit of a noob at git)

jbroadus: You definitely have 24 separate commits in this set.
If you haven't already, set up a remote for this repo. Example: `git remote add upstream git@github.com:clementine-player/Clementine.git` then `git fetch upstream`
You can see how many commits your are ahead of the upstream master with `git cherry -v upstream/master`
Rebase: `git rebase -i upstream/master`
Now if you run `git -v cherry upstream/master` again, you should just see the commits that haven't been squashed and `git diff origin/master` should show no code differences.
Now you can force push your branch.

You might want to branch a backup before you do this in case you make a mistake. Commits can be recovered, but it's easier if you have a reference. Also, in the future, you might want to work off of a task branch instead of your master so you can keep your master up to date with the the upstream.

jbroadus: Also, hatstand had a comment that you didn't address. And he's the one that pushes the merge button.

maarten256: Thanks for the hints! I ran through the commands you suggested and managed to properly squash the commits into one. Now the only thing left to do is to address hatstand's comment which I'd completely forgotten about...

src/core/player.h
maarten256: Eventually I decided to leave the code mostly as-is rather than writing a completely new method - so I applied your suggestion and committed it as a separate commit.

jbroadus:
My concerns have been addressed. cc @hatstand

PR #6891, opened 2021-01-11
transcoder: Fix flacenc "insane" level encoding

### LINKED ISSUE #6876: Cant rip as flac best quality

- Operating System: Linux Mint 20.1
- Clementine version:  1.4.0rc1-398-g0393d865c

To be able to rip FLACS
Rip cd choose flac
in options move quality up to best 
flacs will rip as 0 byte files. 
I am unsure if flacs work in other qualities

jbroadus:
The gstreamer flacenc element defines a set of quality levels that is different
from the standard flac library. Its highest level, labeled "insane", uses
settings that are outside of the streamable subset. Set the streamable-subset
property to false for this level.

Reference: https://xiph.org/flac/format.html#subset

Fixes #6876

### DIFF
--- a/src/transcoder/transcoderoptionsflac.cpp
+++ b/src/transcoder/transcoderoptionsflac.cpp
@@ -26,6 +26,8 @@ const char* TranscoderOptionsFlac::kSettingsGroup = "Transcoder/flacenc";
 TranscoderOptionsFlac::TranscoderOptionsFlac(QWidget* parent)
     : TranscoderOptionsInterface(parent), ui_(new Ui_TranscoderOptionsFlac) {
   ui_->setupUi(this);
+  connect(ui_->quality, SIGNAL(valueChanged(int)), this,
+          SLOT(ValueChanged(int)));
 }
 
 TranscoderOptionsFlac::~TranscoderOptionsFlac() { delete ui_; }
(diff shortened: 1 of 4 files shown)


src/transcoder/transcoderoptionsflac.cpp
```diff
@@ -37,9 +39,30 @@ void TranscoderOptionsFlac::Load() {
   ui_->quality->setValue(s.value("quality", 5).toInt());
 }
 
+void TranscoderOptionsFlac::ValueChanged(int value) {
+  if (!IsInStreamingSubset(value)) {
+    ui_->info->setStyleSheet("QLabel { color : red }");
+    ui_->info->setText(
+        tr("Warning: This compression level is outside of the streamable "
```
hatstand, line 46: You might want to expand on what the practical implications of that are.

src/transcoder/transcoderoptionsflac.cpp
```diff
@@ -37,9 +39,30 @@ void TranscoderOptionsFlac::Load() {
   ui_->quality->setValue(s.value("quality", 5).toInt());
 }
 
+void TranscoderOptionsFlac::ValueChanged(int value) {
+  if (!IsInStreamingSubset(value)) {
+    ui_->info->setStyleSheet("QLabel { color : red }");
```
hatstand, line 44: I think you need to unset this in the `else` block.

src/transcoder/transcoderoptionsflac.cpp
```diff
@@ -37,9 +39,30 @@ void TranscoderOptionsFlac::Load() {
...
+
 void TranscoderOptionsFlac::Save() {
   QSettings s;
   s.beginGroup(kSettingsGroup + settings_postfix_);
 
-  s.setValue("quality", ui_->quality->value());
+  int quality = ui_->quality->value();
+  s.setValue("quality", quality);
+
+  s.setValue("streamable-subset", IsInStreamingSubset(quality));
```
hatstand, line 60: Not sure why you're saving this?

src/transcoder/transcoderoptionsflac.cpp
```diff
@@ -37,9 +39,30 @@ void TranscoderOptionsFlac::Load() {
...
+}
+
+bool TranscoderOptionsFlac::IsInStreamingSubset(int level) {
+  // The gstreamer flacenc element defines its own quality settings. The
```
hatstand, line 64: A link to the FLAC docs would be useful here.

src/transcoder/transcoderoptionsflac.cpp
jbroadus: Added more info, but open to other suggestions.

src/transcoder/transcoderoptionsflac.cpp
jbroadus: Done.

src/transcoder/transcoderoptionsflac.cpp
jbroadus: Done.

src/transcoder/transcoderoptionsflac.cpp
jbroadus: The transcoder picks up the element properties from the settings. It iterates through each element property and checks for a matching setting entry.
https://github.com/clementine-player/Clementine/blob/master/src/transcoder/transcoder.cpp#L552

PR #6892, opened 2021-01-13
Give pipelines unique names

jbroadus:
Currently, all pipelines are named "pipeline". Use a type string in combination
with the unique stream id to give each pipeline a unique name.

### DIFF
--- a/src/engines/gstpipelinebase.h
+++ b/src/engines/gstpipelinebase.h
@@ -24,10 +24,10 @@
 
 class GstPipelineBase : public QObject {
  public:
-  GstPipelineBase();
+  GstPipelineBase(const QString& type);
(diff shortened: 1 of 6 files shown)


src/engines/gstpipelinebase.h
```diff
@@ -24,15 +24,27 @@
...
   GstElement* pipeline_;
+
+ private:
+  QString type_;
+
+  // Using == to compare two pipelines is a bad idea, because new ones often
+  // get created in the same address as old ones.  This ID will be unique for
+  // each pipeline.
+  static std::atomic<int> sId;
+  int id_;
```
hatstand, line 47: `const`

jbroadus: Fixed

Issue #6468, opened 2019-12-04
MP4 tag tmpo BPM


HorstFiedler:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [ ] I checked the issue tracker for similar issues
- [ ] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [ ] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System:
- Clementine version:

HorstFiedler:
The same problem is already reported, but still open (clementine Version 1.3.1-945-g440ac6dda) the Other apps using taglib are able to read mp4  tmpo - attribute, e.g. kid3 and AtomicParsley, in libclementine-tagreader/tagreader.cpp  the code to get BPM for mp4 is simply missing.
One more problem: BPM should be edittable, e.g appear under keywords, not as header tag within UI.

HorstFiedler:
Looking into rc 1.4: regarding tagreader: No changes at all, therefor above fix still working to set tmpo using e.g. kid3 and to display it with clementine, but **never edit metadata using clementine dialog, e.g to set score !** . That will destroy BPM setting.

HorstFiedler:
I solved my problem of loosing BPM when editting any other metadata but for other fields (e.g. "rate") the problem still exists. Please note that MP4 allows date as day, not year only, and that field is truncated to year only when editting metadata. But with that problem I can live with.  
[gitdiff.txt](https://github.com/clementine-player/Clementine/files/4093062/gitdiff.txt)

Summary: For me Clementine is an excellent amarok replacement and if userdefined scripts as Amarok has would be added to Clementine above problem would disappear as one could use external tag editor. At least a "Copy" menu item should be provided to get songs filename.

JulianVolodia:
Hello
Could you clean up here @HorstFiedler ?

Before doing below check if issue is quite recent, and maybe it was already fixed.

Great to have:
- summary *what to do* here when someone get this item to work with,
- fulfill issue again, even changing subject to have it clear with corespondence with current state of issue, leaving it in *great condition* in backlog for future work when someone have time.

Will appreciate if some important (like workarounds, recomendation for fixes, pre-investigation info) will be left with mark with ~shattering~ with `~this~` for fast reference, but best is to leave issue *most compressed as you could*.

Thanks in advance.

[...]

PR #6896, opened 2021-01-19
Regarding Issue #6468  (tagreader problems)

HorstFiedler:
When using Apple Mp4 tags the 'tmpo' property which normally is translated to BPM is not fetched from audio file metadata.
This change adds this feature without fixing other problems, e.g. when changing tags within clementine existing but unchanged tags become deranged. I strongly suggest to change taglib interface in a way kid3 is using (explicit adding/removing tags).
The minor change within utilities is just to  avoid searching for a file within huge direcories as simply showing directory without positioning or at least highlighting is unuseable.

### DIFF
--- a/ext/libclementine-tagreader/tagreader.cpp
+++ b/ext/libclementine-tagreader/tagreader.cpp
@@ -306,8 +306,7 @@ void TagReader::ReadFile(const QString& filename,
     }
 
     if (items.contains("BPM")) {
-      Decode(items["BPM"].toStringList().toString(", "), nullptr,
-             song->mutable_performer());
+      song->set_bpm(TStringToQString(items["BPM"].toString()).trimmed().toFloat());
     }
 
     if (items.contains("PERFORMER")) {
@@ -502,7 +501,12 @@ void TagReader::ReadFile(const QString& filename,
           song->set_score(score);
         }
       }
-
+      if (items.contains("tmpo")) {
+	float bpm = (float)items["tmpo"].toInt();
+	if (song->bpm() <= 0 && bpm > 0) {
+	  song->set_bpm(bpm);
+	}
+      }
       if (items.contains("\251wrt")) {
         Decode(items["\251wrt"].toStringList().toString(", "), nullptr,
                song->mutable_composer());
--- a/ext/libclementine-tagreader/tagreader.cpp
+++ b/ext/libclementine-tagreader/tagreader.cpp
@@ -51,6 +51,10 @@
 #include <textidentificationframe.h>
 #include <trueaudiofile.h>
 #include <tstring.h>
+
+#include <tpropertymap.h>
+#include <tfile.h>
+
 #include <unsynchronizedlyricsframe.h>
 #include <vorbisfile.h>
 #include <wavfile.h>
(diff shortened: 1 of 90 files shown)


ext/libclementine-tagreader/tagreader.cpp
```diff
@@ -502,7 +501,12 @@ void TagReader::ReadFile(const QString& filename,
           song->set_score(score);
         }
       }
-
+      if (items.contains("tmpo")) {
+	float bpm = (float)items["tmpo"].toInt();
+	if (song->bpm() <= 0 && bpm > 0) {
```
jbroadus, line 506: Could `has_bpm()` be used here?

jbroadus:
Minor: typo in commit message: `direcories`

ext/libclementine-tagreader/tagreader.cpp
HorstFiedler: tmpo and BPM are just translations. BPM origins from id3v2 , tmpo is the the equivalent apple/MP4 tagname. I dont think that translation can be used in member name (has_bpm()). As novice I cant answer that.

jbroadus: `song` is a protobuf object and bpm is an optional field, so `song->has_bpm()` would tell you if it has been previously set. But, again, if the bpm may have previously been set to 0, then this looks right.

HorstFiedler: The change worked for me, but as I said: Only to get BPM. For tagging I use kid3, ffmpeg, AtomicParsley, ... and all I wanted was to see BPM in clementine. For dance music BPM is essential and when amarok died I found clementine as a natural successor, Thank you for making me aware of "protobuf", "optional", ... I did not look into song  class. I only compared id3 to mp4 handling in tagreader and made a suggestion, the fix in mainstream code should be done by you as you are quite more familiar with that code.

jbroadus: I don't know, the code looks alright to me. I guess the lint checker doesn't run in /ext because it looks like there are some tabbing issue.

jbroadus:
@HorstFiedler This pull request is from your master branch. Did you intend to push a second commit to this request?

HorstFiedler:
Yes and no. The two commits are distinct, first one addressed getting bpm from mp4 file into clementine. That patched a visible fault in clementine. Second commit, which is  kind of draft addresses possibility to edit bpm (among other tags) for mp4 too. I'm using that version, but there are various points still open:
- using taglib's propertymap is fine for tag variants bpm(id3) <-> tmpo(mp4), one can avoid that cumbersumb code in clementine, mapping works for bpm but other tagnames to be mapped are not yet checked. I will try to do that. 
- clementines metadata editdialog is not yet checked for interferences
- tags which may be essential for other users (score, playcount) are not automatically added as before. Such attributes shall go to sqlite db but not into the file.  Those different views (playlist/ui - audiofile - database) to audio properties need some cleanup. To avoid to blow up song class which has already to much responsibilities I added kind of file view to playlist and the new tagreader interface avoids to use song.
- ... more to come

HorstFiedler:
Still open after last commit:
- editdialog items (lyrics, comment) to be included into nondestructive audio file tag save.
- various functions (e.g tagging ripped audio files tagging) not yet verified
- other tagreader functions should make use of taglibs translations instead of performing translations in clementine code.

HorstFiedler:
Due to too many irrelevant commits in attempt to satisfy lint clang and to remove unnecessessary qLog(Debug) lines the main reason for this pull request becomes invisible. Locally I will continue to use this fork. It works fine with mp4 tags but when storing id3 both frames (id3v1 and v2)  are written and I dont know how to set up taglib to write v2 only

jbroadus:
@HorstFiedler If you want to check the formatting locally, the command that I use is:
`clang-format --style="{BasedOnStyle: Google, DerivePointerBinding: false, Standard: Cpp11}" filename > filename.fixed`

jbroadus:
Again, I think your first commit was good and can be a stand-along MR if the formatting is cleaned up.

PR #6898, opened 2021-01-24
Add a pipeline list to the transcoder details dialog

jbroadus:
List current running transcode pipelines. This will be allow the addition of cancel and debug options. Specifically, I plan to add an option to dump the graph of a pipeline in an upcoming change.

### DIFF
--- a/src/engines/gstpipelinebase.h
+++ b/src/engines/gstpipelinebase.h
@@ -21,6 +21,7 @@
 #include <gst/gst.h>
 
 #include <QObject>
+#include <QStandardItemModel>
 
 class GstPipelineBase : public QObject {
  public:
@@ -45,6 +46,20 @@ class GstPipelineBase : public QObject {
   // each pipeline.
   static std::atomic<int> sId;
   const int id_;
+};
+
+class GstPipelineModel : public QStandardItemModel {
+  Q_OBJECT
+
+ public:
+  explicit GstPipelineModel(QObject* parent = nullptr);
+  void AddPipeline(int id, const QString& name);
+  void RemovePipeline(int id);
+
+ private:
+  int FindRowById(int id) const;
+
+  enum Role { Role_Id = Qt::UserRole + 1 };
 };
 
 #endif  // GSTPIPELINEBASE_H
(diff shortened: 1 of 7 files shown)


src/engines/gstpipelinebase.h
```diff
@@ -47,4 +48,18 @@ class GstPipelineBase : public QObject {
   const int id_;
 };
 
+class GstPipelineModel : public QStandardItemModel {
+  Q_OBJECT
+
+ public:
+  GstPipelineModel(QObject* parent = nullptr);
```
hatstand, line 55: `explicit`

src/engines/gstpipelinebase.h
```diff
@@ -47,4 +48,18 @@ class GstPipelineBase : public QObject {
...
+  void AddPipeline(int id, const QString& name);
+  void RemovePipeline(int id);
+
+ private:
+  int FindRowById(int id);
```
hatstand, line 60: I think this method can be `const`?

src/transcoder/transcodedialog.cpp
```diff
@@ -118,7 +119,7 @@ TranscodeDialog::TranscodeDialog(QWidget* parent)
 }
 
 TranscodeDialog::~TranscodeDialog() {
-  delete log_ui_;
+  delete details_ui_;
```
hatstand, line 122: Can you change these to use `unique_ptr` while you're here?

src/transcoder/transcoder.cpp
```diff
@@ -195,10 +195,16 @@ void Transcoder::JobState::PostFinished(bool success) {
                               new Transcoder::JobFinishedEvent(this, success));
 }
 
+QString Transcoder::JobState::GetDisplayName() {
+  return QFileInfo(job_.input).fileName() + " to " +
```
hatstand, line 199: If you use an arrow here, (`->` perhaps) then we don't need to worry about translations.

src/engines/gstpipelinebase.h
jbroadus: Changed.

src/transcoder/transcodedialog.cpp
jbroadus: Done.

src/transcoder/transcoder.cpp
jbroadus: Changed to use `=>`.

src/engines/gstpipelinebase.h
jbroadus: Yep. Fixed.

PR #6902, opened 2021-01-28
Don't show progress when nothing is selected

jbroadus:
Previously, pressing the "Start transcoding" button with zero files selected would show the progress group with the progress bar in the busy state.

Also removed slots annotation from Transcoder::Start() and Transcoder::Cancel() since they are always called directly.

### DIFF
--- a/src/transcoder/transcodedialog.cpp
+++ b/src/transcoder/transcodedialog.cpp
@@ -135,14 +135,21 @@ void TranscodeDialog::SetWorking(bool working) {
 }
 
 void TranscodeDialog::Start() {
+  QAbstractItemModel* file_model = ui_->files->model();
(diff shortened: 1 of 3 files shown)


src/transcoder/transcodedialog.cpp
```diff
@@ -135,14 +135,21 @@ void TranscodeDialog::SetWorking(bool working) {
 }
 
 void TranscodeDialog::Start() {
+  QAbstractItemModel* file_model = ui_->files->model();
+  int count = file_model->rowCount();
```
hatstand, line 139: `const`

jbroadus: Fixed

PR #6903, opened 2021-01-29
Regarding issue #6468 (tagreader problems)

HorstFiedler:
External tagreader adaptions
to enhance usage of playlist for tag editting including fix to get BPM from mp4 tag 'tmpo'.
Main change: Instead of posting whole song properties to external tagreader
only single tag (aka the one just editted) is posted. That required a new
tagreader interface and changes on whole path from ext to playlist.
As a side effect the tag translation which differs depending on
audio file type (mp3, aac, flac, ...) is delegated to propertymap in
3rdparty tagreader. On the other hand tagnames are made available to
playlist class (matching the columns enumeration) which should provide an
easier way to add additional filepersistence tags.
Sorry for huge commit list (some are simple reverts) but I'm new to git.

### DIFF
--- a/ext/libclementine-tagreader/tagreader.cpp
+++ b/ext/libclementine-tagreader/tagreader.cpp
@@ -892,6 +892,7 @@ bool TagReader::UpdateSongTag(const QString& filename,
   // according taglib tagwriter example
   // to save basic attributes see SaveFile
   
+  // no check for available ones
   TagLib::PropertyMap map = fileref->file()->properties();
 
   TagLib::String key = QStringToTaglibString(tagname);
@@ -901,6 +902,7 @@ bool TagReader::UpdateSongTag(const QString& filename,
   else
     // inserts too if not yet
     map.replace(key, QStringToTaglibString(tagvalue));
+
 
   //qLog(Debug) << TStringToQString(map.toString());
 
--- a/ext/libclementine-tagreader/tagreader.cpp
+++ b/ext/libclementine-tagreader/tagreader.cpp
@@ -51,6 +51,10 @@
 #include <textidentificationframe.h>
 #include <trueaudiofile.h>
 #include <tstring.h>
+
+#include <tpropertymap.h>
+#include <tfile.h>
+
 #include <unsynchronizedlyricsframe.h>
 #include <vorbisfile.h>
 #include <wavfile.h>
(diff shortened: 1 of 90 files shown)


jbroadus:
You mentioned that you're new to git, so here are some quick squash instructions that I posted in a different PR recently: https://github.com/clementine-player/Clementine/pull/6887#discussion_r561620265

src/playlist/playlist.cpp
```diff
@@ -98,6 +98,46 @@ const int Playlist::kUndoItemLimit = 500;
...
+                                                     << "playcount"
+                                                     << "skipcount"
+                                                     << "lastplayed"
+                                                     << "score"
+                                                     << "COMMENT"
+                                                     << "source"
+                                                     << "MOOD"
+                                                     << "performer"
+                                                     << "GROUPING"
+                                                     << "ORIGINALDATE";
```
jbroadus, line 139: A static GetColumnName method with a case statement is probably less error prone.

jbroadus: And if the strings are only for tags, maybe that method should be in TagReaderClient.

ext/libclementine-tagreader/tagreader.cpp
```diff
@@ -872,108 +843,68 @@ bool TagReader::SaveFile(const QString& filename,
...
-  // apart, so we keep specific behavior for some formats by adding another
-  // "else if" block above.
-  if (TagLib::Ogg::XiphComment* tag =
-          dynamic_cast<TagLib::Ogg::XiphComment*>(fileref->file()->tag())) {
-    SetVorbisComments(tag, song);
+#ifdef Q_OS_LINUX
+  if (ret) {
+    // Linux: inotify doesn't seem to notice the change to the file unless we
+    // change the timestamps as well. (this is what touch does)
+    utimensat(0, QFile::encodeName(filename).constData(), nullptr, 0);
```
jbroadus, line 867: Not sure why that would be. The Qt code that handles the inotify specifies the IN_MODIFY flag. And the modification of the file would be updating the timestamp as well. Not sure what games taglib would be playing to circumvent this.

HorstFiedler: It's just a copy from former code. I would like to remove that.

src/playlist/playlist.cpp
HorstFiedler: QStringList Song::kColumns  was taken as example. QStringList might cause more runtime overhead but readability is much better to name enumeration items.

PR #6905, opened 2021-01-31
Provide user feedback in options error condition 

jbroadus:
Remove condition that allows no dialog to be displayed when user selects
options for transcoding. Display the dialog with an error message
instead. This also applies to the transcode options dialog in the
ripping and network remote settings.

This should not occur in current scenarios, but it's groundwork another fix.

### DIFF
--- a/src/transcoder/transcoderoptionsdialog.cpp
+++ b/src/transcoder/transcoderoptionsdialog.cpp
@@ -56,11 +56,15 @@ TranscoderOptionsDialog::TranscoderOptionsDialog(const TranscoderPreset& preset,
       options_ = new TranscoderOptionsWma(this);
       break;
     default:
+      ui_->errorMessage->setText(
+          tr("Unknown encoder type: %1").arg(preset.name_));
       break;
   }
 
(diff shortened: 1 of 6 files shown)


src/transcoder/transcoderoptionsdialog.cpp
```diff
@@ -55,11 +56,14 @@ TranscoderOptionsDialog::TranscoderOptionsDialog(Song::FileType type,
...
     default:
+      ui_->errorMessage->setText(
+          tr("Unknown encoder type: %1").arg(preset.name_));
       break;
   }
 
   if (options_) {
-    setWindowTitle(windowTitle() + " - " + Song::TextForFiletype(type));
+    ui_->errorMessage->setVisible(false);
+    setWindowTitle(windowTitle() + " - " + Song::TextForFiletype(preset.type_));
```
hatstand, line 66: We should fix the translation of this, ideally.

jbroadus: Fixed. I think this is what you had in mind?

PR #6907, opened 2021-02-01
utilities: Add common method for bug report URL generation

jbroadus:
Move code from PodcastParser::ParseItem for reuse.

Example:
Calling Utilities::MakeBugReportUrl("New bug") returns:
https://github.com/clementine-player/Clementine/issues/new?title=New%20bug

### DIFF
--- a/src/core/utilities.cpp
+++ b/src/core/utilities.cpp
@@ -40,6 +40,7 @@
 #include <QTcpServer>
 #include <QTemporaryFile>
 #include <QUrl>
+#include <QUrlQuery>
 #include <QWidget>
 #include <QXmlStreamReader>
 #include <QtDebug>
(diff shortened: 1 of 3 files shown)


src/core/utilities.cpp
```diff
@@ -769,6 +769,15 @@ QString ScrubUrlQueries(const QString& str) {
   return QString(str).replace(rx, "\\1 (query removed)");
 }
 
+QString MakeBugReportUrl(const QString& title) {
+  // Example:
+  // https://github.com/clementine-player/Clementine/issues/new?title=New%20bug
+  return QString(
+             "https://github.com/clementine-player/Clementine/"
+             "issues/new?title=") +
+         QUrl::toPercentEncoding(title);
```
hatstand, line 778: Can we use the `QUrl` helpers for query items here instead?

jbroadus: Done

hatstand (approved):

Issue #5356, opened 2016-04-21
Symbolic link handling in paths is inconsistent.


Ferroin:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [x] I checked the issue tracker for similar issues
- [x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [x] I tried the latest Clementine build from [here](https://builds.clementine-player.org/) Tested a local build from the upstream git tree, built on and for 64-bit Linux

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Any Linux system, possibly also on OS X
- Clementine version:  At least 1.2.3 and newer, possibly earlier versions.

### Expected behaviour / actual behaviour

Currently, Clementine is inconsistent in it's handling of symbolic links in paths.  When adding a path to the Library, Clementine will canonicalize the path (it's internally doing the equivalent of `realpath`), thus eliminating all symbolic links in it.  However, when adding songs to a playlist from the file tab, or by using the command line, or by drag-and-drop from another application, it does not resolve symbolic links.  The inconsistency can cause significant confusion, as well as undesirable side effects (see #5351 for example).

At an absolute minimum, this handling needs to be made consistent (which would avoid things like #5351).

ideally, we shouldn't be canonicalizing the paths that are added to the Library, as this breaks one of the more common use-cases for symbolic links (redirecting things such that you can trivially change the redirection in one place, instead of having to go through numerous applications to do so), and we should just be trusting the VFS layer to resolve things for us properly.

logological:
I think this is a duplicate of Issue #2299 (going back to Clementine 1.0.1).

chrisgraham:
Same issue applies to imported .m3u files. Even though an .m3u is referencing something in the library, using a relative path, it imports as a new file using an absolute path (realpath). This means ratings do not show and can not be set.

chrisgraham:
I should also note that if you try and workaround the issue adding the realpath roots to the library in Preferences, you'll just end up with duplication. This should not happen either.

chrisgraham:
A working workaround is to use bind filesystem instead of symlinks.

braoult:
> A working workaround is to use bind filesystem instead of symlinks.

Except that everybody can make a symbolic link, not a mount. And when source dir is a network automount, a bind is not desirable, preventing automount to do its job (I think).

smithjd15:
[...]

Issue #5065, opened 2015-10-08
Rating not possible in playlist


Lizard09:
Hello,

Since migration to 1.2.3-1359-g86ce06e it's not possible to rate a song in a playlist when this one is just loaded from disk.

How to reproduce : 
1/ Create a playlist
2/ Add some songs in it
3/ Save & close the playlist
4/ Reload the saved playlist
5/ Impossible to rate the playlst's songs

Note : if you add again the song manually, it's possible to rate it again.

Cheers,
Florian

narunlifescience:
unable to reproduce this issue..

Chocobozzz:
@Lizard09 Can you provide us some logs?

mohicanb:
Hi, I have an issue that look like yours.

System: Ubuntu 14.04
Clementine version 1.2 (from Ubuntu repositories) and 1.3 (from ppa)

**For some files, when adding them from the disk files**, rating is not available. I also observed that in this case **automatic tags such as last played date, number of listening and score, do not update**.

Sometimes this only occurs when adding with the context menu, sometimes it occurs on the opposite when adding with drag and drop, sometimes it occurs with both methods.

For the moment I have not been able to determine what differentiates files for which it happens.
It doesn't look to be related to file format.
It is not related to disk partition format nor to a specific partition.

mohicanb:
I figured out what the problem I found is and described it in issue #5351.
It may - or not - be the same problem.

Issue #2299, opened 2013-12-10
When adding music files to the library, symbolic links are being resolved, 


Clementine-Issue-Importer:
_From [mattal...@gmail.com](https://code.google.com/u/117263509933843417938/) on October 15, 2011 19:51:01_

What steps will reproduce the problem? 1. Create a symbolic link to another directory.
2. In Clementine, go to Preferences > Music Library.
3. Click "Add new folder ... " and choose the symbolic link.
4. Press "Open". What is the expected output? What do you see instead? I expected to see the symbolic link in the list of directories, but I actually saw the symbolic link's target directory. What version of the product are you using? On what operating system? Ubuntu 11.04, 32-bit Please provide any additional information below. I understand that for some, this might be a feature and not a bug. 

I always have symbolic links in my home directory (always at the same path) that link to various resources on my network (not always at the same place). For example, the path to my files is always "~/files". Right now I have my file storage drive mounted on my workstation, so ~/files is a symbolic link that points to "/mnt/files". But last week, the files partition was only available to this workstation as a NFS share, so then, ~/files linked to my autofs dir: /auto/nfs/fileserver/filesshare. 

So you see, I want Clementine to consider ~/files/music its library. I do not want it to resolve the link and use /mnt/files/music instead, because if I change the target of ~/files, Clementine will have a bunch of broken links.

_Original issue: http://code.google.com/p/clementine-player/issues/detail?id=2299_

dev-jam:
Workaround for this is to mount bind the nfs folder to your target folder in home.
It seems saving the playlist with relative paths, results in using the original (nfs) folder for all files. When saving with absolute paths, it depends on the way they were added. Files from added from the media library showed the nfs path, files added by a file manager from the home folder showed the home path.

logological:
Still reproducible in Clementine 1.3.1.

braoult:
As 7 years passed without fix, maybe a dirty workaround (1.3.1, XUbuntu 18.04) could be the following.

My library configuration:
* ~/Music/iTunes -> /mnt/music/iTunes/Music
* /mnt/music -> cifs/music/daily-01 (/mnt/cifs/X being autofs mounts - something like share "X" on server "X")

Clementine behavior:
* clementine library chosen path in UI: ~/music/iTunes
* clementine resolved path: /mnt/cifs/music/daily-01/iTunes/Music

I directly changed the path in database,  and did not notice any side effect (yet). **Please see [*] below**.
[...]

PR #6919, opened 2021-02-12
Added library paths should be absolute paths, not a canonical representation

smithjd15:
Canonical path conversion resolves symlinks, which breaks saving of (portable) relative path playlists. Canonical path conversion also resolves dot-segments, although the library add path dialog seems to avoid this anyway.

### DIFF
--- a/src/library/librarybackend.cpp
+++ b/src/library/librarybackend.cpp
@@ -218,11 +218,11 @@ void LibraryBackend::UpdateTotalSongCount() {
 }
 
 void LibraryBackend::AddDirectory(const QString& path) {
-  QString canonical_path = QFileInfo(path).canonicalFilePath();
-  QString db_path = canonical_path;
+  QString absolute_path = QFileInfo(path).absoluteFilePath();
+  QString db_path = absolute_path;
 
   if (Application::kIsPortable && Utilities::UrlOnSameDriveAsClementine(
-                                      QUrl::fromLocalFile(canonical_path))) {
+                                      QUrl::fromLocalFile(absolute_path))) {
     db_path = Utilities::GetRelativePathToClementineBin(db_path);
     qLog(Debug) << "db_path" << db_path;
   }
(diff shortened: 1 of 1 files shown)


src/library/librarybackend.cpp
```diff
@@ -218,11 +218,11 @@ void LibraryBackend::UpdateTotalSongCount() {
...
 
 void LibraryBackend::AddDirectory(const QString& path) {
-  QString canonical_path = QFileInfo(path).canonicalFilePath();
-  QString db_path = canonical_path;
+  QString absolute_path = QFileInfo(path).absoluteFilePath();
+  QString db_path = absolute_path;
 
   if (Application::kIsPortable && Utilities::UrlOnSameDriveAsClementine(
-                                      QUrl::fromLocalFile(canonical_path))) {
+                                      QUrl::fromLocalFile(absolute_path))) {
```
jbroadus, line 225: Shouldn't we resolve symlinks for this check?

smithjd15: This will only return false on Windows.

IMO, Unix or otherwise, symlinks are not of implicit occurance, and the user ought to already have a reasonable awareness of consequence. Thus, Clementine should only malfunction when the user (or the OS) does, and have no capability for or awareness beyond that.

smithjd15:
Replace all remaining canonical path resolves with absolute equivalents.

smithjd15:
Closes https://github.com/clementine-player/Clementine/issues/5356.

smithjd15:
Closes https://github.com/clementine-player/Clementine/issues/2299.

src/library/librarybackend.cpp
smithjd15: Resolve this only for Windows within UrlOnSameDriveAsClementine().

smithjd15:
Closes https://github.com/clementine-player/Clementine/issues/5065.

PR #6927, opened 2021-02-20
protobuf: Fix namespace conflict

### LINKED ISSUE #6926: Use of pb namespace incompatible with protobuf 3.15.0
The new protobuf 3.15.0 seems to add a new `pb` namespace alias that breaks compilation of clementine. Replacing the `pb` namespace with `cpb` allows clementine to compile. A sample patch is available at https://github.com/ahesford/void-packages/blob/protobuf/srcpkgs/clementine/patches/pb_namespace.patch

The patch applies cleanly to 1.4.0rc1 but I am not sure about the head of `master`.

jbroadus:
Protobuf 3.15 adds a namespace alias for "pb" that conflicts with
Clementine's. Modify Clementine to use "cpb".

Patch provided by @ahesford

Reference: https://github.com/protocolbuffers/protobuf/blob/5c028d6cf42e426d47f5baa6ea3f0f5c86b97beb/src/google/protobuf/port.h#L44

Fixes #6926

### DIFF
--- a/ext/clementine-spotifyblob/spotifyclient.cpp
+++ b/ext/clementine-spotifyblob/spotifyclient.cpp
@@ -40,7 +40,7 @@ const int SpotifyClient::kSpotifyImageIDSize = 20;
 const int SpotifyClient::kWaveHeaderSize = 44;
 
 SpotifyClient::SpotifyClient(QObject* parent)
-    : AbstractMessageHandler<pb::spotify::Message>(nullptr, parent),
+    : AbstractMessageHandler<cpb::spotify::Message>(nullptr, parent),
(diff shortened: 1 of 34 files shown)


jbroadus:
@mbruel  Will this require changes to the remote as well?

hatstand:
> @mbruel Will this require changes to the remote as well?

I don't think it will as that has its own copy of the protos and specifies `java_package` anyway.

jbroadus:
Fixed the formatting issues.

mbruel:
@jbroadus what's the conflict exactly?
The `pb::remote` namespace is still available right?

PR #6929, opened 2021-02-21
oauth: Fix redirect

jbroadus:
Query params were not added to redirect url prior to inclusion in oauth request. This change fixes OneDrive authentication.

Regression introduced in 8b226c2171


jbroadus:
This does not fix onedrive integration, just the initial auth.

PR #6931, opened 2021-02-22
Fix onedrive

jbroadus:
The live API was deprecated in 2018. This change implements basic
onedrive access using the MS graph API.

The URL scheme was also changed from skydrive to onedrive. This is based
on the assumption that existing playlists won't have compatible item
ids.

Known issues:
- Directories with over 200 items will be truncated.
- No mechanism for discovering changes at runtime.
- No mechanism for removing deleted items or rescanning.

Reference: https://docs.microsoft.com/en-us/onedrive/developer/rest-api/concepts/migrating-from-live-sdk?view=odsp-graph-online

### DIFF
--- a/ext/libclementine-tagreader/cloudstream.cpp
+++ b/ext/libclementine-tagreader/cloudstream.cpp
@@ -113,6 +113,8 @@ TagLib::ByteVector CloudStream::readBlock(ulong length) {
                        QString("bytes=%1-%2").arg(start).arg(end).toUtf8());
   request.setAttribute(QNetworkRequest::CacheLoadControlAttribute,
                        QNetworkRequest::AlwaysNetwork);
+  // TODO: Use RedirectPolicyAttribute after baseline moves to Qt 5.9.
+  request.setAttribute(QNetworkRequest::FollowRedirectsAttribute, true);
 
   QNetworkReply* reply = network_->get(request);
   connect(reply, SIGNAL(sslErrors(QList<QSslError>)),
--- a/src/internet/core/oauthenticator.cpp
+++ b/src/internet/core/oauthenticator.cpp
@@ -102,15 +102,17 @@ QByteArray OAuthenticator::ParseHttpRequest(const QByteArray& request) const {
 }
 
 void OAuthenticator::RequestAccessToken(const QByteArray& code,
-                                        const QUrl& url) {
+                                        const QUrl& redirect_url) {
   typedef QPair<QString, QString> Param;
   QList<Param> parameters;
   parameters << Param("code", code) << Param("client_id", client_id_)
-             << Param("client_secret", client_secret_)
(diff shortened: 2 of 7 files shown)


jbroadus:
@hatstand to get onedrive to work again, we're going to have to migrate to the Microsoft graph API.

It doesn't seem to bad, but it looks like we need a new device ID:
https://docs.microsoft.com/en-us/graph/auth-register-app-v2

hatstand:
I registered a native app (i.e., no client secret) with these details:
Client ID: `905def38-34d2-4e32-8ba7-c37bcc329047`
Redirect URI: https://clementine-data.appspot.com/skydrive (although maybe it should use the native flow via https://login.microsoftonline.com/common/oauth2/nativeclient ?)

I gave it these permissions:
```
Files.Read
Files.Read.All
User.Read
```

jbroadus:
I'm unclear how `nativeclient` works. It seems like it requires the application to own the browser to get the auth code seamlessly back to the app.

jbroadus:
@hatstand As noted, I also took the opportunity to change the URL scheme from "skydrive" to "onedrive". I'm assuming that old libraries won't have matching IDs, though I don't have any data to check that. Also, the feature has been broken for a while now. I can switch back to "skydrive" if you prefer.

hatstand:
> I'm unclear how `nativeclient` works. It seems like it requires the application to own the browser to get the auth code seamlessly back to the app.

Yeah, the docs aren't very clear here but it would be nice to use the proper native app flow here and avoid using a web view too.


> @hatstand As noted, I also took the opportunity to change the URL scheme from "skydrive" to "onedrive". I'm assuming that old libraries won't have matching IDs, though I don't have any data to check that. Also, the feature has been broken for a while now. I can switch back to "skydrive" if you prefer.

I think that's fine.

jbroadus:
Okay, this document is clearer: 
https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow

```
For native & mobile apps, you should use one of the recommended values - https://login.microsoftonline.com/common/oauth2/nativeclient (for apps using embedded browsers) or http://localhost (for apps that use system browsers).
```

jbroadus:
I created a test app and was able to use the localhost method. If you add `http://localhost/` to your redirect list, we can use that.

hatstand:
> I created a test app and was able to use the localhost method. If you add `http://localhost` to your redirect list, we can use that.

Done for `905def38-34d2-4e32-8ba7-c37bcc329047`

jbroadus:
> > I created a test app and was able to use the localhost method. If you add `http://localhost` to your redirect list, we can use that.
> 
> Done for `905def38-34d2-4e32-8ba7-c37bcc329047`

Yep, it's working. I updated the change to use the localhost method.

PR #6942, opened 2021-03-06
build: Don't reset cflags in moodbar cmake

jbroadus:
Append CMAKE_CXX_FLAGS and CMAKE_C_FLAGS instead of setting. Setting the
variables causes breakage in cross-compiled systems where sysroots and
machine types are set.


hatstand (approved):

jbroadus:
I'm not sure why mac builds started failing in qsqlite, but maybe it's time to revisit #6564 and make it conditional.

jbroadus:
> I'm not sure why mac builds started failing in qsqlite, but maybe it's time to revisit #6564 and make it conditional.

Rebased on the fix.

PR #6943, opened 2021-03-07
Use qt sqlite3 driver by default

jbroadus:
The 3rdParty qsqlite plugin was kept to insure that fts3 support was
included. In qt5, the option was added by default, so it's unlikely
that there are distros that lack the support. A build option,
USE_SYSTEM_SQLITE_DRIVER, can be disabled to switch back to the 3rd
party driver.

Reference: https://github.com/qt/qtbase/commit/ee69c935c0f7b4ad3747e369fc18373d31e2521b

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -366,14 +366,20 @@ if(HAVE_VISUALISATIONS)
   endif(USE_SYSTEM_PROJECTM)
 endif(HAVE_VISUALISATIONS)
 
-
-# Build our copy of QSqlLiteDriver.
-# We do this because we can't guarantee that the driver shipped with Qt exposes the
-# raw sqlite3_ functions required for FTS support. This way we know that those symbols
-# exist at compile-time and that our code links to the same sqlite library as the
-# Qt driver.
-add_subdirectory(3rdparty/qsqlite)
-include_directories("3rdparty/qsqlite")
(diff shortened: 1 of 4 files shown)


jbroadus:
Event though this wasn't the root of the problem, we might want to consider taking this change. Now we know that some of these 3rd party libraries will break since qt6 requires c++17. (https://www.qt.io/blog/2019/08/07/technical-vision-qt-6). That may just be a matter of changing the CXX flags in the makefiles, or there may be bigger pain points.

hatstand:
> Event though this wasn't the root of the problem, we might want to consider taking this change. Now we know that some of these 3rd party libraries will break since qt6 requires c++17. (https://www.qt.io/blog/2019/08/07/technical-vision-qt-6). That may just be a matter of changing the CXX flags in the makefiles, or there may be bigger pain points.

Typically, the limitation preventing us from using newer C++ variants has been distro/OS compiler support. The lagging compiler here is probably whatever xenial or debian stable shipped with.

jbroadus:
> Typically, the limitation preventing us from using newer C++ variants has been distro/OS compiler support. The lagging compiler here is probably whatever xenial or debian stable shipped with.

Okay. Then I'll just keep this branch around in case we need it in the future. I assume these deprecation messages that we're seeing right now are going to mean something in Qt6.

Issue #6950, opened 2021-03-15
Track informations/tags not editable since build 340


liofeu:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [ x] I checked the issue tracker for similar issues
- [x ] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [ x] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Mac OSX 10.15.7 (Catalina)
- Clementine version: 1.4.0rc1-498-g681e7bea5

### Expected behaviour / actual behaviour

Since build 340, any track information/tag not editable. It remains stuck on "loading tracks..." when opening the track information window.

On build 332 and before, no problem at all.

So the issue should come from somewhere there https://github.com/clementine-player/Clementine/compare/1.4.0rc1-332-g62d2f0de9...1.4.0rc1-340-g2172732b1 

### Steps to reproduce the problem (only for bugs)

Right click on a track > Edit track Information

It remains stuck on "loading tracks..."

jbroadus:
I don't see this issue on Fedora, so maybe limited to mac. Out of that range, this is really the only commit that touches code: https://github.com/clementine-player/Clementine/commit/341dc7326f075a2ae454dfe4548686ae0a7d01d4
And in that change, I think reordered includes is a likely culprit.

jbroadus:
Okay, I _may_ have found something.

Here, `#include "player.h"` is moved after #include `"tagreaderclient.h"`:
https://github.com/clementine-player/Clementine/compare/1.4.0rc1-332-g62d2f0de9...1.4.0rc1-340-g2172732b1#diff-f7910d8ace8a3c68b0c35617510d98030a8526fd8f075109572680e2ab19ec91L21

player.h includes config.h. tagreaderclient.h does not. tagreaderclient.h includes core/workerpool.h which has this mac-specific macro with input that should be provided by config.h:
```
#if defined(Q_OS_MAC) && defined(USE_BUNDLE)
  search_path << qApp->applicationDirPath() + "/" + USE_BUNDLE_DIR;
#endif
```

liofeu:
Many thanks @jbroadus, let me know if I can test anything.

liofeu:
Yes, it does!
Many thanks

Issue #6937, opened 2021-03-01
Cannot read any audio file anymore


liofeu:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [x ] I checked the issue tracker for similar issues
- [ x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [x ] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: macOS Catalina 10.15.7
- Clementine version: last version 01 March 2021 (In the menu "Clementine > About Clementine" is grey, I cannot click on it..)

### Expected behaviour / actual behaviour

I cannot read any file from my library, even after doing a full rescan of the library. It behaves like if each audio file were 0 second and does it for all the files of the playlist. After having tried all the playlist files, Clementine returns the message "Playlist finished". 

### Steps to reproduce the problem (only for bugs)

Create a new playlist and import a sound file. You should be able to see it (title, duration, ...). Click play.

liofeu:
It started suddenly today (I had a previous version of Clementine 1.4) while it was working with the same version a couple of days before. Then I downloaded the last Clementine version 1.4.0rc1-482-g0c099ab6f, and it is the same.

Same after restarting the computer.

jbroadus:
So, clementine can still discover the music during scan and read tags? What kind of music files are they?

liofeu:
Yes, it can still discover the music (at least it can open a new file 
and put it in a playlist)

The music files are my whole music library (mp3, oggs, mp4, ...) which I 
have used with Clementine for years

liofeu:
Hum, I have just remembered that before this issue started, I unticked the option  "Music Library > Monitor the library". Then I ticked it back and I am now able to read my library

The weird thing is that I don't manage to reproduce the issue by unticking back the option, even by restarting Clementine...

liofeu:
I reopen this issue because it comes back from time to time. Restarting Clementine sometimes solves the issue.

nugged:
found very similar behavior - I was on very old clementine, installed latest 496 build, and any file I add – hanged on "reading tracks: 0%",

I tried to downgrade, one-by-one, jumped over few releases, so what I found:

when I downgraded at last to build 315, it started to work. And when now I upgraded back even to hours-ago fresh build 498 – it still works, **and adds files now.** Hurray!

[...]

PR #6951, opened 2021-03-16
Mac: Fix exec search path 

jbroadus:
workerpool.h utilizes the MAC-specific USE_BUNDLE option, but the file didn't directly include the config header file. Cleanup change 341dc73 reordered some headers, so it was no longer included as a side-effect either.

This also moves /src/config.h to /include/clementine-config.h to remove dependencies of /ext on /src.


jbroadus:
This _might_ fix #6950 and #6937

PR #6954, opened 2021-03-17
Remove /ext dependencies on /src

jbroadus:
timeconstants.h was ext's last dependency on src, so move to global include. Keep the core/ path to minimize impact. Remove the src include paths in /etc.


hatstand:
I think we'd move `timeconstants.h` to `ext/libclementine-common/core` given it's already Apache licensed.

Issue #6611, opened 2020-03-07
Rain extra SSL handshake error


jbroadus:
```
$ curl -L -v http://data.clementine-player.org/rainymood
*   Trying 104.24.120.206:80...
* TCP_NODELAY set
* Connected to data.clementine-player.org (104.24.120.206) port 80 (#0)
> GET /rainymood HTTP/1.1
> Host: data.clementine-player.org
> User-Agent: curl/7.66.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 301 Moved Permanently
< Date: Sat, 07 Mar 2020 07:12:09 GMT
< Transfer-Encoding: chunked
< Connection: keep-alive
< Cache-Control: max-age=3600
< Expires: Sat, 07 Mar 2020 08:12:09 GMT
< Location: https://data.clementine-player.org/rainymood
< Server: cloudflare
< CF-RAY: 570281ac4e09f58d-SEA
< 
* Ignoring the response-body
* Connection #0 to host data.clementine-player.org left intact
* Issue another request to this URL: 'https://data.clementine-player.org/rainymood'
*   Trying 104.24.121.206:443...
* TCP_NODELAY set
* Connected to data.clementine-player.org (104.24.121.206) port 443 (#1)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
* TLSv1.3 (IN), TLS handshake, Finished (20):
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.3 (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / TLS_AES_256_GCM_SHA384
* ALPN, server accepted to use h2
* Server certificate:
*  subject: C=US; ST=CA; L=San Francisco; O=Cloudflare, Inc.; CN=sni.cloudflaressl.com
*  start date: Jan 28 00:00:00 2020 GMT
*  expire date: Oct  9 12:00:00 2020 GMT
[...]

PR #6959, opened 2021-03-22
settings: Recategorize background stream settings

jbroadus:
Move the background streams settings page to the general category.

General seemed like a better place for this since only 1 of the 3 is a cloud stream.


jbroadus:
Note that the rain stream is still broken due to a service configuration issue. See #6611

PR #6962, opened 2021-03-27
settings: Make the internet header selectable

jbroadus:
Make the "Internet services" tab the internet providers category header instead of a tab under "User Interface".


jbroadus:
Let me know what you think of this, @hatstand .  It is a little strange that this category header different from the others, so we could make them all selectable and present an empty page for the others. We also lose the internet services icon. Another option would be to just move the "Internet services" tab under "Internet providers", but that also feels a bit awkward.

hatstand:
🤷

Issue #6958, opened 2021-03-22
Smart playlist feature 'file name starts with' not working on Ubuntu 20.10


dingeling:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [x ] I checked the issue tracker for similar issues
- [ x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [ x] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Ubuntu 20.10
- Clementine version: 1.4 rc2 from Ubuntu repository and latest git 1.4.0rc1-516-g8b566b2a7

### Expected behaviour / actual behaviour
using smart playlist feature with term "file name starts with" found files should be visible in list. files are located at "/mnt/Music". the executed query as seen in debug output is:

"SELECT ROWID,title, album, artist, albumartist, composer, track, disc, bpm, year, genre, comment, compilation, bitrate, samplerate, directory, filename, mtime, ctime, filesize, sampler, art_automatic, art_manual, filetype, playcount, lastplayed, rating, forced_compilation_on, forced_compilation_off, effective_compilation, skipcount, score, beginning, length, cue_path, unavailable, effective_albumartist, etag, performer, grouping, lyrics, originalyear, effective_originalyear FROM songs WHERE (filename LIKE 'file:///mnt/Music%') AND unavailable = 0 ORDER BY random()"

no results are shown. using SQLiteStudio with the same query i found 3269 entries.

i debugged clementine and found out "while (query->next()) { ... }" at src/library/librarybackend.cpp line 1106  isn't triggered

on windows the same step is working as expected.

### Steps to reproduce the problem (only for bugs)
use smart playlist feature with term "file name starts with"

![SQLiteStudio](https://user-images.githubusercontent.com/5067143/111993611-1d7a6000-8b17-11eb-96b1-5a05c01380ea.png)

jbroadus:
This works on my fedora machine, but I can repro the issue on an ubuntu vm. It also fails on queries with the sqlite command line.

jbroadus:
Okay, the filename is stored as a blob. On the newer sqlite version, the `like` function must be handling the conversion to text. Using the expression `CAST(filename as TEXT)` in the like clause works on the older Ubuntu version.

dingeling:
[...]

PR #6963, opened 2021-03-27
Fix smart playlist search terms

jbroadus:
- Cast filename to TEXT in queries
In some versions/builds of sqlite, the LIKE function doesn't match blobs.
- Escape pattern characters in search term
When using the LIKE function, use \ as an escape character and prepend that to occurrences of the pattern characters _ and % in the search term.

Fix for #6958

### DIFF
--- a/src/smartplaylists/searchterm.cpp
+++ b/src/smartplaylists/searchterm.cpp
@@ -23,6 +23,13 @@
 
 namespace smart_playlists {
 
+const char* kEscClause = " ESCAPE '#'";
+
+// Replace % and _ with #% and #_
+static QString Escape(QString term) {
+  return term.replace(QRegExp("([%_#])"), "#\\1");
+}
+
 SearchTerm::SearchTerm() : field_(Field_Title), operator_(Op_Equals) {}
 
 SearchTerm::SearchTerm(Field field, Operator op, const QVariant& value)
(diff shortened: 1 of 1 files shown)


src/smartplaylists/searchterm.cpp
```diff
@@ -23,6 +23,13 @@
 
 namespace smart_playlists {
 
+const QString esc_clause(" ESCAPE \"\\\"");
```
hatstand, line 26: Best to avoid non-POD static data.

jbroadus: Fixed.

src/smartplaylists/searchterm.cpp
```diff
@@ -23,6 +23,13 @@
 
 namespace smart_playlists {
 
+const char* esc_clause = " ESCAPE \"\\\"";
```
hatstand, line 26: `kEscClause`

src/smartplaylists/searchterm.cpp
```diff
@@ -23,6 +23,13 @@
...
+
+// Replace % and _ with \% and \_
+static QString Escape(QString term) {
+  return term.replace(QRegExp("([%_])"), "\\\\1");
```
hatstand, line 30: Escaping with something other than `\` might make this easier to read.

src/smartplaylists/searchterm.cpp
jbroadus: Fixed.

src/smartplaylists/searchterm.cpp
jbroadus: Changed to #. Also realized that we need to escape the escape character since we're not always encoding.

src/smartplaylists/searchterm.cpp
```diff
@@ -74,8 +81,10 @@ QString SearchTerm::ToSql() const {
...
 
   // File paths need some extra processing since they are stored as
-  // encoded urls in the database.
+  // encoded urls in the database. In some versions of sqlite, the
+  // LIKE function doesn't handle the blob form, so cast it to TEXT.
   if (field_ == Field_Filepath) {
+    col = "CAST (" + col + " AS TEXT)";
     if (operator_ == Op_StartsWith || operator_ == Op_Equals) {
       value = QUrl::fromLocalFile(value).toEncoded();
     } else {
```
jbroadus, line 90: There's another problem in the else case. The url encoding doesn't work for incomplete URLs. That can be fixed separately.

PR #6965, opened 2021-03-27
Radiobrowser url handler: emit metadata for radio stations

ctrlaltca:
Radio-browser.info stations can be added to the "saved radio" list.
Right now, when playing a station from the "saved radio", no metadata about the radio station is available and only the "radiostation:<uuid>" is shown ad the station name in the playlist.
With this change, the radiobrowser service will emit a signal to update the stream metadata when a station using the "radiobrowser:" protocol gets played.
Unfortunately this required an additional network call to retrieve the station art.

While implementing this, i found out that saved radios have an "url_logo" property that is currently useless: even if you set it, the image url won't be passed to any method and won't be shown nowhere.
If i get it correctly the same seems to apply to the "name" field: even if it's set to the QStandardItem, `Playlist` will actually only use the station url.

### DIFF
--- a/src/internet/radiobrowser/radiobrowserservice.cpp
+++ b/src/internet/radiobrowser/radiobrowserservice.cpp
@@ -43,8 +43,8 @@ Q_DECLARE_METATYPE(QStandardItem*)
 namespace {
 struct Branch {
   QString name;
-  QString listUrl;
-  QString itemsUrl;
+  QString list_url;
+  QString items_url;
   RadioBrowserService::Type type;
 };
 
@@ -187,8 +187,8 @@ void RadioBrowserService::RefreshRootItem() {
         IconLoader::Load("icon_radio", IconLoader::Lastfm), QString());
     item->setText(branch.name);
     item->setData(branch.type, InternetModel::Role_Type);
-    item->setData(branch.listUrl, RadioBrowserService::Role_ListUrl);
-    item->setData(branch.itemsUrl, RadioBrowserService::Role_ItemsUrl);
+    item->setData(branch.list_url, RadioBrowserService::Role_ListUrl);
+    item->setData(branch.items_url, RadioBrowserService::Role_ItemsUrl);
     item->setData(true, InternetModel::Role_CanLazyLoad);
     root_->appendRow(item);
   }
@@ -200,10 +200,10 @@ void RadioBrowserService::RefreshCategory(QStandardItem* item) {
     return;
   }
 
-  QString determinedUrl = item->data(RadioBrowserService::Role_ListUrl)
-                              .toString()
-                              .arg(main_server_url_);
-  QUrl url(determinedUrl);
+  QString determined_url = item->data(RadioBrowserService::Role_ListUrl)
+                               .toString()
+                               .arg(main_server_url_);
+  QUrl url(determined_url);
 
   QNetworkReply* reply = network_->get(QNetworkRequest(url));
   int task_id = app_->task_manager()->StartTask(tr("Getting channels"));
@@ -236,10 +236,10 @@ void RadioBrowserService::RefreshCategoryItem(QStandardItem* item) {
   }
 
   QStandardItem* parent = item->parent();
-  QString determinedUrl = parent->data(RadioBrowserService::Role_ItemsUrl)
-                              .toString()
-                              .arg(main_server_url_, item->text());
-  QUrl url(determinedUrl);
+  QString determined_url = parent->data(RadioBrowserService::Role_ItemsUrl)
+                               .toString()
+                               .arg(main_server_url_, item->text());
+  QUrl url(determined_url);
 
   QNetworkReply* reply = network_->get(QNetworkRequest(url));
   int task_id = app_->task_manager()->StartTask(tr("Getting channels"));
@@ -254,10 +254,10 @@ void RadioBrowserService::RefreshTop100(QStandardItem* item) {
     return;
   }
 
-  QString determinedUrl = item->data(RadioBrowserService::Role_ItemsUrl)
-                              .toString()
-                              .arg(main_server_url_);
-  QUrl url(determinedUrl);
+  QString determined_url = item->data(RadioBrowserService::Role_ItemsUrl)
+                               .toString()
+                               .arg(main_server_url_);
+  QUrl url(determined_url);
 
   QNetworkReply* reply = network_->get(QNetworkRequest(url));
   int task_id = app_->task_manager()->StartTask(tr("Getting channels"));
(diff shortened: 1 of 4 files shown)


src/internet/radiobrowser/radiobrowserservice.cpp
```diff
@@ -419,9 +420,28 @@ void RadioBrowserService::ResolveStationUrl(const QUrl& original_url) {
...
 
-    QJsonObject item = document.object();
+    StreamList list;
+    QJsonArray contents = document.array();
+    QJsonValue item = contents.first();
+    if (item.isUndefined()) return;
+
+    Song ret;
+    ret.set_valid(true);
+    ret.set_title(item["name"].toString());
```
jbroadus, line 430: You're build is failing on some platforms because this operator[] override was added in Qt 5.10, but Stretch is still on 5.7.

src/internet/radiobrowser/radiobrowserservice.cpp
```diff
@@ -419,9 +420,28 @@ void RadioBrowserService::ResolveStationUrl(const QUrl& original_url) {
...
     QUrl url(item["url"].toString());
-    emit StationUrlResolved(original_url, url);
+    ret.set_url(url);
+    ret.set_art_automatic(item["favicon"].toString());
+
+    emit StreamMetadataFound(original_url, ret);
+    emit StationUrlResolved(original_url, ret.url());
```
jbroadus, line 436: Would it make sense to emit one signal with the metadata and let the URL handler extract the URL?

jbroadus:
It looks like usage of the StreamMetadataFound path was discontinued several years ago in 4ab4bbc23. It's probably worth digging though that code a bit to make sure it's still sound.

src/internet/radiobrowser/radiobrowserservice.cpp
```diff
@@ -419,9 +420,28 @@ void RadioBrowserService::ResolveStationUrl(const QUrl& original_url) {
     reply->deleteLater();
     QJsonDocument document = ParseJsonReply(reply);
 
-    QJsonObject item = document.object();
+    StreamList list;
+    QJsonArray contents = document.array();
+    QJsonValue item = contents.first();
+    if (item.isUndefined()) return;
```
jbroadus, line 426: Should the url handler emit AsyncLoadComplete with LoadResult::NoMoreTracks in this case? (That might invalidate my suggestion that we use a single signal.)

src/internet/radiobrowser/radiobrowserservice.cpp
```diff
@@ -419,9 +420,28 @@ void RadioBrowserService::ResolveStationUrl(const QUrl& original_url) {
...
+    emit StreamMetadataFound(original_url, ret);
+    emit StationUrlResolved(original_url, ret.url());
+
+    // inform the server that we played the station
+    QString determinedUrl =
+        PlayClickUrl.arg(main_server_url_, original_url.path());
+    QUrl playUrl(determinedUrl);
+    QNetworkReply* playReply = network_->get(QNetworkRequest(playUrl));
+    connect(playReply, &QNetworkReply::finished,
+            [=] { playReply->deleteLater(); });
```
jbroadus, line 444: This last part is where you might consider using the Player::SongChangeRequestProcessed signal.

src/internet/radiobrowser/radiobrowserservice.cpp
ctrlaltca: Fixed, thank you

src/internet/radiobrowser/radiobrowserservice.cpp
ctrlaltca: Yes, the double emit looks fishy to me, too, but  as far as i understand `StreamMetadataFound` can only be emitted by subclasses of `InternetModel` and `InternetService`.
Maybe metadata could be injected inside the `LoadResult` object by the url handler?

ctrlaltca:
> It looks like usage of the StreamMetadataFound path was discontinued several years ago in [4ab4bbc](https://github.com/clementine-player/Clementine/commit/4ab4bbc23ff9f090612c23c0a78322f4789e2aa5). It's probably worth digging though that code a bit to make sure it's still sound.

When I found it I saw it was unused elsewhere, but it looks like it's still working.
Maybe since it's unused it's a good moment to remote it for a better alternative

src/internet/radiobrowser/radiobrowserservice.cpp
ctrlaltca: Right, this is an error, fixed

src/internet/radiobrowser/radiobrowserservice.cpp
ctrlaltca: Done, thank you

src/internet/radiobrowser/radiobrowserservice.cpp
jbroadus: I was thinking that this would just emit StreamMetadataFound and RadioBrowserUrlHandler connect to that signal. But, again, I think the fact that you need a signal in the error case above makes this a less reasonable solution.

jbroadus:
> > It looks like usage of the StreamMetadataFound path was discontinued several years ago in [4ab4bbc](https://github.com/clementine-player/Clementine/commit/4ab4bbc23ff9f090612c23c0a78322f4789e2aa5). It's probably worth digging though that code a bit to make sure it's still sound.
> 
> When I found it I saw it was unused elsewhere, but it looks like it's still working.
> Maybe since it's unused it's a good moment to remote it for a better alternative

I think it's fine to use this path if that's what it's for. We just need to be aware that there may have been some regressions introduced while the code wasn't being exercised.

src/internet/radiobrowser/radiobrowserservice.cpp
```diff
@@ -419,12 +422,38 @@ void RadioBrowserService::ResolveStationUrl(const QUrl& original_url) {
...
+    emit StationUrlResolved(original_url, ret.url());
   });
 }
 
+void RadioBrowserService::SongChangeRequestProcessed(const QUrl& url,
+                                                     bool valid) {
+  if (!valid || url.scheme() != url_handler_->scheme()) return;
+
+  // inform the server that the station is playing
+  QString determinedUrl = PlayClickUrl.arg(main_server_url_, url.path());
```
hatstand, line 450: Snake case for variable names.

ctrlaltca: Fixed, thank you

src/internet/radiobrowser/radiobrowserservice.cpp
ctrlaltca: Reworked using StreamMetadataFound for success and another signal only for failures.

PR #6968, opened 2021-03-31
Fix a typo in a translatable string

fitojb:


hatstand:
I don't think you need to include the generated translation files here as they'll get fixed automatically later.

fitojb:
Thanks, I’ve removed those files from the commit.

Issue #6969, opened 2021-03-31
5.5 Star Rating


xrayspx:
- Operating System: Kubuntu 20.10
- Clementine version: 1.4 rc2

I've found that it's possible to rate songs at higher than 5 stars using the Rating column to slide the rating to 5.5.  This will cause issues when creating playlists with a rule of "Rating equals 5 stars" for example.

### Steps to reproduce the problem (only for bugs)

In the playlist window, expand the Rating column so it's wider than it needs to be to display the 5 stars.  While rating a song, move the mouse beyond the 5 star rating to the right and click in the blank space of the expanded column.  This should rate the track 5.5 stars fairly consistently. 

To test, I've created a smart playlist with the rule Rating is Greater Than 5-Stars which currently contains 10% of songs I intended to rate a 5.

jbroadus:
Fixed in the latest build.
https://github.com/clementine-player/Clementine/releases/tag/1.4.0rc1-533-gf4e70face

xrayspx:
Brilliant, I've installed the new build and it's working as expected thanks!

PR #6970, opened 2021-04-01
Properly bind the ratings values

jbroadus:
RatingPainter::RatingForPos should return a value in 0-1. But if the
passed position was outside of the widget's draw rectangle, a value
outside of that range could be returned.

This is a fix for #6969


src/widgets/ratingwidget.cpp
```diff
@@ -79,6 +79,10 @@ double RatingPainter::RatingForPos(const QPoint& pos, const QRect& rect) {
   const QRect contents = Contents(rect);
   const double raw = double(pos.x() - contents.left()) / contents.width();
 
+  // Check if the position was to the right or left of the rectangle.
+  if (raw < 0) return 0;
```
hatstand, line 83: Probably too new but could use `std::clamp`

PR #6993, opened 2021-04-21
Remove expired Equifax CA root

jbroadus:
The Equifax certificate expired in 2018, so any certificate that was signed by that CA should also be expired.


ext/clementine-tagreader/main.cpp
```diff
@@ -60,8 +60,6 @@ int main(int argc, char** argv) {
 
   QSslSocket::addDefaultCaCertificates(
       QSslCertificate::fromPath(":/certs/godaddy-root.pem", QSsl::Pem));
```
jbroadus, line 62: @hatstand it looks like the godaddy certificate was added to support UbuntuOne. Do you recall?

hatstand: No idea, sorry

PR #6998, opened 2021-04-25
Optimization to avoid save+delete of remote playlists to temporary file

diracsbracket:
This is my first pull request ever, so forgive me if I did something wrong.

### DIFF
--- a/src/core/songloader.cpp
+++ b/src/core/songloader.cpp
@@ -213,7 +213,8 @@ SongLoader::Result SongLoader::LoadLocal(const QString& filename) {
 SongLoader::Result SongLoader::LoadLocalAsync(const QString& filename) {
   // First check to see if it's a directory - if so we will load all the songs
   // inside right away.
-  if (QFileInfo(filename).isDir()) {
+    QFileInfo info = QFileInfo(filename);
+    if (info.isDir()) {
     LoadLocalDirectory(filename);
     return Success;
   }
@@ -239,7 +240,8 @@ SongLoader::Result SongLoader::LoadLocalAsync(const QString& filename) {
     qLog(Debug) << "Parsing using" << parser->name();
 
     // It's a playlist!
-    LoadPlaylist(parser, filename);
+    file.reset();
+    songs_ = parser->Load(&file, filename, info.path());
     return Success;
   }
 
@@ -292,12 +294,6 @@ void SongLoader::EffectiveSongLoad(Song* song) {
     QString filename = song->url().toLocalFile();
     TagReaderClient::Instance()->ReadFileBlocking(filename, song);
   }
-}
-
-void SongLoader::LoadPlaylist(ParserBase* parser, const QString& filename) {
-  QFile file(filename);
-  file.open(QIODevice::ReadOnly);
-  songs_ = parser->Load(&file, filename, QFileInfo(filename).path());
 }
 
 static bool CompareSongs(const Song& left, const Song& right) {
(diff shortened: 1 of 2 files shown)


jbroadus:
To check the formatting locally, you can use clang-format:
`clang-format --style="{BasedOnStyle: Google, DerivePointerBinding: false, Standard: Cpp11}" <filename>`
You can also use the -i option to modify the file in place.

src/core/songloader.cpp
```diff
@@ -674,21 +670,6 @@ bool SongLoader::LoadRemotePlaylist(const QUrl& url) {
...
-  QString playlist_filename =
-      Utilities::SaveToTemporaryFile(data_reply->readAll());
-  if (playlist_filename.isEmpty()) {
-    qLog(Error) << url.toString()
-                << "could not write contents to temporary file";
-    return false;
-  }
-
-  qLog(Debug) << url.toString() << "with MIME" << mime_type << "loading from"
-              << playlist_filename;
```
jbroadus, line 687: We might want to keep the url and mime_type parts of this log message.

diracsbracket: OK. Added the log entry, and clang-formatted the file. Thanks for showing me how.

src/core/songloader.cpp
```diff
@@ -674,21 +670,8 @@ bool SongLoader::LoadRemotePlaylist(const QUrl& url) {
...
-
-  // ...and load it.
-  LoadPlaylist(parser, playlist_filename);
+  qLog(Debug) << "Loading" << url.toString() << "with MIME" << mime_type;
 
-  QFile(playlist_filename).remove();
+  songs_ = parser->Load(data_reply, QString(), QString());
```
jbroadus, line 675: Minor: Load has default args for playlist_path and dir, so you shouldn't need the last two args. Also note that the last param is actually a QDir, not a QString.

diracsbracket: Done. I really do need to be more careful though...
Thanks~

jbroadus:
And finally, you might want to squash your commits.

src/core/songloader.cpp
```diff
@@ -213,7 +213,8 @@ SongLoader::Result SongLoader::LoadLocal(const QString& filename) {
 SongLoader::Result SongLoader::LoadLocalAsync(const QString& filename) {
   // First check to see if it's a directory - if so we will load all the songs
   // inside right away.
-  if (QFileInfo(filename).isDir()) {
+  QFileInfo info = QFileInfo(filename);
```
hatstand, line 216: `QFileInfo info(filename);`

diracsbracket:
> And finally, you might want to squash your commits.

Ah. I remember reading about that indeed. Will do next time. Cheers~

src/core/songloader.cpp
```diff
@@ -674,21 +670,8 @@ bool SongLoader::LoadRemotePlaylist(const QUrl& url) {
     return false;
   }
 
-  // Save them to a temporary file...
```
hatstand, line 677: The comment at the start of the function is no longer accurate.

src/core/songloader.cpp
```diff
@@ -252,7 +254,7 @@ SongLoader::Result SongLoader::LoadLocalAsync(const QString& filename) {
 
     SongList song_list = cue_parser_->Load(&cue, matching_cue,
                                            QDir(filename.section('/', 0, -2)));
-    for (Song song : song_list) {
+    for (Song& song : song_list) {
```
hatstand, line 257: Avoid non-const references.

diracsbracket: It should really be also using `qAsConst()` but that is only available since Qt5.7.
Shall I replace it with
```
#if (QT_VERSION < QT_VERSION_CHECK(5, 7, 0))
for (const Song& song : song_list) 
#else
for (const Song& song : qAsConst(song_list)) 
#endif
```
or simply not bother the CLazy warnings and simply use
```
for (const Song& song : song_list) 
```
since Clementine's code generates quite a few more of those CLazy warnings...

hatstand: `for (const Song& song : song_list) ` is fine.

Issue #7004, opened 2021-05-01
ipod


policeman0077:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [ ] I checked the issue tracker for similar issues
- [ ] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [ ] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System:
- Clementine version:

jbroadus:
It looks like libmtp isn't being detected in the build.
https://github.com/clementine-player/Clementine/runs/2492525259?check_suite_focus=true
I'll see if I can add it to the brew file.

jbroadus:
Could you give this version a try?
https://github.com/clementine-player/Clementine/releases/tag/1.4.0rc1-591-g579d86904

policeman0077:
> Could you give this version a try?
> https://github.com/clementine-player/Clementine/releases/tag/1.4.0rc1-591-g579d86904

Thanks man for your quick reply. Just tried the latest build, it now can detect the iPod. But has a new error which says libgpod is not compiled.

<img width="1131" alt="Snipaste_2021-05-05_11-58-16" src="https://user-images.githubusercontent.com/982502/117095587-5144d880-ad99-11eb-8e99-2548d34c868a.png">
<img width="1129" alt="Snipaste_2021-05-05_11-58-42" src="https://user-images.githubusercontent.com/982502/117095591-543fc900-ad99-11eb-902c-d5cb4c3a5c31.png">
<img width="1130" alt="Screen Shot 2021-05-05 at 11 58 48 AM" src="https://user-images.githubusercontent.com/982502/117095594-573ab980-ad99-11eb-9c25-60c39feca8cc.png">

jbroadus:
The change that @jonaski made on Strawberry can probably be backported until the libpod package is available.

jbroadus:
The latest has Jonas' change: https://github.com/clementine-player/Clementine/releases/tag/1.4.0rc1-596-g590bcf1c7

policeman0077:
thanks man i will have a test and report back

PR #7008, opened 2021-05-03
Enable libmtp on mac builds

jbroadus:
Add libmtp to BrewFile and add link directory in CMakeLists.txt.


jbroadus:
This is a potential fix for #7004, but I don't have a mac to test with. It probably needs a smoke test before merging.

PR #7010, opened 2021-05-06
Remove obsolete DeviceKit.

jbroadus:
DeviceKit-disks, renamed udisks, was superseded by udisks2 in 2012.


hatstand:
Can we just remove it instead?

jbroadus:
> Can we just remove it instead?

Yeah, I don't think anybody will miss it.

PR #7011, opened 2021-05-07
Add libgpod for mac

jbroadus:
This change was taken from Strawberry. When and if a homebrew libgpod package becomes available, this change can be replaced.

Reference: strawberrymusicplayer/strawberry@f2c7df3


jbroadus:
Follow-up change for #7004

.github/workflows/all.yml
```diff
@@ -890,6 +890,31 @@ jobs:
...
+          patch -p0 < patch-tools-generic-callout.c.diff
+          patch -p1 < libgpod-libplist.patch
+          chmod u+x autogen.sh
+          ./autogen.sh
+          ./configure --disable-more-warnings \
+                      --disable-silent-rules \
+                      --disable-udev \
+                      --disable-pygobject \
+                      --with-python=no
+          make -j8
```
hatstand, line 915: `-j8` is optimistic on the instances we use here. `-j2` makes more sense.

Brewfile
```diff
@@ -17,3 +17,14 @@ brew 'protobuf'
 brew 'protobuf-c'
 brew 'qt@5'
 brew 'tomahawk-player/homebrew-tomahawkqt5/liblastfm'
+
+# These were added for libgpod support and can probably be removed if/when the
+# libgpod package is available.
+brew 'wget'
```
hatstand, line 23: alpha-sort

.github/workflows/all.yml
jbroadus: Yeah, this was cut-and-paste from Strawberry. I'll fix.

Brewfile
jbroadus: Done

Brewfile
```diff
@@ -1,19 +1,27 @@
...
 brew 'glib'
+brew 'gnome-common'
 brew 'google-sparsehash'
 brew 'gst-libav'
 brew 'gst-plugins-bad'
 brew 'gst-plugins-base'
 brew 'gst-plugins-good'
 brew 'gst-plugins-ugly'
 brew 'gstreamer'
+brew 'gtk-doc'
```
jbroadus, line 18: I'm not sure why this is needed. Running a test to see if it can be removed.

PR #7021, opened 2021-05-15
Improved CD metadata loading

### LINKED ISSUE #7019: CD Ripping Dialog displaying metadata only after first opening
### Before posting

- [x] I tried the latest Clementine build from ~~[here](https://builds.clementine-player.org/)~~ the master branch

### System information

- Operating System: Kubuntu 20.04 "Focal Fossa"
- Clementine version: repository HEAD (commit 75de5970; 1.4.0rc1)

Expected: Upon opening the CD Ripping dialog, clementine fetches metadata and fills the dialog.
Currently: The first time after switching the disc, opening the dialog only displays song durations but no other metadata. Closing the dialog window and opening it again makes metadata appear. Restarting Clementine and opening the dialog will display metadata immediately if the disc was not changed.

- Start Clementine
- Insert a CD
- Open CD ripping dialog; observe that songtitles, album name and artist will not be filled in and debug log will not print musicbranz discid
- Close CD rippign dialog
- Open CD ripping dialog; observe that all metadata are now filled in (and debug log prints discid)

## Additional info
I belief this is related to the timeout for reading TOC and TAG messages in CddaSongLoader being a bit too low, at least for my CD drive. It might work for other drives, so the bug might not be reproducible on all systems. PR coming up.

### LINKED ISSUE #7020: CD Ripping dialog displaying wrong number of tracks if discid is not unique
### Before posting

- [x] I tried the latest Clementine build from ~~[here](https://builds.clementine-player.org/)~~ master branch

### System information

- Operating System: Kubuntu 20.04 "Focal Fossa"
- Clementine version: repository head (1.4.0rc1, commit 75de5970 )

Expected: CD ripping dialog displays song titles for medium correctly.

Actual: For the record "Virus" of "Haken", CD ripping dialog displays twice the amount of tracks: the correct one and an instrumental version. This is due to the existence of a "deluxe" release variant of the album that includes a second disc containing instrumental versions of the same trakcks. Due to the track lengths being identical, the musicbrainz discid for both discs is also identical (which arguably makes this an upstream problem with the specification of the discid). MusicBrainzClient therefore reads in tracks for both discs.

Screenshot of the problem below. The disc only has 11 tracks, but 22 are displayed:
screenshot

discid: [cK6nbxjfivIszpqGaVChWzZzmmU-](https://musicbrainz.org/ws/2/discid/cK6nbxjfivIszpqGaVChWzZzmmU-?inc=artists+recordings)

- Insert the Haken - Virus record (or any other disc with a conflicting discid)
- Open CD ripping dialog, observe the issue

lumip:
closes #7019 and closes #7020:
- For #7019: I increased the timeout for reading CD tags in `CddaSongLoader` and also made sure that waiting for messages stops after reading a TOC and a TAG message each, to ensure that no waiting until timeout happens after receiving messages.
- For #7020: If a MusicBrainz metadata requests returns an XML with several media with identical discid, MusicBrainzClient stops after parsing tracks from the first contained medium.
- Additional for #7020: As an additional safeguard, RipCDDialog will ignore a metadata update if the number of tracks contained does not match the number of tracks on the disc.

The solution for parsing the MusicBrainz metadata is not ideal: It implicitly assumes that the first medium contained in the API response is the correct one. The justificiation is that I assume that if identical discids happen, they will most likely correspond to remasters/different variants of the same record of the same artist. So even if the track list is wrong, it will probably be close enough and thus better than the default "Track x" naming.
However, long term it would be better to either let the user choose the correct track list, which would require an additional popup dialog, or read additional metadata from the disc (CD-text, if present) to disambiguate (based on my dabbling with it, I'm not sure if this is possible with gstreamer, though).

### DIFF
--- a/src/devices/cddasongloader.cpp
+++ b/src/devices/cddasongloader.cpp
@@ -49,6 +49,7 @@ QUrl CddaSongLoader::GetUrlFromTrack(int track_number) const {
 void CddaSongLoader::LoadSongs() {
   QtConcurrent::run(this, &CddaSongLoader::LoadSongsFromCdda);
 }
+
 void CddaSongLoader::LoadSongsFromCdda() {
   QMutexLocker locker(&mutex_load_);
   cdio_ = cdio_open(url_.path().toLocal8Bit().constData(), DRIVER_DEVICE);
@@ -155,6 +156,8 @@ void CddaSongLoader::LoadSongsFromCdda() {
           songs[i++].set_length_nanosec(duration);
         }
       }
+      g_list_free(entries);
+      gst_toc_unref(toc);
     }
     gst_message_unref(msg_toc);
   }
@@ -172,9 +175,10 @@ void CddaSongLoader::LoadSongsFromCdda() {
       emit MusicBrainzDiscIdLoaded(musicbrainz_discid);
 
       g_free(string_mb);
-      gst_message_unref(msg_tag);
-      gst_tag_list_free(tags);
     }
+
+    gst_message_unref(msg_tag);
+    gst_tag_list_free(tags);
   }
 
   gst_element_set_state(pipeline, GST_STATE_NULL);
--- a/src/devices/cddasongloader.cpp
+++ b/src/devices/cddasongloader.cpp
@@ -123,12 +123,12 @@ void CddaSongLoader::LoadSongsFromCdda() {
   GstMessage* msg = nullptr;
   GstMessage* msg_toc = nullptr;
   GstMessage* msg_tag = nullptr;
-  GstMessageType msg_filter = static_cast<GstMessageType>(GST_MESSAGE_TOC | GST_MESSAGE_TAG);
-  while ((msg = gst_bus_timed_pop_filtered(
-              GST_ELEMENT_BUS(pipeline), 10 * GST_SECOND,
-              msg_filter))) {
-
-    msg_filter = static_cast<GstMessageType>(static_cast<int>(msg_filter) ^ static_cast<int>(GST_MESSAGE_TYPE(msg)));
+  GstMessageType msg_filter =
+      static_cast<GstMessageType>(GST_MESSAGE_TOC | GST_MESSAGE_TAG);
+  while ((msg = gst_bus_timed_pop_filtered(GST_ELEMENT_BUS(pipeline),
+                                           10 * GST_SECOND, msg_filter))) {
+    msg_filter = static_cast<GstMessageType>(
+        static_cast<int>(msg_filter) ^ static_cast<int>(GST_MESSAGE_TYPE(msg)));
     if (GST_MESSAGE_TYPE(msg) == GST_MESSAGE_TOC) {
       if (msg_toc)
         gst_message_unref(msg_toc);  // Shouldn't happen, but just in case
--- a/src/devices/cddasongloader.cpp
+++ b/src/devices/cddasongloader.cpp
@@ -156,7 +156,6 @@ void CddaSongLoader::LoadSongsFromCdda() {
           songs[i++].set_length_nanosec(duration);
         }
       }
-      g_list_free(entries);
       gst_toc_unref(toc);
     }
     gst_message_unref(msg_toc);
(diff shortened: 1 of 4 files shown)


src/devices/cddasongloader.cpp
```diff
@@ -122,9 +123,12 @@ void CddaSongLoader::LoadSongsFromCdda() {
...
   GstMessage* msg_tag = nullptr;
-  while ((msg = gst_bus_timed_pop_filtered(
-              GST_ELEMENT_BUS(pipeline), 2 * GST_SECOND,
-              (GstMessageType)(GST_MESSAGE_TOC | GST_MESSAGE_TAG)))) {
+  GstMessageType msg_filter =
+      static_cast<GstMessageType>(GST_MESSAGE_TOC | GST_MESSAGE_TAG);
+  while ((msg = gst_bus_timed_pop_filtered(GST_ELEMENT_BUS(pipeline),
+                                           10 * GST_SECOND, msg_filter))) {
+    msg_filter = static_cast<GstMessageType>(
+        static_cast<int>(msg_filter) ^ static_cast<int>(GST_MESSAGE_TYPE(msg)));
```
jbroadus, line 131: According to the docs, we may receive multiple TAG messages with different pieces of data. Maybe we should parse the message in the loop and filter the message after we find the disc id?
https://gstreamer.freedesktop.org/documentation/application-development/advanced/metadata.html

src/devices/cddasongloader.cpp
```diff
@@ -152,6 +156,8 @@ void CddaSongLoader::LoadSongsFromCdda() {
           songs[i++].set_length_nanosec(duration);
         }
       }
+      g_list_free(entries);
```
jbroadus, line 159: I don't think we're supposed to free the entries list. gst_toc_get_entries documentation shows "transfer none".

src/devices/cddasongloader.cpp
lumip: Good point, thanks. Not sure how to handle this best - I think I'll look into setting up a callback that waits for a tag message containing the discid.

src/devices/cddasongloader.cpp
lumip: Ah, I see. I'll revert that. Not entirely familiar with gstreamer and find their docs somewhat confusing at times.

lumip: the gst_toc_unref seems to be misplaced also .. it should, of course, go into the if(toc) {} body. I'll fix that as well.

src/devices/cddasongloader.cpp
```diff
@@ -120,58 +121,54 @@ void CddaSongLoader::LoadSongsFromCdda() {
...
+          int i = 0;
+          for (GList* node = entries; node != nullptr; node = node->next) {
+            GstTocEntry* entry = static_cast<GstTocEntry*>(node->data);
+            quint64 duration = 0;
+            gint64 start, stop;
+            if (gst_toc_entry_get_start_stop_times(entry, &start, &stop))
+              duration = stop - start;
+            songs[i++].set_length_nanosec(duration);
+          }
+          gst_toc_unref(toc);
```
jbroadus, line 145: I think you had this unref in the correct spot before.

lumip: You are right, foolish me :D Should be fixed now.

src/devices/cddasongloader.cpp
```diff
@@ -120,58 +121,52 @@ void CddaSongLoader::LoadSongsFromCdda() {
...
+      }
+    } else if (GST_MESSAGE_TYPE(msg) == GST_MESSAGE_TAG) {
+      // Handle TAG message: generate MusicBrainz DiscId
+
+      GstTagList* tags = nullptr;
+      gst_message_parse_tag(msg, &tags);
+      char* string_mb = nullptr;
+      if (gst_tag_list_get_string(tags, GST_TAG_CDDA_MUSICBRAINZ_DISCID,
+                                  &string_mb)) {
+        QString musicbrainz_discid(string_mb);
```
hatstand, line 159: `::fromAscii` or `::fromUtf8()` for clarity?

lumip: gstreamer doc states that strings returned by `gst_tag_list_get_string` are UTF-8 encoded. Changed above to using `::fromUtf8()` for clarity accordingly.

PR #7024, opened 2021-05-15
Detecting media change in CD ripper dialog.

### LINKED ISSUE #7023: Improvement: Update data in CD rip dialog when medium is changed
Currently the CD ripping dialog does not show any reaction when the disc is removed or changed until the rip button is pressed. It would be great for the dialog to automatically read in new metadata when the disc is changed.

lumip:
Closes #7023. I essentially added a thread to `RipCDDialog` that periodically polls `Ripper::MediaChanged()` and issues a reloading of track list / metadata if the return value is `true`.

Added a mutex that controls access to data updates in `RipCDDialog` to avoid potential race conditions from media changes and musicbrainz metadata responses (not sure if those could really happen, but figured I'd rather be on the safe side).

### DIFF
--- a/src/ripper/ripcddialog.cpp
+++ b/src/ripper/ripcddialog.cpp
@@ -60,7 +60,6 @@ RipCDDialog::RipCDDialog(QWidget* parent)
       loader_(new CddaSongLoader(QUrl(), this)),
       mediaChangedWatchingThread_(*this),
       mutex_() {
-
   QMutexLocker l(&mutex_);
   // Init
   ui_->setupUi(this);
@@ -138,7 +137,8 @@ RipCDDialog::RipCDDialog(QWidget* parent)
 }
 
 RipCDDialog::~RipCDDialog() {
-  QMutexLocker l(&mutex_);  // ensure now one is holding a lock on the mutex when destroying..
+  QMutexLocker l(&mutex_);  // ensure now one is holding a lock on the mutex
+                            // when destroying..
 }
 
 bool RipCDDialog::CheckCDIOIsValid() { return ripper_->CheckCDIOIsValid(); }
@@ -344,8 +344,9 @@ void RipCDDialog::ResetDialog() {
   ui_->discLineEdit->clear();
 }
 
-RipCDDialog::MediaChangedPollingThread::MediaChangedPollingThread(RipCDDialog& dialog)
-  : QThread(&dialog), dialog_(dialog) { }
+RipCDDialog::MediaChangedPollingThread::MediaChangedPollingThread(
+    RipCDDialog& dialog)
+    : QThread(&dialog), dialog_(dialog) {}
 
 RipCDDialog::MediaChangedPollingThread::~MediaChangedPollingThread() {
   requestInterruption();
@@ -354,13 +355,13 @@ RipCDDialog::MediaChangedPollingThread::~MediaChangedPollingThread() {
 
 void RipCDDialog::MediaChangedPollingThread::run() {
   while (!isInterruptionRequested()) {
-      if (dialog_.ripper_->MediaChanged()) {
-        qLog(Debug) << "media changed!";
-        dialog_.ResetDialog();
-        if (dialog_.CheckCDIOIsValid()) {
-          dialog_.loader_->LoadSongs();
-        }
+    if (dialog_.ripper_->MediaChanged()) {
+      qLog(Debug) << "media changed!";
+      dialog_.ResetDialog();
+      if (dialog_.CheckCDIOIsValid()) {
+        dialog_.loader_->LoadSongs();
       }
+    }
     msleep(500);
   }
 }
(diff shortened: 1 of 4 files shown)


src/ripper/ripcddialog.cpp
```diff
@@ -128,9 +131,15 @@ RipCDDialog::RipCDDialog(QWidget* parent)
...
   }
+
+  // Start thread for polling
+  mediaChangedWatchingThread_.start();
 }
 
-RipCDDialog::~RipCDDialog() {}
+RipCDDialog::~RipCDDialog() {
+  QMutexLocker l(&mutex_);  // ensure now one is holding a lock on the mutex
+                            // when destroying..
```
jbroadus, line 141: The thread locks the mutex multiple times per cycle, so it may call this class's methods after this. The Qt child objects aren't destroyed until the base QObject destructor is called. It would be safer to tell the thread to exit and wait for it here.

src/ripper/ripcddialog.cpp
```diff
@@ -324,10 +335,33 @@ QString RipCDDialog::ParseFileFormatString(const QString& file_format,
...
+    if (dialog_.ripper_->MediaChanged()) {
+      qLog(Debug) << "media changed!";
+      dialog_.ResetDialog();
+      if (dialog_.CheckCDIOIsValid()) {
+        dialog_.loader_->LoadSongs();
+      }
+    }
+    msleep(500);
+  }
+}
```
jbroadus, line 367: Could this be done with a QTimer in RipCDDialog instead of a separate thread? You shouldn't need the mutex in that case.

lumip: Just to confirm whether I understand Qt correctly: There would no longer be need for the mutex because the timer event would be handled by the main event loop, i.e., no concurrency actually happens?

jbroadus: That's correct. You can connect the timer's timeout() signal to a slot and, by default, the slot code will run on the recipient's thread. It looks like there were some new callOnTimeout methods added to QTimer in 5.12, but we need to be compatible with Qt 5.7 until the Debian Stretch EOL.

src/ripper/ripcddialog.cpp
```diff
@@ -128,6 +129,9 @@ RipCDDialog::RipCDDialog(QWidget* parent)
       break;
     }
   }
+
+  // Start thread for polling
```
hatstand, line 133: This isn't really a thread.

src/ripper/ripcddialog.h
```diff
@@ -83,5 +85,6 @@ class RipCDDialog : public QDialog {
   Ripper* ripper_;
   bool working_;
   CddaSongLoader* loader_;
+  QBasicTimer media_changed_timer_;
```
hatstand, line 88: Why not use `QTimer`? It's more consistent with what we use elsewhere.

hatstand (commented):
I don't think the lock is necessary as the timer events happen on the event loop thread as does everything else AFAICT.

src/ripper/ripcddialog.cpp
lumip: true, it's not anymore. I'll fix the comment

src/ripper/ripcddialog.h
lumip: I based this on the implementation of progress polling in `TranscodeDialog`, which uses the `QBasicTimer`. Seems then like I have picked the one example that's inconsistent with the rest of the code. I can of course change this to `QTimer` here, but shouldn't we then change the `TranscodeDialog` as well then?

src/ripper/ripcddialog.cpp
```diff
@@ -331,3 +335,13 @@ void RipCDDialog::ResetDialog() {
   ui_->yearLineEdit->clear();
   ui_->discLineEdit->clear();
 }
+
+void RipCDDialog::timerEvent(QTimerEvent* e) {
+  if (ripper_->MediaChanged()) {
+    qLog(Debug) << "media changed!";
+    ResetDialog();
+    if (CheckCDIOIsValid()) {
```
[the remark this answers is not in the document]
jbroadus: It looks like it originally just checked that `cdio_` wasn't null. They should have given it a better name here:
https://github.com/clementine-player/Clementine/commit/4251fd43315c176223764ed568e0a77cffae7571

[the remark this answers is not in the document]
lumip: That change seems odd to me and there's not rationale for it given in the commit. I would at least now rename it to `EnsureCDIOIsOpened` or, preferrably, move opening of the cdio object back into the class constructor (or at least otherwise treat it internally and not expose this to other classes, ensuring access to resources like this should be encapsulated...)

src/ripper/ripcddialog.h
lumip: I replace the `QBasicTimer` with a `QTimer` now. Should I look into doing the same in `TranscodeDialog` then @hatstand ?

lumip:
Even though I think this should now be done in the sense of having addressed the review comments, I'll convert this to a draft for now while I'm thinking about how to make PR #7030 work - it might be good to combine those two, so it wouldn't make sense to merge this one at this point.

lumip:
Made obsolete by #7037

Issue #6952, opened 2021-03-16
Songs with an empty Album tag get assigned the folder name as Album tag


AllKind:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [ ] I checked the issue tracker for similar issues
- [ ] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [ ] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Linux Mint 19.3
- Clementine version: 1.4.0rc1-509-g89e9b20df

### Expected behaviour / actual behaviour

Expected: Songs with an empty Album tag should not be altered.
Actual:
The folder name the songs (mp3 and flac) reside in is taken as Album tag. Trying to remove the tag (inline in the Playlist view), results in the tag being reset to the folder name instantly.
This happens since a few releases of 1.4.0rc1, but I cannot tell which one exactly.
The behavior only happens to newly written files.

### Steps to reproduce the problem (only for bugs)
Take some songs with an empty album tag and write them to a (sub)folder of your library.

jbroadus:
I think that was intentional:
https://github.com/clementine-player/Clementine/pull/6743
But I see what you're saying. Seems like that behavior should be configurable.

AllKind:
Thank you for looking into it and answering.
A option to configure that behavior would very much be appreciated.
There enough scenarios, where this logic of folder name guessing does not make sense.

abika:
I also find this very annoying. I use the "Edit track information" functionality a lot and have to remove the Album tag every time before saving.

How can a new feature be introduced replacing the default behavior with a new one only valid for a limited number of users - and without an option to change it back?

jbroadus:
The latest build has a new setting to disable this feature. See the new Song Metadata settings page.
https://github.com/clementine-player/Clementine/releases/tag/1.4.0rc1-617-g776bd3b02

I think this is still broken behavior when the the option is selected, so let's keep the bug open.

petersonespacoporto:
No version for Hirsute at the release link so cannot test... Would love to though this is driving me crazy

jbroadus:
Oh yeah, we should add that build.

jbroadus:
#7033

If you want to try it out before the PR is merged, you can find the artifacts here:
https://github.com/clementine-player/Clementine/pull/7033/checks

[...]

PR #7026, opened 2021-05-16
tagreader: Relocate artist/album/title guessing code

jbroadus:
Move the code that attempts to fill missing song metadata out of the
tagreader worker. In the main process, it will be controllable using
settings and calling context.

The methods were moved into a new SongPathParser class that checks new
settings to determine if action should be taken. Settings UI will be added in an upcoming pull request.


jbroadus:
Partial fix for #6952

jbroadus:
I'm not sure why the workflow isn't running.

PR #7030, opened 2021-05-22
CddaSongLoader now reads tags from CD-Text

lumip:
### Changes so far
In `CddaSongLoader::LoadSongsFromCdda`: I changed the gstreamer source element to `cdiocddasrc` (from uri source), which includes metadata from CD-Text in the tag message for each track and seek through all the tracks. On a high level `LoadSongsFromCdda` now works as follows:

1. read the number of tracks, emit `SongsLoaded` (unchanged)
2. handle TOC message, update song durations, emit `SongDurationLoaded` (unchanged)
3. seek through tracks, fetch CD-Text metadata, emit `SongMetadataLoaded` (new)
4. emit `MusicBrainzDiscIdLoaded`
5. fetch metadata from musicbrainz and emit `SongMetadataLoaded` (as reaction to `MusicBrainzDiscIdLoaded`)

The benefits are that, if CD-Text is present on the disc, at least some metadata are still available even if the user does not have Internet connectivity (or musicbrainz service is unresponsive); it could also help disambiguating non-unique discids (see #7020 ), but that's not currently implemented.

In terms of clients of `CddaSongLoader`, I have adapted the `RipCDDialog` to only update metadata if it receives previously unseed tags, which means it currently fills with CD-Text tags first and then completes everything missing from the musicbrainz response. The only other client is `SongLoader`, which I have left unchanged so far because I wasn't sure of the exact implications of changing stuff there.

#### Other changes in `CddaSongLoader`:

- The `cdio_` object was opened but never used in `LoadSongsFromCdda`. I have removed it there and now initialise it during class construction, would need some input on preferred ways of error handling there (throw an exception?).
- This has removed the need for the mutex as cdio_ is only ever accessed from the main thread (and only via the `HasChanged`) method.
- There was a concurrency bug where the destructor might have been called while the thread running `LoadSongsFromCdda` was still active. I resolved that by storing the `QFuture` returned by `QtConcurrent::run` and waiting on it in the class destructor.

### Going further
I am thinking making the following further changes but would like some feedback on them first:
I dislike that `CddaSongLoader` now emits the `SongMetadataLoaded` signal two times containing potentially different metadata and it is up to the clients to deal with that. I'm thinking of either having `CddaSongLoader` maintain the `SongList` of CD tracks internally and update it when it gets input from a metadata source, then emit the `SongMetadataLoaded` signal ((maybe then rename that to `SongMetadataUpdated`) with a guarantee that previously emitted tags do not change. Clients could then always just use the latest SongList received, which would avoid duplicating code for dealing with this in the clients. While this would introduce more state to `CddaSongLoader`, I don't think there would be any harm in that. It could even be beneifical because the class could then encapsulate media changed checks, effectively only rerunning `LoadSongsFromCdda` if a media change was detected. This logic is currently pushed into its clients.

A stateless alternative would be to not have `CddaSongLoader` do a musicbrainz request and leave that to clients subscribing to the `MusicBrainzDiscIdLoaded` signal, if they need to.

#### Final Note
Let me know if you want me to stop requesting to merge stuff into master - not sure how far along you are with the release candidate...

### DIFF
--- a/src/core/songloader.cpp
+++ b/src/core/songloader.cpp
@@ -156,10 +156,13 @@ SongLoader::Result SongLoader::LoadLocalPartial(const QString& filename) {
 SongLoader::Result SongLoader::LoadAudioCD() {
 #ifdef HAVE_AUDIOCD
   CddaSongLoader* cdda_song_loader = new CddaSongLoader;
-  connect(cdda_song_loader, SIGNAL(SongsDurationLoaded(SongList)), this,
-          SLOT(AudioCDTracksLoadedSlot(SongList)));
-  connect(cdda_song_loader, SIGNAL(SongsMetadataLoaded(SongList)), this,
-          SLOT(AudioCDTracksTagsLoaded(SongList)));
+  connect(cdda_song_loader, &CddaSongLoader::SongsUpdated, this,
+          &SongLoader::AudioCDTracksLoadedSlot);
+  connect(cdda_song_loader, &CddaSongLoader::Finished,
+          [this, cdda_song_loader]() {
+            cdda_song_loader->deleteLater();
+            emit LoadAudioCDFinished(true);
+          });
   cdda_song_loader->LoadSongs();
   return Success;
 #else  // HAVE_AUDIOCD
@@ -171,13 +174,6 @@ SongLoader::Result SongLoader::LoadAudioCD() {
 void SongLoader::AudioCDTracksLoadedSlot(const SongList& songs) {
   songs_ = songs;
   emit AudioCDTracksLoaded();
-}
-
-void SongLoader::AudioCDTracksTagsLoaded(const SongList& songs) {
-  CddaSongLoader* cdda_song_loader = qobject_cast<CddaSongLoader*>(sender());
-  cdda_song_loader->deleteLater();
-  songs_ = songs;
-  emit LoadAudioCDFinished(true);
 }
 #endif  // HAVE_AUDIOCD
 
--- a/src/devices/cddadevice.cpp
+++ b/src/devices/cddadevice.cpp
@@ -34,6 +34,8 @@ CddaDevice::CddaDevice(const QUrl& url, DeviceLister* lister,
       cdda_song_loader_(url) {
   connect(&cdda_song_loader_, SIGNAL(SongsUpdated(SongList)), this,
           SLOT(SongsLoaded(SongList)));
+  connect(&cdda_song_loader_, SIGNAL(Finished()), this,
+          SLOT(SongsLoadingFinished()));
   connect(this, SIGNAL(SongsDiscovered(SongList)), model_,
           SLOT(SongsDiscovered(SongList)));
   connect(&disc_changed_timer_, SIGNAL(timeout()), SLOT(CheckDiscChanged()));
@@ -69,7 +71,10 @@ void CddaDevice::WatchForDiscChanges(bool watch) {
     disc_changed_timer_.stop();
 }
 
-void CddaDevice::ForceLoadSongs() { cdda_song_loader_.LoadSongs(); }
+void CddaDevice::ForceLoadSongs() {
+  cdda_song_loader_.LoadSongs();
+  disc_changed_timer_.stop();
+}
 
 void CddaDevice::LoadSongs() {
   SongList songs = cdda_song_loader_.cached_tracks();
@@ -80,7 +85,19 @@ void CddaDevice::SongsLoaded(const SongList& songs) {
   model_->Reset();
   song_count_ = songs.size();
   emit SongsDiscovered(songs);
+  // When a disc is inserted, cdio_get_media_changed will
+  // return true for two times with a bit of delay in between
+  // (at least on linux).
+  // We clear cdio_get_media_changed after songs are
+  // loaded, so we don't potentially re-read the same disc.terminal
+  // There's a slight chance that this hides an actual
+  // media change, but this should be rare enough to not
+  // be a problem in practice and is easily rectified
+  // by user cycling the disc once more.
+  cdio_get_media_changed(cdio_);
 }
+
+void CddaDevice::SongsLoadingFinished() { disc_changed_timer_.start(); }
 
 void CddaDevice::CheckDiscChanged() {
   if (!cdio_) return;  // do nothing if not initialized
--- a/src/devices/cddadevice.cpp
+++ b/src/devices/cddadevice.cpp
@@ -32,11 +32,7 @@ CddaDevice::CddaDevice(const QUrl& url, DeviceLister* lister,
       cdio_(nullptr),
       disc_changed_timer_(),
       cdda_song_loader_(url) {
-  connect(&cdda_song_loader_, SIGNAL(SongsLoaded(SongList)), this,
-          SLOT(SongsLoaded(SongList)));
-  connect(&cdda_song_loader_, SIGNAL(SongsDurationLoaded(SongList)), this,
-          SLOT(SongsLoaded(SongList)));
-  connect(&cdda_song_loader_, SIGNAL(SongsMetadataLoaded(SongList)), this,
+  connect(&cdda_song_loader_, SIGNAL(SongsUpdated(SongList)), this,
           SLOT(SongsLoaded(SongList)));
   connect(this, SIGNAL(SongsDiscovered(SongList)), model_,
           SLOT(SongsDiscovered(SongList)));
--- a/src/devices/cddadevice.cpp
+++ b/src/devices/cddadevice.cpp
@@ -50,7 +50,7 @@ bool CddaDevice::Init() {
   if (!cdio_) {
     cdio_ = cdio_open(url_.path().toLocal8Bit().constData(), DRIVER_DEVICE);
     if (!cdio_) return false;
-    LoadSongs();
+    ForceLoadSongs();
     WatchForDiscChanges(true);
   }
   return true;
@@ -69,12 +69,17 @@ void CddaDevice::WatchForDiscChanges(bool watch) {
     disc_changed_timer_.stop();
 }
 
-void CddaDevice::LoadSongs() { cdda_song_loader_.LoadSongs(); }
+void CddaDevice::ForceLoadSongs() { cdda_song_loader_.LoadSongs(); }
+
+void CddaDevice::LoadSongs() {
+  SongList songs = cdda_song_loader_.cached_tracks();
+  SongsLoaded(songs);
+}
 
 void CddaDevice::SongsLoaded(const SongList& songs) {
   model_->Reset();
-  emit SongsDiscovered(songs);
   song_count_ = songs.size();
+  emit SongsDiscovered(songs);
 }
 
 void CddaDevice::CheckDiscChanged() {
@@ -89,6 +94,6 @@ void CddaDevice::CheckDiscChanged() {
     song_count_ = 0;
     SongList no_songs;
     SongsLoaded(no_songs);
-    LoadSongs();
+    ForceLoadSongs();
   }
 }
--- a/src/devices/cddadevice.cpp
+++ b/src/devices/cddadevice.cpp
@@ -52,7 +52,7 @@ bool CddaDevice::Init() {
   if (!cdio_) {
     cdio_ = cdio_open(url_.path().toLocal8Bit().constData(), DRIVER_DEVICE);
     if (!cdio_) return false;
-    ForceLoadSongs();
+    LoadSongs();
     WatchForDiscChanges(true);
   }
   return true;
@@ -71,14 +71,9 @@ void CddaDevice::WatchForDiscChanges(bool watch) {
     disc_changed_timer_.stop();
 }
 
-void CddaDevice::ForceLoadSongs() {
+void CddaDevice::LoadSongs() {
   cdda_song_loader_.LoadSongs();
   disc_changed_timer_.stop();
-}
-
-void CddaDevice::LoadSongs() {
-  SongList songs = cdda_song_loader_.cached_tracks();
-  SongsLoaded(songs);
 }
 
 void CddaDevice::SongsLoaded(const SongList& songs) {
@@ -111,6 +106,8 @@ void CddaDevice::CheckDiscChanged() {
     song_count_ = 0;
     SongList no_songs;
     SongsLoaded(no_songs);
-    ForceLoadSongs();
+    LoadSongs();
   }
 }
+
+SongList CddaDevice::songs() const { return cdda_song_loader_.cached_tracks(); }
--- a/src/devices/cddasongloader.cpp
+++ b/src/devices/cddasongloader.cpp
@@ -1,5 +1,6 @@
 /* This file is part of Clementine.
    Copyright 2014, David Sansome <me@davidsansome.com>
+   Copyright 2021, Lukas Prediger <lumip@lumip.de>
 
    Clementine is free software: you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
@@ -27,9 +28,21 @@
 #include "core/timeconstants.h"
 
 CddaSongLoader::CddaSongLoader(const QUrl& url, QObject* parent)
-    : QObject(parent), url_(url), cdda_(nullptr), may_load_(true) {
-  connect(this, SIGNAL(MusicBrainzDiscIdLoaded(const QString&)),
-          SLOT(LoadAudioCDTags(const QString&)));
+    : QObject(parent), url_(url), cdda_(nullptr), may_load_(true), disc_() {
+  connect(this, &CddaSongLoader::MusicBrainzDiscIdLoaded, this,
+          &CddaSongLoader::LoadAudioCDTags);
+  connect(this, &CddaSongLoader::SongsLoaded,
+          [this](const SongList& song_list) {
+            SetDiscTracks(song_list, /*has_titles=*/false);
+          });
+  connect(this, &CddaSongLoader::SongsDurationLoaded,
+          [this](const SongList& song_list) {
+            SetDiscTracks(song_list, /*has_titles=*/false);
+          });
+  connect(this, &CddaSongLoader::SongsMetadataLoaded,
+          [this](const SongList& song_list) {
+            SetDiscTracks(song_list, /*has_titles=*/true);
+          });
 }
 
 CddaSongLoader::~CddaSongLoader() {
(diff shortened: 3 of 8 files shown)


src/devices/cddasongloader.cpp
```diff
@@ -30,10 +30,19 @@ CddaSongLoader::CddaSongLoader(const QUrl& url, QObject* parent)
     : QObject(parent), url_(url), cdda_(nullptr), cdio_(nullptr) {
   connect(this, SIGNAL(MusicBrainzDiscIdLoaded(const QString&)),
           SLOT(LoadAudioCDTags(const QString&)));
+
+  cdio_ = cdio_open(url_.path().toLocal8Bit().constData(), DRIVER_DEVICE);
+  Q_ASSERT(cdio_ != nullptr);  // todo: assert for now, but we need to actually
+                               // handle error cases here
```
jbroadus, line 36: Maybe we can add a separate Init() method that returns a bool? CddaDevice already has an Init method that could call it, though it doesn't currently return anything.

src/devices/cddasongloader.cpp
```diff
@@ -30,10 +30,19 @@ CddaSongLoader::CddaSongLoader(const QUrl& url, QObject* parent)
...
 }
 
 CddaSongLoader::~CddaSongLoader() {
-  if (cdio_) cdio_destroy(cdio_);
+  // The LoadSongsFromCdda methods runs concurrently in a thread and we need to
+  // wait for it to terminate. There's no guarantee that it has terminated when
+  // destructor is invoked.
+  loading_future_.waitForFinished();
```
jbroadus, line 43: Would there be a way to interrupt LoadSongsFromCdda? There are a couple loops that could check for abort flags.

src/devices/cddasongloader.cpp
```diff
@@ -214,13 +304,9 @@ void CddaSongLoader::AudioCDTagsLoaded(
 }
 
 bool CddaSongLoader::HasChanged() {
-  if ((cdio_ && cdda_) && cdio_get_media_changed(cdio_) != 1) {
-    return false;
-  }
-  // Check if mutex is already token (i.e. init is already taking place)
-  if (!mutex_load_.tryLock()) {
```
jbroadus, line 221: From cddadevice.cpp:
```
void CddaDevice::Init() {
  song_count_ = 0;  // Reset song count, in case it was already set
  cdda_song_loader_.LoadSongs();
}

void CddaDevice::Refresh() {
  if (!cdda_song_loader_.HasChanged()) {
    return;
  }
  Init();
}
```
So, the mutex is preventing the LoadSongs from being when there's a loading thread running. That synchronization logic really shouldn't be spanning two classes, but it looks like it was the result of CddaDevice being split into two pieces. But if you do want to stick with this existing pattern, then I think you could check the state of loading_future_ here.

The old code also has a timing hole. You could potentially call Refresh multiple times in succession and kick off several threads before the first thread has a chance to grab the lock. I think that could also be fixed by checking the future object.

src/devices/cddasongloader.cpp
lumip: I'm not a fan of Init() methods if they can be avoided. Having an instance of CddaSongLoader that is not in a properly initialised state (i.e. where `Init()` would have returned `false`) seems pointless because it's unusable. Better to fail during construction, imo.

src/devices/cddasongloader.cpp
lumip: Haven't looked into that yet because `LoadSongsFromCdda` shouldn't really take too long to complete in any case, but I can add a flag for it to abort.

src/devices/cddasongloader.cpp
lumip: I think this could neatly be adressed by my idea of making `CddaSongLoader` more stateful and encapsulating the logic for only re-reading data if the medium was changed. `CddaDevice::Refresh` could then just call `Init` without any checks there

src/devices/cddasongloader.cpp
lumip: I have now introduced a `may_load_` flag and have `LoadSongsFromCdda` interrupt its metadata parsing loop if that is no longer set.

src/devices/cddasongloader.cpp
lumip: I have now encapsulated checks for `HasChanged` in  `CddaSongLoader::LoadSongs`. Subsequent calls to `LoadSongs` without a media change will not perform any reading but simply emit a signal with the complete track list with all metadata once. `CddaDevice::Refresh` therefore can just call `::Init()` now and no more synchronisation issues can occur there.

src/devices/cddasongloader.cpp
```diff
@@ -47,18 +69,86 @@ QUrl CddaSongLoader::GetUrlFromTrack(int track_number) const {
 }
 
 void CddaSongLoader::LoadSongs() {
-  QtConcurrent::run(this, &CddaSongLoader::LoadSongsFromCdda);
+  if (!disc_.has_been_read || HasChanged()) {
+    if (!loading_future_.isRunning()) {
```
jbroadus, line 73: There's isStarted and isRunning. I assumed that isStarted() meant that QtConcurrent::run had been called and that isRunning() meant that the thread entry point had been called. But I may be wrong about that. The QFuture documentation isn't terribly helpful.

src/devices/cddasongloader.cpp
```diff
@@ -47,18 +69,86 @@ QUrl CddaSongLoader::GetUrlFromTrack(int track_number) const {
...
+      disc_.has_been_read = true;
+      loading_future_ =
+          QtConcurrent::run(this, &CddaSongLoader::LoadSongsFromCdda);
+    }
+  } else
+    emit SongsUpdated(disc_.tracks, false);
```
jbroadus, line 79: `CddaDevice::Refresh` is called when the view asks for the items display text, though I'm not sure why:
https://github.com/clementine-player/Clementine/blob/master/src/devices/devicemanager.cpp#L194

That means that this signal could be emitted frequently.

src/devices/cddadevice.cpp
```diff
@@ -42,17 +38,15 @@ CddaDevice::~CddaDevice() {}
...
+                                  // in call to SongsLoaded again
 }
 
-void CddaDevice::Refresh() {
-  if (!cdda_song_loader_.HasChanged()) {
-    return;
-  }
-  Init();
-}
+void CddaDevice::Refresh() { Init(); }
```
jbroadus, line 46: I know this is the existing pattern, but, in my option, it makes more sense to put the code in Refresh() and have Init() call Refresh().

src/devices/cddasongloader.h
```diff
@@ -42,29 +45,69 @@ class CddaSongLoader : public QObject {
 
   // Load songs.
   // Signals declared below will be emitted anytime new information will be
-  // available.
+  // available. Calling LoadSongs will always emit a SongsUpdated signal,
+  // but actual loading from disc will only occur if the medium was changed
+  // since the last time LoadSongs() was called.
   void LoadSongs();
+  // Returns true if the medium/disc has changed since the last check.
   bool HasChanged();
```
jbroadus, line 53: Can this be private now?

src/devices/cddasongloader.cpp
lumip: Hm. I'm not sure that's a problem, and if it is, I'd argue it's not one of `CddaSongLoader`. After all, if you ask me to load some songs for you, you shouldn't be surprised when you get a list of songs... I'll see if I can think of a way to make the `CddaDevice::Refresh` smarter (or maybe something even more upstream of that..)

src/devices/cddasongloader.cpp
lumip: Briefly looking at the source of QFuture didn't help me to figure out the difference either. I tried making that condition  `!(loading_future_.isRunning() || loading_future_.isStarted())` to cover both but found that that prevents loading completely. Apparently it holds that `loading_future_.isStarted() == true` after initialisation of `loading_future_` in the class constructor, which seems really odd to me...

src/devices/cddadevice.cpp
lumip: Done

src/devices/cddasongloader.h
lumip: Done

src/devices/cddasongloader.cpp
jbroadus: Yeah, the refresh looks like it was a bit of a hack.

This loader is different from the other device loaders (mtp, gpod, etc) in that its lifecycle is the same as the device object. The others are created, preform a single load, and then are deleted. And this is the pattern that SongLoader currently follows when using CddaSongLoader. So, maybe CddaSongLoader should be pared down and all of the synchronization should be handled by the device class instead? I think the cdio handle would need to be passed into CddaSongLoader in that case.

lumip: That's a good point. I was going in the other direction so far (just pushed my pending changes): I have now extended CddaDevice to basically suppress SongsUpdated signals unless there was a substantial change. How this works now is that `CddaSongLoader::LoadSongs` will emit a `DiscChanged` signal if it detects a disc change before it starts reading the disc information. `CddaDevice` will then change an internal flag to accumulate all incoming `SongsUpdated` signals until it encounters one with `futher_updates_possible==false`, upon which it will emit its `SongsLoaded` signal and turn off the internal flag.
For calls to `Refresh` that do not result in a `DiscChanged` signal being received, `CddaDevice` ignores/suppresses the `SongsUpdated` signal it receives from `CddaSongLoader`.

It works now but I'm not really happy with it, because it increases the complexity of the signalling protocol of `CddaSongLoader`. Ideally I'd want that protocol to be stateless, i.e., `SongsUpdated` signals just happen and are self-contained - but that would require some very smart solution to the refreshing issue..

I'm actually thinking of absorbing the media change detection of PR #7024 into `CddaSongLoader` (or `CddaDevice`) and get rid off the `ConnectedDevice::Refresh` method entirely (its only used in `CddaDevice` anyways..). 

Another issue I've now run into is that the `cdio_get_media_changed` seems to not be working reliably when more than one `cdio` object is opened, which is now the case for the cd ripping (once held by `Ripper` and once by `CddaSongLoader`, both of which are used by `RipCDDialog`), which really messes with my changes now (which is another reason that motivates me to centralise all cdio related things in one class).

Thoughts?

jbroadus: I wonder if the cdio instance references a singleton? One solution to that could be a static change counter that increments every time a change is detected and each instance could keep a local copy of the count last time it checked. Or maybe the CddaDevice instance can serve the ripper. I'm not sure how it works right now if you try to rip and play the CD at the same time.

I agree with getting rid of `ConnectedDevice::Refresh`. A timer in CddaDevice or CddaSongLoader makes more sense to me.

Also, doing this in #7024, or even a different PR, sounds like a good idea to me.

lumip: > I wonder if the cdio instance references a singleton?
Looking at the code of the linux driver implementation of cdio, this doesn't seem to be the case. Each call to `cdio_open` creates a new instance and opens a new handle to the device. `cdio_get_media_changed` is just a wrapper around the `CDROM_MEDIA_CHANGED` ioctl call. Therefore, it seems that this might actually be a Linux kernel issue (and might not be an issue on other systems). I have somewhat confirmed this with a small test program that uses the ioctl call directly and could confirm that the call only returns 1 once when invoked for several separate file descriptors - this even holds across several independent processes.
I'll keep your suggestions in mind and try to reduce the number of cdio instances held by Clementine to one.

lumip:
I have now rebased this to incorporate the changes made to cdio object handling and device access in #7037 .

src/devices/cddasongloader.cpp
```diff
@@ -140,43 +225,70 @@ void CddaSongLoader::LoadSongsFromCdda() {
...
+                                                 GST_MESSAGE_TAG);
     }
     gst_message_unref(msg);
   }
+  if (loaded_cd_tags) emit SongsMetadataLoaded(tagged_song_list);
+
+  if (!musicbrainz_discid.isEmpty())
+    emit MusicBrainzDiscIdLoaded(musicbrainz_discid);
+  else
+    SongsUpdated(tagged_song_list, false);
```
jbroadus, line 286: Should this be an emit?

src/devices/cddasongloader.cpp
```diff
@@ -55,17 +68,88 @@ bool CddaSongLoader::IsActive() const { return loading_future_.isRunning(); }
...
+  song.set_track(track_number);
+  song.set_id(track_number);
+  song.set_filetype(Song::Type_Cdda);
+  song.set_valid(true);
+  song.set_url(GetUrlFromTrack(track_number));
+  return has_loaded_tags;
+}
+
 void CddaSongLoader::LoadSongsFromCdda() {
   if (!may_load_) return;
```
jbroadus, line 149: Is the object leaked in this case?

src/core/songloader.cpp
```diff
@@ -156,10 +156,8 @@ SongLoader::Result SongLoader::LoadLocalPartial(const QString& filename) {
 SongLoader::Result SongLoader::LoadAudioCD() {
 #ifdef HAVE_AUDIOCD
   CddaSongLoader* cdda_song_loader = new CddaSongLoader;
-  connect(cdda_song_loader, SIGNAL(SongsDurationLoaded(SongList)), this,
-          SLOT(AudioCDTracksLoadedSlot(SongList)));
-  connect(cdda_song_loader, SIGNAL(SongsMetadataLoaded(SongList)), this,
-          SLOT(AudioCDTracksTagsLoaded(SongList)));
+  connect(cdda_song_loader, SIGNAL(SongsUpdated(SongList, bool)), this,
+          SLOT(AudioCDTracksLoadedSlot(SongList, bool)));
```
jbroadus, line 160: While I think that unifying these is mostly correct, I wonder if we should have a separate Finished signal where the connected slot would just delete the object and handle any errors. I think it might be a little clearer than the boolean passed to SongsUpdated.

src/devices/cddasongloader.cpp
lumip: Which object exactly do you mean? I had a brief look at SongLoader::LoadLocalPartial but can't really see how that interacts with this as it seems concerned with local files, not CDs. Correct me if I'm wrong please.

src/core/songloader.cpp
lumip: That's a good idea and I have changed it accordingly.

src/devices/cddasongloader.cpp
jbroadus: Sorry, I read the wrong method name. I meant SongLoader::LoadAudioCD(). That creates an instance of CddaSongLoader and relies on a signal to tell it when it can delete the object. We should make sure your finished signal is emitted in all cases.

lumip: Ah, I see. I have added a `Finished` signal to `CddaSongLoader` according to your suggestion below and double checked that every branch that ends the loading process in `CddaSongLoader`, abort or success, emits `Finished`, so I think that the changes work well with `SongLoader::LoadAudioCD`. It is a bit hard to test with 100% certainty, though.

src/devices/cddasongloader.cpp
lumip: True, I have fixed this now.

src/devices/cddasongloader.h
```diff
@@ -42,30 +43,70 @@ class CddaSongLoader : public QObject {
...
   // Load songs.
-  // Signals declared below will be emitted anytime new information will be
+  // Signals declared below will be emitted anytime new information becomes
   // available.
   void LoadSongs();
   bool IsActive() const;
 
+  // The list of currently cached tracks. This gets updated during calls
+  // LoadSongs() is called. Not thread-safe.
+  SongList chached_tracks() const;
```
jbroadus, line 53: Typo

src/devices/cddadevice.cpp
```diff
@@ -73,12 +71,28 @@ void CddaDevice::WatchForDiscChanges(bool watch) {
...
-void CddaDevice::LoadSongs() { cdda_song_loader_.LoadSongs(); }
+void CddaDevice::ForceLoadSongs() {
+  cdda_song_loader_.LoadSongs();
+  disc_changed_timer_.stop();
+}
+
+void CddaDevice::LoadSongs() {
+  SongList songs = cdda_song_loader_.chached_tracks();
+  SongsLoaded(songs);
+}
```
jbroadus, line 82: Maybe there's a better name for this method. It isn't actually loading the tracks anymore.

src/ripper/ripcddialog.cpp
```diff
@@ -293,20 +291,12 @@ void RipCDDialog::DeviceSelected(int device_index) {
...
-  connect(loader_, SIGNAL(SongsDurationLoaded(SongList)),
+  connect(cdda_device_.get(), SIGNAL(SongsDiscovered(SongList)),
           SLOT(BuildTrackListTable(SongList)));
-  connect(loader_, SIGNAL(SongsMetadataLoaded(SongList)),
-          SLOT(UpdateTrackListTable(SongList)));
-  connect(loader_, SIGNAL(SongsMetadataLoaded(SongList)),
-          SLOT(AddAlbumMetadataFromMusicBrainz(SongList)));
+  connect(cdda_device_.get(), SIGNAL(SongsDiscovered(SongList)),
+          SLOT(SetAlbumMetadata(SongList)));
+  cdda_device_->LoadSongs();
```
jbroadus, line 298: If I'm reading things correctly, the purpose of this is to re-emit the SongsDiscovered signal. Could this have side effects on other users?

jbroadus: Maybe we could keep the song cache in the CddaDevice so that it's always up to date with the model. Then we could just fetch those when this initializes.

src/ripper/ripcddialog.cpp
```diff
@@ -293,20 +291,12 @@ void RipCDDialog::DeviceSelected(int device_index) {
...
-
-  connect(loader_, SIGNAL(SongsDurationLoaded(SongList)),
+  connect(cdda_device_.get(), SIGNAL(SongsDiscovered(SongList)),
           SLOT(BuildTrackListTable(SongList)));
-  connect(loader_, SIGNAL(SongsMetadataLoaded(SongList)),
-          SLOT(UpdateTrackListTable(SongList)));
-  connect(loader_, SIGNAL(SongsMetadataLoaded(SongList)),
-          SLOT(AddAlbumMetadataFromMusicBrainz(SongList)));
+  connect(cdda_device_.get(), SIGNAL(SongsDiscovered(SongList)),
+          SLOT(SetAlbumMetadata(SongList)));
```
jbroadus, line 297: Would it make sense to connect SongsDiscovered to a single slot?

src/devices/cddasongloader.h
lumip: Indeed. Fixed.

src/ripper/ripcddialog.cpp
lumip: This only re-emits the SongsDiscovered signal if songs were previously already loaded, however, it can also start a genuine loading process if the device wasn't yet opened/loaded. It has side effects on other consumers of the SongsDiscovered signal insofar as there will be a re-emission of that signal. However, the same signal will also happen if the disc was changed and consumers are therefore expected to (and do) treat this as an update to the song list in a reasonable way.


I was thinking of containing the song cache in CddaDevice, but that has the problem that CddaDevice cannot currently detect disc changes, which require an emptying of the cache, because CddaDevice is now the only instance that has access to the cdio object. The whole split between CddaSongLoader and CddaDevice is a bit problematic as they really heavily depend on each other to function properly but I haven't yet found a good solution to that.

src/devices/cddadevice.cpp
lumip: True. Any suggestions? Maybe `RequestSongs` or `AnnounceSongs`/`SignalSongs`?

src/ripper/ripcddialog.cpp
lumip: We could do that, but I think the logical separation of updating the track list and updating the metadata fields makes some sense as well and helps readability of the code. We could call `BuildTrackListTable` and `SetAlbumMetadata` from inside a new handler for SongsDiscovered, to avoid the multiple connections, though.

src/devices/cddadevice.cpp
jbroadus: Hmm. Maybe `UpdateFromLoader`?

lumip: But it's also not really updating anything, using just the cached information. The only thing it really does is (re)-emit the signal with the known song list (which gets updated automatically upon media change or on first read). So maybe just `EmitSongs`?

Thinking about it, it seems to indicate that using the signal as the main way of getting the song list to the consumers might not be the best design decision, because it forces this odd re-emission. I could change it so that the `CddaDevice` (and potentially also `CddaSongLoader`) just expose methods to get the song information simply emit a signal whenver that changes. Consumers can then just call the method when receiving the signal. That would also take care of the concern you have expressed in the discussion below, as the signal would not be re-emitted if a new consumer wanted to query the tracks.

jbroadus: Yeah, as long as the fetch is re-entrant, which I think it is.

lumip: I'm not quite sure it is. I have now done the following:
- protected access to the tracklist cache in CddaSongLoader with a mutex
- added method CddaDevice::songs (which forwards to CddaSongLoader::cached_tracks)
- removed CddaDevice::LoadSongs
- renamed the private CddaDevice::ForceLoadSongs back to CddaDevice::LoadSongs
- RipCDDialog now invokes CddaDevice::songs to get the song list from the device instead of provoking it to emit a signal

I didn't change the behavior of the existing signals, so those still provide the updated list of songs to the receiver.

src/devices/cddadevice.cpp
```diff
@@ -73,12 +71,23 @@ void CddaDevice::WatchForDiscChanges(bool watch) {
...
-  emit SongsDiscovered(songs);
   song_count_ = songs.size();
+  emit SongsDiscovered(songs);
+}
+
+void CddaDevice::SongsLoadingFinished() {
+  // Ensure that cdio_get_media_changed is cleared
+  // after songs are loaded, so we don't potentially
+  // re-read the same disc.
+  cdio_get_media_changed(cdio_);
```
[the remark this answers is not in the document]
lumip: The problem I was trying to address with this is that cdio_get_media_changed is returns true twice when a disc is inserted, with a bit of delay between both (I think they roughly correspond to "tray is closed" and "disc was detected/read/spun up/whatever"). There's no way to programmatically distinguish them and for the current code this means that the first time cdio_get_media_changed "fires" we start reading the disc, deactivating the disc change polling timer while doing so. After reading is complete, the timer gets reactivated and, due to the behavior described above, sees cdio_get_media_changed returned true, causing another reading of the disc. Therefore I decided to "clear" the cdio_get_media_changed response after reading to avoid this.

I can see your concern with what happens if the media changes during the load process. I tried to provoke this, but I was unable to as the reading is quite fast and I think the tray is locked while that is in progress. I could only get the drive to eject the disc before the read started (while it was still spinning up) or after the read completed (with a bit of a delay for spin down). I think it is near impossible to actually have a disc change in the time between the read completing and this reset of cdio_get_media_changed occuring and it wouldn't be a huge problem anyways (user would just have to cycle the tray once in case it happens).

That said, it just occured to me that I still need to check this without an internet connection since waiting for a response from musicbrainz may delay this long enough for a physical disc change to actually take place...

jbroadus (approved):

hatstand:
@lumip can you fix the conflicts here?

lumip:
I'm working on a rebase to resolve conflicts and nicefy the commits after the edits from the review. Expect that soon.

lumip:
@hatstand @jbroadus I took me a bit as I was busy otherwise, but I have now resolved all conflicts and squashed commits somewhat. I've also fixed a bug emerging from what appears to be a subtle race-condition with gstreamer that resulted in CddaSongLoader to skip reading tags for a track randomly in rare cases. Otherwise there are no substantial changes from the last review.

PR #7037, opened 2021-05-30
Consolidating access to cd drives in `CddaDevice`

### LINKED ISSUE #7023: Improvement: Update data in CD rip dialog when medium is changed

lumip:
Reduces the amount of separate `CdIo_t` instances held by Clementine for a CD device to one, which is maintained by the `CddaDevice` class. This is to address problems with holding multiple instances discussed in PR #7030. `CddaDevice` now also watches for disc changes and emits a signal if one is detected. This also closes #7023 and therefore makes PR #7024 unnecessary.

Changes to other classes:
- `ConnectedDevice::Refresh` now became obsolete and was removed; it was only used for polling for disc changes for `CddaDevice`
- `RipCDDialog` now requires a `CddaDevice` to be passed in at construction (from which it borrows the `CddaSongLoader` instance instead of constructing a separate one)
- `Ripper` requires access to a `CdIo_t` instance for reading from the CD. This is currently solved by passing in the instance held by `CddaDevice`. However, ultimately I'd like to use a gstreamer pipeline for ripping and remove direct cdio dependency from `Ripper` (already have a prototype implementation for that).


src/ui/mainwindow.cpp
```diff
@@ -2217,15 +2221,33 @@ void MainWindow::AddStreamAccepted() {
...
+            CddaDevice::url_schemes());
+    std::shared_ptr<ConnectedDevice> device;
+    if (!cd_device_infos.isEmpty()) {
+      DeviceInfo* device_info =
+          cd_device_infos[0];  // TODO: currently just picking the first cd
+                               // drive; what if there are several?
+      device = app_->device_manager()->Connect(
+          device_info);  // connect the device; will do nothing if already
+                         // connected
+    }
```
jbroadus, line 2235: Could we move all of this into the ripper dialog? This seems like something the mainwindow shouldn't need to know about. Also, if there are multiple CD devices, you could have a list of devices in the dialog.

src/ui/mainwindow.cpp
```diff
@@ -2217,15 +2221,33 @@ void MainWindow::AddStreamAccepted() {
...
+                            tr("Failed reading CD drive"));
+      cdio_fail.exec();
+      return;
+    }
+
+    std::shared_ptr<CddaDevice> cdda_device =
+        std::dynamic_pointer_cast<CddaDevice>(device);
+    Q_ASSERT(cdda_device);  // since we are requesting for cdda devices, the
+                            // cast must work; assert it
+    rip_cd_dialog_.reset(new RipCDDialog(std::move(cdda_device)));
```
jbroadus, line 2248: I don't think the move is necessary here.

src/ui/mainwindow.cpp
lumip: Done.

src/ui/mainwindow.cpp
lumip: Probably not, didn't hurt either, though. Now it's a moot point because mainwindow doesn't deal with CddaDevices anymore.

lumip:
I have added a drop down menu for selecting the CD drive at the top of the CD ripping dialog now:
cdripperdialog

Let me know if you have other preferences for the layout of this.

src/devices/devicemanager.cpp
```diff
@@ -348,6 +347,26 @@ DeviceInfo* DeviceManager::FindEquivalentDevice(DeviceInfo* info) const {
   return nullptr;
 }
 
+QList<DeviceInfo*> DeviceManager::FindDeviceByUrlSchemes(
```
jbroadus, line 350: Minor: Should this be `FindDevicesByUrlSchemes`?

src/devices/cddadevice.cpp
```diff
@@ -36,24 +39,46 @@ CddaDevice::CddaDevice(const QUrl& url, DeviceLister* lister,
           SLOT(SongsLoaded(SongList)));
   connect(this, SIGNAL(SongsDiscovered(SongList)), model_,
           SLOT(SongsDiscovered(SongList)));
+  //   connect(&cd_device_, SIGNAL(DiscChanged()), SLOT(DiscChangeDetected()));
+  cdio_ = cdio_open(url_.path().toLocal8Bit().constData(), DRIVER_DEVICE);
+  Q_ASSERT(cdio_);  // todo: error handling?
```
jbroadus, line 44: I still don't think that assertions are right for this. The cdio_open documentation states that it can return null on an error. What if, for example, the device open failed with EBUSY?
https://git.savannah.gnu.org/cgit/libcdio.git/tree/lib/driver/_cdio_generic.c#n135

src/devices/devicemanager.cpp
lumip: done

src/devices/cddadevice.cpp
lumip: Right, forgot to replace that with something proper. Done now.

lumip:
Something that I still need to fix is that currently creating a `CddaDevice` instance seems to block the main thread for a second or so while calling `cdio_open`. As there is no way to make this nice there currently without major rework of the ? DeviceManager` and other `ConnectedDevice` (sub)classes, I guess I have to abandon RAII and move that stuff back to the `Init` method. Then the whole class needs proper guards for method calls. No time to do it now, but it will come. Don't merge yet.

lumip:
Turns out the blocking I was seeing had nothing to do with initialisation, so much less drastic action was required. It was caused by the device change watchdog timer in `CddaDevice` blocking in the `cdio_get_media_changed` while `CddaSongLoader` was accessing the disc for reading. Fixed by not checking for disc changes while a disc is being read.

I think this is completed now, unless further things come up in review.

src/devices/cddadevice.cpp
```diff
@@ -36,24 +40,53 @@ CddaDevice::CddaDevice(const QUrl& url, DeviceLister* lister,
           SLOT(SongsLoaded(SongList)));
   connect(this, SIGNAL(SongsDiscovered(SongList)), model_,
           SLOT(SongsDiscovered(SongList)));
+  //   connect(&cd_device_, SIGNAL(DiscChanged()), SLOT(DiscChangeDetected()));
+  cdio_ = cdio_open(url_.path().toLocal8Bit().constData(), DRIVER_DEVICE);
+  if (!cdio_) {
+    throw DeviceError(url.toString(),
+                      "Cannot open device: cdio_open returned nullptr");
```
jbroadus, line 47: Exceptions aren't used for error handling anywhere else in the code base. It'll be up to @hatstand to decide if he wants to introduce them.

src/devices/cddadevice.cpp
```diff
@@ -36,24 +40,53 @@ CddaDevice::CddaDevice(const QUrl& url, DeviceLister* lister,
...
-void CddaDevice::Refresh() {
-  if (!cdda_song_loader_.HasChanged()) {
-    return;
-  }
-  Init();
+void CddaDevice::Init() { LoadSongs(); }
+
+CddaSongLoader* CddaDevice::loader() { return &cdda_song_loader_; }
+
+CdIo_t* CddaDevice::raw_cdio() { return cdio_; }
```
jbroadus, line 62: Have you found cdio documentation anywhere that discusses thread safety? It doesn't seem to be very well documented in that regard. It may be necessary to examine the code.

lumip: cdio does not seem to state anything about thread safety, but it seems to be blocking/synchronised via the system calls it does in the background.

src/devices/cddadevice.cpp
lumip: I'd argue it's the best way to do this here: There is no other way to pass out an error condition, neither in the constructor nor in the `Init` currently. So we could have an error code returned by `Init`, which would mean touching all classes implementing `ConnectedDevice`. Alternatively we could (re)introduce an `IsValid` method for `CddaDevice`. Checking that after instantiation would then make it essentially an exception in disguise (and also be more error prone, as it would be possible to forget checking that). We could also check validity every time the object would be about to be used, but at that point we're just looking for ways to make stuff complicated.

src/ripper/ripper.cpp
```diff
@@ -38,23 +38,23 @@ const char kWavFileTypeFormatChunk[] = "WAVEfmt ";
...
     : QObject(parent),
+      cdio_(cdio),
       transcoder_(new Transcoder(this)),
       cancel_requested_(false),
       finished_success_(0),
       finished_failed_(0),
       files_tagged_(0) {
-  cdio_ = cdio_open(NULL, DRIVER_UNKNOWN);
-
+  Q_ASSERT(cdio_);  // TODO: error handling
```
lumip, line 49: This one also need proper handling as of now; I'll wait for a decision on the exception matter before tackling this, though.

src/devices/cddadevice.cpp
hatstand: Sorry, please avoid exceptions.

lumip: alright.. done. Have modified `Init()` to return a boolean indicating success. Please see if this is fine.

src/devices/cddadevice.h
```diff
@@ -18,31 +18,43 @@
...
+  // Initializes the CddaDevice but does not open a handle to the device.
+  // Using code MUST call Init() after construction before calling and
+  // check that it returns true any other methods, to ensure that a valid
+  // device handle was correctly opened. Using class methods without
+  // a valid device handle is undefined behavior and might result in crashes.
   Q_INVOKABLE CddaDevice(const QUrl& url, DeviceLister* lister,
                          const QString& unique_id, DeviceManager* manager,
-                         Application* app, int database_id, bool first_time);
+                         Application* app, int database_id, bool first_time,
+                         bool watch_for_disc_changes = true);
```
jbroadus, line 44: Is there (or will there be) a use case where this isn't true?

src/ripper/ripper.cpp
jbroadus: If cdio_ is null here, then it's a Clementine bug, isn't it? In that case, I think the assert is fine.

src/ripper/ripcddialog.cpp
```diff
@@ -132,36 +128,65 @@ RipCDDialog::RipCDDialog(QWidget* parent)
...
 
+  // there is no device that is already connected; just select the first one
+  if (!cdda_devices_.isEmpty()) ui_->cd_drive_selection->setCurrentIndex(0);
+}
+
+void RipCDDialog::ClickedRipButton() {
+  Q_ASSERT(cdda_device_);
+
+  // create and connect Ripper instance for this task
+  Ripper* ripper = new Ripper(cdda_device_->raw_cdio(), this);
```
jbroadus, line 179: What happens to ripper if the media changes now? Does it fail and signal Finished?

lumip: Yes. Same as before.

src/devices/cddadevice.h
lumip: I guess not, can probably remove that from the constructor.

jbroadus: If you're planning on doing that, you could also start the timer from Init() and you wouldn't have to worry about the timer being called before initialization.

src/ripper/ripcddialog.cpp
```diff
@@ -132,36 +128,65 @@ RipCDDialog::RipCDDialog(QWidget* parent)
...
+  // add all devices to drop down selection
+  for (const DeviceInfo* device_info : cdda_devices_) {
+    ui_->cd_drive_selection->addItem(device_info->friendly_name_);
+  }
+  // ensure that selecting the first device below will emit a
+  // currentIndexChanged signal
+  ui_->cd_drive_selection->setCurrentIndex(-1);
+
+  connect(ui_->cd_drive_selection, SIGNAL(currentIndexChanged(int)),
+          SLOT(DeviceSelected(int)));
```
jbroadus, line 158: Does this get called on the initial selection, or do you need to click an item?

jbroadus (commented):
Overall, the change looks pretty good to me,though it ended up being larger than I thought. Are you planning on squashing some of those commits?

lumip:
I might look into combining some of the smaller commits and getting rid off ones that are effectively undone already, but I'm not very enthusiastic about spending time one that tbh.

src/devices/cddadevice.h
lumip: done

lumip:
I think all comments are now resolved. I'll have a rebase to group relevant changes together coming tomorrow and then we could merge this, if you agree.

src/ripper/ripcddialog.cpp
```diff
@@ -132,36 +128,65 @@ RipCDDialog::RipCDDialog(QWidget* parent)
...
+    if (device_info->device_) {
+      // found one!
+      ui_->cd_drive_selection->setCurrentIndex(i);
+      return;
     }
-    return;
   }
 
+  // there is no device that is already connected; just select the first one
+  if (!cdda_devices_.isEmpty()) ui_->cd_drive_selection->setCurrentIndex(0);
```
jbroadus, line 172: Have you had a change to try this on a machine without a drive? If not, I can try it on my laptop tonight.

jbroadus:
> I think all comments are now resolved. I'll have a rebase to group relevant changes together coming tomorrow and then we could merge this, if you agree.

It's looking good to me. You might want to get @hatstand 's feedback before you rebase.

src/ripper/ripcddialog.cpp
lumip: Unfortunately I did not.

lumip:
Got a rebased version [here](https://github.com/lumip/Clementine/commits/only_one_cdio_object_rebased) now. If that one is looking better, I can force push it over for this PR.

src/ripper/ripcddialog.cpp
jbroadus: I tried on my laptop and it works as I would expect. Nothing listed in the drive drop-down and the start option is disabled.

src/devices/cddadevice.h
```diff
@@ -18,31 +18,42 @@
...
   bool DeleteFromStorage(const MusicStorage::DeleteJob&) { return false; }
-
+  CddaSongLoader* loader();
+  // Access to the raw cdio device handle.
+  CdIo_t* raw_cdio();  // TODO: not ideal, but Ripper needs this currently
+  // Check whether a valid device handle was opened.
+  bool IsValid() const;
+  void WatchForDiscChanges(bool watch);
+
+  static const int kDiscChangePollingIntervalMs;
```
hatstand, line 56: I think this can be omitted from the header and be in an anonymous namespace in the `.cpp`

src/ripper/ripcddialog.cpp
```diff
@@ -52,12 +54,15 @@ const int kTrackDurationColumn = 3;
 const char* RipCDDialog::kSettingsGroup = "Transcoder";
 const int RipCDDialog::kMaxDestinationItems = 10;
 
-RipCDDialog::RipCDDialog(QWidget* parent)
+RipCDDialog::RipCDDialog(DeviceManager& device_manager, QWidget* parent)
```
hatstand, line 57: Should be a pointer or a const ref.

src/ripper/ripcddialog.cpp
```diff
@@ -241,11 +266,57 @@ void RipCDDialog::InvertSelection() {
...
+  ResetDialog();
+  if (device_index < 0)
+    return;  // Invalid selection, probably no devices around
+
+  Q_ASSERT(device_index < cdda_devices_.size());
+
+  DeviceInfo* device_info = cdda_devices_[device_index];
+  std::shared_ptr<ConnectedDevice> device =
+      device_manager_.Connect(device_info);
+  cdda_device_ = std::dynamic_pointer_cast<CddaDevice>(device);
```
hatstand, line 283: Can we avoid this dynamic cast?

src/ripper/ripcddialog.cpp
```diff
@@ -241,11 +266,57 @@ void RipCDDialog::InvertSelection() {
...
+          SLOT(AddAlbumMetadataFromMusicBrainz(SongList)));
+
+  // load songs from new SongLoader
+  loader_->LoadSongs();
+  rip_button_->setEnabled(true);
+}
+
+void RipCDDialog::Finished() {
+  SetWorking(false);
+  Ripper* ripper = dynamic_cast<Ripper*>(sender());
```
hatstand, line 312: Good to avoid `dynamic_cast` if possible. `NewClosure` is a good way to avoid needing `sender()` at all or alternatively use a `qobject_cast` if you really have to.

src/ripper/ripcddialog.h
```diff
@@ -85,8 +90,10 @@ class RipCDDialog : public QDialog {
   QPushButton* close_button_;
   QPushButton* rip_button_;
   std::unique_ptr<Ui_RipCDDialog> ui_;
-  Ripper* ripper_;
+  DeviceManager& device_manager_;
```
hatstand, line 93: Should be a pointer.

src/ripper/ripcddialog.cpp
jbroadus: The functor variant of connect can also be used with a lambda. It feels lighter weight to me.

src/ripper/ripcddialog.cpp
lumip: I don't really see how. At least not without modifying `DeviceManager` and adding a method that explicitely returns `CddaDevice`s - and that would probably just push the `dynamic_cast` into the `DeviceManager` unless we really have a separate store of the `CddaDevice`s there internally..

src/devices/cddadevice.h
lumip: I think `IsValid`  should be a public member as there is no other way to know whether any given instance of `CddaDevice` is properly set up.

`WatchForDiscChanges` modifies the instance state and it makes sense to me to keep it as a class method.

src/ripper/ripcddialog.h
lumip: done

src/ripper/ripcddialog.cpp
lumip: done with lambda

src/ripper/ripcddialog.cpp
lumip: done

src/devices/cddadevice.h
hatstand: I meant specifically `kDiscChangePollingIntervalMs` which I don't think needs to be here.

lumip:
Note that the failed test seems to have run into an issue in dependency installation, not anything with the actual code. Would it be possible to rerun that?

jbroadus:
> Note that the failed test seems to have run into an issue in dependency installation, not anything with the actual code. Would it be possible to rerun that?

I usually just amend the commit message on the most recent commit and force push to trigger a rebuild.

src/devices/cddadevice.h
lumip: I see. That's done.

Issue #7042, opened 2021-06-09
Playing .cue or single album, multi-track flac files - missing first 3 tracks


RDorville:
- Operating System: Ubuntu 20.04 LTS
- Clementine version: 1.41 rc1

When loading .cue or .flac (single file, mutiple track, embedded cue file), first 3 tracks are mssing.  Play starts from track 4, as displayed on the track window e.g. cue/flac has 12 tracks, only track 4~12 are displayed.
Verified with DeadBeef Player and the tracks are there.

### Expected behaviour / actual behaviour
Should display tracks 1-12 instead of 4-12.

### Steps to reproduce the problem (only for bugs)
Attached .cue file for checking (renamed to .cue.txt)
[Acoustic Alchemy - Roseland.cue.txt](https://github.com/clementine-player/Clementine/files/6621412/Acoustic.Alchemy.-.Roseland.cue.txt)

jbroadus:
Strange. It seems to be skipping everything with a timestamp < 10:00:00

jbroadus:
Okay, somebody used a regex to parse this that expects 2-3 digits in the minute space.
```(\\d{2,3}):(\\d{2}):(\\d{2})```
So, this works:
```00:00:00```
This does not:
```0:00:00```

jbroadus:
For reference, do you happen to know what program generated this playlist?

jbroadus:
Fixed here: https://github.com/clementine-player/Clementine/releases/tag/1.4.0rc1-637-gffdaeba09

PR #7044, opened 2021-06-12
cueparser: Be more lenient when parsing indexes.

jbroadus:
Most specs have mm:ss:ff as the index time format, but cue files have
been found with single digit minutes. e.g. 0:00:00 instead of 00:00:00

Reference: https://www.gnu.org/software/ccd2cue/manual/html_node/CUE-sheet-format.html

This is a fix for #7042


jbroadus:
Build is failing because we're grabbing a gpodder patch from from strawberrymusicplayer.org, but that's down at the moment. Should probably grab a local copy of that file.

PR #7054, opened 2021-07-02
De-duplicate my translator credit

fitojb:
Trivial change to fix Transifex’s mess.


hatstand:
This will probably get overwritten but we'll see what happens 🤷

fitojb:
You were right. :(

Issue #7040, opened 2021-06-04
Clementine 1.4 rc1 crashes with a segmentation fault


Gestas:
- [x] I checked the issue tracker for similar issues
- [x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [x] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information
- Operating System: 
```
NAME="Ubuntu"
VERSION="20.04.1 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.1 LTS"
VERSION_ID="20.04"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```
- Clementine version:  
```
1.4rc1
```

### Issue
Expected behaviour: Clementine updates all podcasts when I right click and select "Update All Podcasts".
Actual behaviour: Clementine crashes with a segmentation fault.

### Steps to reproduce the problem (only for bugs)
1. Have some podcast subscriptions. 
2. Start Clementine
3. Right click on "Podcasts"
4. Choose "Update All Podcasts"

### Troubleshooting steps - 
I started Clementine using `strace`. 
```
$ strace clementine --verbose; date +"%T.%3N";

# ... lots of output

# ... tail of output ->

stat("/etc/localtime", {st_mode=S_IFREG|0644, st_size=2836, ...}) = 0
stat("/etc/localtime", {st_mode=S_IFREG|0644, st_size=2836, ...}) = 0
stat("/etc/localtime", {st_mode=S_IFREG|0644, st_size=2836, ...}) = 0
stat("/etc/localtime", {st_mode=S_IFREG|0644, st_size=2836, ...}) = 0
stat("/etc/localtime", {st_mode=S_IFREG|0644, st_size=2836, ...}) = 0
stat("/etc/localtime", {st_mode=S_IFREG|0644, st_size=2836, ...}) = 0
stat("/etc/localtime", {st_mode=S_IFREG|0644, st_size=2836, ...}) = 0
stat("/etc/localtime", {st_mode=S_IFREG|0644, st_size=2836, ...}) = 0
stat("/etc/localtime", {st_mode=S_IFREG|0644, st_size=2836, ...}) = 0
stat("/etc/localtime", {st_mode=S_IFREG|0644, st_size=2836, ...}) = 0
[...]

PR #7070, opened 2021-07-26
internet/podcasts: Fix crash when updating podcasts.

jbroadus:
When a podcast is updated and the number of visible items is set in the
podcast settings, child items that disappear from the view, and are
deleted, are still referenced by the database id map.

Move the removal code from SubscriptionRemoved to a common method and
use that for this case.


jbroadus:
Related to #7040

smalltimer:
Hi, I think this might be leading to duplication of podcast entries when you update feeds!

Issue #7034, opened 2021-05-28
Potential Security Issue


JamieSlome:
Hello,

We recently received a vulnerability disclosure against your repository. I tried looking for a contactable e-mail in your repository, and couldn't find a defined security process.

If you would like me to send the details of the vulnerability disclosure to an e-mail or elsewhere, let me know, and will be happy to facilitate.

Otherwise, you can check out the advisory [here](https://huntr.dev/bounties/1-other-clementine-player/Clementine/). It is private to yourselves and the discloser.

Let me know if you have any questions, here to help.

-- Jamie from huntr.dev 🍰

lumip:
Hey @JamieSlome . I unfortunately cannot check the advisory following the link, but based on the history of the reporting user, I hazard a guess that this is about API keys being embedded in the code. If so, we are aware of that, but have decided that this is not critical enough to warrant the complications involved with fixing it.

PR #7071, opened 2021-08-10
Remove all hardcoded API keys and secrets from code

lumip:
And have them be re-injected into builds via github secrets. 
As API keys are supposed to be kept secret, they should not be publicly visible in code on github.

Accepting this PR means that the api keys need to be added as the following Github Secrets (and ideally new ones should be generated before that as the current ones have been exposes for a while now):
- BOX_SERVICE_CLIENT_SECRET
- DISCOGS_API_ACCESS_KEY
- DISCOGS_API_SECRET_KEY
- DROPBOX_CLIENT_SECRET
- FLICKR_API_KEY
- GOOGLE_DRIVE_CLIENT_SECRET
- LASTFM_API_KEY
- LASTFM_API_SECRET
- SONGKICK_API_KEY
- SPOTIFY_API_KEY

I believe this fixes the security issue alluded to by #7034 (based on checking past vulnerability reports from that user on that vulnerability hunting page).

### DIFF
--- a/.github/workflows/all.yml
+++ b/.github/workflows/all.yml
@@ -187,6 +187,16 @@ jobs:
       image: fedora:34
     env:
       RPM_BUILD_NCPUS: "2"
+      BOX_SERVICE_CLIENT_SECRET: ${{ secrets.BOX_SERVICE_CLIENT_SECRET }}
+      DISCOGS_API_ACCESS_KEY: ${{ secrets.DISCOGS_API_ACCESS_KEY }}
+      DISCOGS_API_SECRET_KEY: ${{ secrets.DISCOGS_API_SECRET_KEY }}
+      DROPBOX_CLIENT_SECRET: ${{ secrets.DROPBOX_CLIENT_SECRET }}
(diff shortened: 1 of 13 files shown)


jbroadus:
This is a good idea, but I think it will be a problem for maintainers of official distro packages.

lumip:
So, do you think there is a way of collaborating with those maintainers to push this through or do you want me to close this?

This wouldn't be a technical problem: The keys and tokens are read into the program from environment variables during the cmake configuration stage of the build, so it would be possible for the maintainers to set up the build environment accordingly if the keys/tokens are shared with them (or they could get dedicated key/tokens for the distro package). It could be a bit of a hassle for them to set it up appropriately (but shouldn't take any more effort after this one time setup), though.

hatstand:
Thanks for the work you've put in here but it doesn't really make sense to do this for a few reasons:
1. These secrets have been public for _years_
2. They would still be public and extractable from publicly distributed binaries
3. It makes it difficult for users to compile from source and get a usable configuration
4. Most of these providers are aware that these keys aren't really secret for desktop apps.

PR #7076, opened 2021-08-22
Enable to build on systems lacking execinfo.h

ismaell:
The Q_OS_UNIX macro is inappropriate for detection because many UNIX-like
platforms may lack backtrace support in the libc. E.g.: Darwin / Mac OS X,
Musl libc, OpenBSD, OpenIndiana.

### DIFF
--- a/ext/libclementine-common/CMakeLists.txt
+++ b/ext/libclementine-common/CMakeLists.txt
@@ -39,3 +39,6 @@ target_link_libraries(libclementine-common
   ${TAGLIB_LIBRARIES}
   ${CMAKE_THREAD_LIBS_INIT}
 )
+
+find_package(Backtrace)
+configure_file(core/conf_backtrace.h.in conf_backtrace.h)
--- a/ext/libclementine-common/core/logging.cpp
+++ b/ext/libclementine-common/core/logging.cpp
@@ -21,9 +21,7 @@
 #include <cxxabi.h>
 
 #include <QtGlobal>
-#ifdef Q_OS_UNIX
-#include <execinfo.h>
-#endif
+#include "conf_backtrace.h"
 
 #include <glib.h>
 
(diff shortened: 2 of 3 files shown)


jonaski:
I think it's better to use: `find_package(Backtrace)`

ismaell:
@jonaski  Thanks, I'll try that.

ismaell:
@jonaski is that enough?

ext/libclementine-common/core/logging.cpp
```diff
@@ -21,7 +21,7 @@
 #include <cxxabi.h>
 
 #include <QtGlobal>
-#ifdef Q_OS_UNIX
+#ifdef HAVE_BACKTRACE
 #include <execinfo.h>
```
jbroadus, line 25: `execinfo.h` also may not be correct. FindBacktrace recommends generating a file that includes `${Backtrace_HEADER}`
https://cmake.org/cmake/help/v3.0/module/FindBacktrace.html

ismaell: Yes, that's why I was asking... I went for the trivial change...

ismaell: Solved. I think.

ext/libclementine-common/CMakeLists.txt
```diff
@@ -39,3 +39,6 @@ target_link_libraries(libclementine-common
   ${TAGLIB_LIBRARIES}
   ${CMAKE_THREAD_LIBS_INIT}
 )
+
+find_package(Backtrace)
+configure_file(core/conf_backtrace.h.in conf_backtrace.h)
```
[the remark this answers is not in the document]
jbroadus: Oh yeah. I didn't notice that you had moved the ifdef.

jbroadus (approved):

Issue #7110, opened 2021-10-26
Overwrite transcoded files in source folder 


cmutts:
Hello Team, I have a huge library with about 10,000 files in source folder. I want to mass transcode all the files from sample rate 48000 hertz to  44100 hertz. I don't want the files renamed or given a new extension. I just want them replaced or overwritten with the new transcoded files. I can't see this option at all. Is it possible?

lumip:
This is currently not possible. I might add an option for this in the near future, though, this seems a nice feature to have.

lumip:
Hi @cmutts.  Happy to announce that the newest pre-release builds [link](https://github.com/clementine-player/Clementine/releases/tag/1.4.0rc1-787-g1aaf74788) now have the functionality you requested.

PR #7118, opened 2021-11-06
Added option to remove/replace originals in transcoder dialog.

lumip:
This behavior is enabled via a new checkbox in the transcoder dialog.
When enabled, transcoding will replace original files if the transcoded file would end up with the same path name (same format/extension and destination "alongside the originals"). In all other cases, the original files will simply be deleted after successful transcoding.
If transcoding of a file fails for any reason, the original will not be deleted/replaced.

Requested in #7110 .

### DIFF
--- a/src/transcoder/transcodedialog.cpp
+++ b/src/transcoder/transcodedialog.cpp
@@ -98,6 +98,9 @@ TranscodeDialog::TranscodeDialog(QWidget* parent)
     }
   }
 
+  ui_->remove_original->setChecked(
+      s.value("overwrite_existing", false).toBool());
+
   // Add a start button
   start_button_ = ui_->button_box->addButton(tr("Start transcoding"),
                                              QDialogButtonBox::ActionRole);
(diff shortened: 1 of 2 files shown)


src/transcoder/transcodedialog.cpp
```diff
@@ -205,6 +209,20 @@ void TranscodeDialog::JobComplete(const QString& input, const QString& output,
 
   UpdateStatusText();
   UpdateProgress();
+
+  bool overwrite_existing = ui_->remove_original->isChecked();
+  QFileInfo input_fileinfo(input);
+  QFileInfo output_fileinfo(output);
+  bool same_extension = input_fileinfo.suffix() == output_fileinfo.suffix();
+  bool same_path =
+      input_fileinfo.absolutePath() == output_fileinfo.absolutePath();
```
jbroadus, line 218: Could most of this be moved into the if block below?

lumip: It could and has been

jbroadus (approved):

lumip:
@hatstand this one is also ready to be merged and closes #7110

Issue #7120, opened 2021-11-08
gstreamer bus watch issues


lumip:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [x] I checked the issue tracker for similar issues
- [ ] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [x] I tried the latest Clementine build

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Kubuntu 20.04
- Clementine version: 1.4.0rc1-759-gd033b38c4

### Expected behaviour / actual behaviour
With `export GST_DEBUG=3`
the following is output after a song quits playing
```
ERROR GST_BUS gstbus.c:1087:gst_bus_remove_watch:<bus8> no bus watch was present
```
@jonaski reported this happening first after merging PR #7101, which was addressing another issue caused by an attempt to remove the bus watch.

I had previously identified the cause as the following:
Since the bus watch callback was implemented to return false, it was removed from the bus after its first invocation (cf. "The watch can be removed using gst_bus_remove_watch or by returning FALSE from func.", [Gstreamer docs](https://gstreamer.freedesktop.org/documentation/gstreamer/gstbus.html?gi-language=c#gst_bus_add_watch)). This not only causes the above error, but also means we miss all but the first message emitted on the bus.
I proposed PR #7112 that instead returned true from the callback, which fixed the above issue but caused crashes/hanging for several users (apparently mostly on Arch Linux and Linux Mint) (cf. #7115) and was subsequently reverted (PR #7116).

The above error therefore is again present. Unfortunately, the gstreamer docs don't seem to give much information what the expected behavior for the callback would be and what could be the cause of the crashes in the changeset of the reverted PR.

Issue #7115, opened 2021-11-02
GStreamer-CRITICAL: Trying to dispose element source, but it is in PAUSED instead of the NULL state.


AllKind:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [x ] I checked the issue tracker for similar issues
- [ ] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [x ] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Linux Mint 19.3
- Clementine version: 1.4.0rc1-756-ga25887be5

### Expected behaviour / actual behaviour
Clementine should not freeze just playing local files.

### Steps to reproduce the problem (only for bugs)
Not sure how to reproduce, happend to me three times today. It just freezes, while you can hear the currently playing song until it finishes.
I think just by playing local files sequentially.

This is shown in the console:

> 17:20:32.912 DEBUG MainWindow:1596                  position: 310 , scrobble point: 315 , lastfm status: 0 , play count point: 315 , is local library item: true , playlist have incremented playcount:  false
> 17:20:36.982 DEBUG TagReader:1030                   Saving song rating tags to "/storage/pub/music/beets/albums/Salsa/Afro-Cuban_All_Stars/A_Toda_Cuba_le_Gusta/05-Afro-Cuban_All_Stars-Fiesta_de_la_Rumba.flac"
> 17:20:36.987 DEBUG MessageReply<MessageType>:89     Releasing ID 96 (finished)
> 17:20:36.987 DEBUG LibraryWatcher:707               Subdir "/storage/pub/music/beets/albums/Salsa/Afro-Cuban_All_Stars/A_Toda_Cuba_le_Gusta" changed under directory "/storage/pub/music" id 1
> 17:20:42.912 DEBUG MainWindow:1596                  position: 320 , scrobble point: 315 , lastfm status: 0 , play count point: 315 , is local library item: true , playlist have incremented playcount:  true
> 17:20:43.409 DEBUG LibraryWatcher:397               "/storage/pub/music/beets/albums/Salsa/Afro-Cuban_All_Stars/A_Toda_Cuba_le_Gusta/05-Afro-Cuban_All_Stars-Fiesta_de_la_Rumba.flac" changed
> 17:20:43.409 DEBUG _MessageReplyBase:24             Waiting on ID 97
[...]

PR #7121, opened 2021-11-10
Fixing gst bus watch issues

lumip:
Second attempt for fixing #7120 (after the initial PR addressing that was reverted) while also addressing the issue of #7115 .

I did some digging and the problem that arose from the initial pull request seems to have been caused by replacing the uridecodebin in `GstEnginePipeline::TransitionToNext`. After removing the old uridecodebin from the pipeline, its reference count dropped to zero before its state was set to `GST_STATE_NULL`, which resulted in the crash reported in #7115.
This behavior is documented in the [docs for `gst_bin_remove`](https://gstreamer.freedesktop.org/documentation/gstreamer/gstbin.html?gi-language=c#gst_bin_remove): "Removes the element from the bin, unparenting it as well. Unparenting the element means that the element will be dereferenced, so if the bin holds the only reference to the element, the element will be freed in the process of removing it from the bin. If you want the element to still exist after removing, you need to call gst_object_ref before removing it from the bin."

I added a fix for this that increments the reference count on the old decodebin before removing it from the pipeline and accordingly modified `GstElementDeleter::DeleteElement` to decrement the reference count after setting its state to `GST_STATE_NULL`.

Now, the real question is why this was not a problem earlier. I have monitored the reference counts of all decodebins and noticed that previously they were never reaching zero. The change of the return value of `BusCallback` changed this. I'm not sure why this is but the current new behavior seems to be the correct one. I haven't had the opportunity to look into this in more depth, so if someone could verify that there was a memory leak of decodebins earlier and give an intuition why changing the return value of `BusCallback` might have solved this, that would be very helpful.

### DIFF
--- a/src/engines/gstelementdeleter.cpp
+++ b/src/engines/gstelementdeleter.cpp
@@ -28,4 +28,5 @@ void GstElementDeleter::DeleteElementLater(GstElement* element) {
 
 void GstElementDeleter::DeleteElement(GstElement* element) {
   gst_element_set_state(element, GST_STATE_NULL);
+  gst_object_unref(element);
 }
--- a/src/engines/gstenginepipeline.cpp
+++ b/src/engines/gstenginepipeline.cpp
@@ -613,7 +613,7 @@ gboolean GstEnginePipeline::BusCallback(GstBus*, GstMessage* msg,
       break;
   }
 
-  return FALSE;
+  return TRUE;
 }
 
 GstBusSyncReply GstEnginePipeline::BusCallbackSync(GstBus*, GstMessage* msg,
--- a/src/engines/gstenginepipeline.cpp
+++ b/src/engines/gstenginepipeline.cpp
@@ -1146,6 +1146,7 @@ void GstEnginePipeline::SourceSetupCallback(GstURIDecodeBin* bin,
 
 void GstEnginePipeline::TransitionToNext() {
   GstElement* old_decode_bin = uridecodebin_;
+  gst_object_ref(old_decode_bin);
 
   ignore_tags_ = true;
 


lumip:
@AllKind @trougnouf @mohamedation @weearc @d10sfan @heldbrendel @ivovegter @shaygover @eris-price 
If you have time and opportunity, could you be so kind as to check whether this works for you (i.e. the crashes no longer occur)? (and sorry for my earlier mess-up with this)

trougnouf:
> @AllKind @trougnouf @mohamedation @weearc @d10sfan @heldbrendel @ivovegter @shaygover @eris-price If you have time and opportunity, could you be so kind as to check whether this works for you (i.e. the crashes no longer occur)? (and sorry for my earlier mess-up with this)

Yes :) I've been on clementine 1.4.0rc1+759+gd033b38c4 with no issue, thank you!

lumip:
> > @AllKind @trougnouf @mohamedation @weearc @d10sfan @heldbrendel @ivovegter @shaygover @eris-price If you have time and opportunity, could you be so kind as to check whether this works for you (i.e. the crashes no longer occur)? (and sorry for my earlier mess-up with this)
> 
> Yes :) I've been on clementine 1.4.0rc1+759+gd033b38c4 with no issue, thank you!

I meant the pending changes in this pull request, which are not included in 1.40rc1+759 yet ;)

trougnouf:
Sorry I thought I was on the bug report :s
My connection is being too slow to download today. If anyone would like to test on Arch, here is the [PKGBUILD](https://github.com/clementine-player/Clementine/files/7515888/PKGBUILD.txt)
:

trougnouf:
I got it :) So far no problem when skipping close to the end of songs and waiting for the next to play.

I will report back here if I encounter any issue (which would be at most 24 hours since I use Clementine all the time I'm awake)

weearc:
@lumip Hi, I use the latest commit from https://github.com/lumip/Clementine/tree/fixing_gst_bus_watch_issues. Sometimes I got terminal output like this:
```
22:16:57.421 DEBUG GstEnginePipeline:896            Ignoring native format since pipeline is already running.
22:17:02.548 DEBUG GstEnginePipeline:657            New segment started, EOS will signal on next buffer discontinuity
22:17:02.548 DEBUG GstEnginePipeline:1039           Buffer discontinuity - emitting EOS
```
But it actually works perfectly as before. Thank you for your quick fix!

AllKind:
> @AllKind @trougnouf @mohamedation @weearc @d10sfan @heldbrendel @ivovegter @shaygover @eris-price If you have time and opportunity, could you be so kind as to check whether this works for you (i.e. the crashes no longer occur)? (and sorry for my earlier mess-up with this)

I don't build from source, so I don't think I can check. But of course thank you for your work!

lumip:
> > @AllKind @trougnouf @mohamedation @weearc @d10sfan @heldbrendel @ivovegter @shaygover @eris-price If you have time and opportunity, could you be so kind as to check whether this works for you (i.e. the crashes no longer occur)? (and sorry for my earlier mess-up with this)
> 
> I don't build from source, so I don't think I can check. But of course thank you for your work!

Hi @AllKind , if you are interested, you can download binaries from the github build action for this change from [here](https://github.com/clementine-player/Clementine/actions/runs/1445873498) (scroll down to the bottom of the page to the list titled "Artifacts" and pick the one you need). However, since we've already got some reports that this is working, there's no urgent need for you to do so.

AllKind:
Ok, did that. Seems to work with the bionic build on Linux Mint 19.3.
Thanks!

Issue #5239, opened 2016-02-05
unavailable files in XSPF playlists are not imported


dev-jam:
When opening a XSPF playlist, unavailable files are not imported. This results in not knowing what files/songs are missing. 
PLS and M3U files do import unavailable files and are greyed out when this is selected in the settings.

Additionally, it would be nice to have a 'find' feature when right clicking an unavailable file to search the song in your library. Or an option to drag and drop the file/song in the search box of the media library.  

Using clementine 1.3rc1 on debian jessie. Same problem occurs with the official debian version which has XSPF formatting problems.

Ferroin:
+1

At the very least we need to return an error showing that stuff was dropped from the playlist, ideally with a count of songs dropped.  In an ideal situation, we'd have an option to auto-match the title and other metadata from the playlist (if present) to songs in the library, and have this for any playlist format which provides more than just a location.

For what it's worth, I'm pretty certain that WPL playlists import unavailable songs as well, so it looks like it's just XSPF format that has this issue.

PR #7127, opened 2021-11-26
Playlist Parsers: Parse all (even invalid) playlist entries.

smithjd15:
There are a number of different ways to rectify unfound entries if they aren't silently dropped.


smithjd15:
Closes https://github.com/clementine-player/Clementine/issues/7135.

smithjd15:
Closes https://github.com/clementine-player/Clementine/issues/5239.

smithjd15:
CUE Parser: Avoid overriding song metadata if it's already in the library, AND if the cue is not single-file.

PR #7129, opened 2021-11-27
Make background steams build-optional

xif-fr:
Added a build option `ENABLE_BACKGROUND_STREAMS`. When disabled, removes background streams (hypnotoad, rain, enterprise) code from the UI and engine with a bunch of `#ifdef`. Stays enabled by default. Both option values have been tested.

This is simply to make it possible to remove these potentially unwanted features.

### DIFF
--- a/src/engines/gstengine.cpp
+++ b/src/engines/gstengine.cpp
@@ -90,6 +90,8 @@ const char* GstEngine::kAutoSink = "autoaudiosink";
 const char* GstEngine::kOutFormatDetect = "";
 const char* GstEngine::kOutFormatS16LE = "S16LE";
 const char* GstEngine::kOutFormatF32LE = "F32LE";
+
+#ifdef HAVE_BACKGROUND_STREAMS
 const char* GstEngine::kHypnotoadPipeline =
     "audiotestsrc wave=6 ! "
     "audioecho intensity=1 delay=50000000 ! "
@@ -100,6 +102,7 @@ const char* GstEngine::kHypnotoadPipeline =
 const char* GstEngine::kEnterprisePipeline =
     "audiotestsrc wave=5 ! "
     "audiocheblimit mode=0 cutoff=120";
+#endif // HAVE_BACKGROUND_STREAMS
 
 GstEngine::GstEngine(Application* app)
     : Engine::Base(),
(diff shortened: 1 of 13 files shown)


src/ui/mainwindow.cpp
```diff
@@ -495,9 +501,15 @@ MainWindow::MainWindow(Application* app, SystemTrayIcon* tray_icon, OSD* osd,
...
           SIGNAL(NewDebugConsole(Console*)));
 
+#ifdef HAVE_BACKGROUND_STREAMS
   background_streams_->AddAction("Rain", ui_->action_rain);
   background_streams_->AddAction("Hypnotoad", ui_->action_hypnotoad);
   background_streams_->AddAction("Make it so!", ui_->action_enterprise);
+#else
+  ui_->action_rain->setVisible(false);
+  ui_->action_hypnotoad->setVisible(false);
+  ui_->action_enterprise->setVisible(false);
```
jbroadus, line 511: Could these be removed from mainwindow.ui and added to the extras menu dynamically?

src/engines/gstengine.cpp
```diff
@@ -860,7 +864,9 @@ shared_ptr<GstEnginePipeline> GstEngine::CreatePipeline(
       qLog(Error) << "Could not initialize pipeline" << kEnterprisePipeline;
       ret.reset();
     }
-  } else {
+  } else 
+#endif // HAVE_BACKGROUND_STREAMS
```
jbroadus, line 868: Maybe cleaner to return in the two if blocks above and get rid of the else?

src/ui/mainwindow.cpp
xif-fr: It would be a cleaner solution indeed, but I'm not sure how to deal with the translations of the menus, given that the translations are managed automatically.

src/engines/gstengine.cpp
xif-fr: Done.

src/ui/mainwindow.cpp
jbroadus: I think you can just call tr() like this: https://github.com/clementine-player/Clementine/blob/master/src/ui/mainwindow.cpp#L666

xif-fr: Done. I'm still unsure about the translation files because there are remaining `#: ../bin/src/ui_mainwindow.h:748` which don't match (now it's used in `mainwindow.cpp`). I don't know if it could break something. At least it compiles and works fine.

PR #7140, opened 2022-01-13
Delete README-TURKISH.md

kryger:
What was the motivation for singling out Turkish for translation? I haven't seen this "idiom" in any other github project. The fact that it's already outdated after a single update to the main README (#edb8c3b4ed8f976c6 not reflected in the Turkish version) proves it's not a good idea, clearly it's not guaranteed to maintained as part of the mainline and could cause confusion.
In 2022 I'd assume anyone determined enough to find and visit a project's page on github would know how to translate the page into their own language if they need to.


hatstand:
It was useful for someone 🤷

PR #7142, opened 2022-01-14
Remove the out-of-place Easter Egg

kryger:
Discovered while trying to understand how the filtering components are interconnected, this felt a bit surprising. Removing for the sake of clarity - unless it's used for testing or something?


cmorty:
Isn't this in the 1.3 release notes: https://github.com/clementine-player/Clementine/blob/master/Changelog

hatstand:
> We apologise for the inconvenience.

PR #7148, opened 2022-01-23
Regular progress bar updates for CD ripping.

lumip:
Previously the progress bar of the CD ripping dialog would only update
after a track completed, now it gets updated continuously during the
ripping process.

### DIFF
--- a/src/ripper/ripcddialog.cpp
+++ b/src/ripper/ripcddialog.cpp
@@ -217,16 +217,16 @@ void RipCDDialog::ClickedRipButton() {
   connect(cancel_button_, SIGNAL(clicked()), ripper, SLOT(Cancel()));
 
   connect(ripper, &Ripper::Finished, this, [this, ripper]() {
-    this->Finished(*ripper, /*progress_to_display = */ 1.0f);
+    this->Finished(ripper, /*progress_to_display = */ 1.0f);
   });
   connect(ripper, &Ripper::Cancelled, this, [this, ripper]() {
-    this->Finished(*ripper, /*progress_to_display = */ 0.0f);
+    this->Finished(ripper, /*progress_to_display = */ 0.0f);
   });
 
   ui_->progress_bar->setRange(0, 100);
   transcoding_progress_timer_connection_ =
       connect(&transcoding_progress_timer_, &QTimer::timeout, this,
-              [this, ripper]() { this->TranscodingProgressTimeout(*ripper); });
+              [this, ripper]() { this->TranscodingProgressTimeout(ripper); });
 
   // Add tracks and album information to the ripper.
   ripper->ClearTracks();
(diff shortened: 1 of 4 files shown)


lumip:
@jbroadus If you have the time, might I ask for another PR review? :)

src/ripper/ripcddialog.cpp
```diff
@@ -211,15 +213,24 @@ void RipCDDialog::ClickedRipButton() {
 
   // create and connect Ripper instance for this task
   Ripper* ripper = new Ripper(cdda_device_->song_count(), this);
+  if (!ripper) {
```
jbroadus, line 216: I don't think this can ever happen. If enabled, the assert in the Ripper constructor would call abort.

src/ripper/ripcddialog.cpp
```diff
@@ -243,6 +254,7 @@ void RipCDDialog::ClickedRipButton() {
 
   SetWorking(true);
   ripper->Start();
+  transcoding_progress_timer_.start(kTranscodingProgressIntervalMs);
```
jbroadus, line 257: Technically, there's a race condition here, though you should never hit it with the 500ms timer. If no tracks were added, the timer callback would call Ripper::GetProgress and divide by zero.

lumip: Good catch! I added a check in `Ripper::GetProgress` to prevent division by zero now.

src/ripper/ripcddialog.cpp
lumip: You are of course right, for some stupid reason I was scared of new returning null, but that can't happen... Removed the check now.

lumip:
Another thing about this is that I have moved the active reporting of progress via signals completely out of `Ripper` and instead made `RipCDDialog` poll for progress. This is now similar to how `Transcoder` and `TranscodeDialog` interact - before I was bugged by how differently this was handled. An alternative would be to let introduce progress reporting signals to `Transcoder` (which `Ripper` completely relies on to do the heavy lifting), which would remove the need for the timers and polling in the dialog classes... What do you think?

jbroadus (approved):

src/ripper/ripcddialog.cpp
```diff
@@ -343,21 +351,13 @@ void RipCDDialog::DeviceSelected(int device_index) {
           SLOT(SongsLoaded(SongList)));
 }
 
-void RipCDDialog::Finished(Ripper* ripper) {
+void RipCDDialog::Finished(Ripper& ripper, float progress_to_display) {
```
hatstand, line 354: Please follow the style guide and only use const references or pointers for non-primitive parameters. https://google.github.io/styleguide/cppguide.html

lumip: The guide states

> Non-optional input parameters should usually be values or const references, while **non-optional** output and **input/output parameters should usually be references** (which cannot be null). 

so that's what I did here. I can't make it const since the callback must clean up the instance using the `deleteLater` function, which is not const.

hatstand: They've changed the style guide since I last read it :-)

In the interest of maintaining a consistent style throughout the codebase I'd still prefer it if this were changed to `Ripper*`. We don't use non-const references anywhere else.

lumip: Done :)

PR #7204, opened 2022-08-18
Rewrite the Block Analyzer to improve performance.

areading:
The block analyzer was doing lots of repeated, out-of-order blits to the
widget's canvas. To improve performance and reduce CPU usage, this has
been rewritten to generate the canvas contents using only a single buffer.
Cache thrashing has been greatly reduced by writing to memory only
sequentially and in one single write pass. Further, the raw format is
now guaranteed to be in a format efficient for Qt.

The results are visually identical to what they were previously, but
result in a CPU usage reduction between 2 and 6 percent depending on refresh
rate and Psychadelic Mode value. In particular, there used to be a ~3 percent
overhead for Psychadelic Mode, and this has been eliminated.

The specific details of the block analyzer and explanations for how it works
(and used to work) have been documented via fairly extensive comments
in blockanalyzer.cpp.

### DIFF
--- a/src/analyzers/analyzerbase.cpp
+++ b/src/analyzers/analyzerbase.cpp
@@ -65,13 +65,15 @@ Analyzer::Base::Base(QWidget* parent, uint scopeSize)
       timeout_(40),  // msec
       fht_(new FHT(scopeSize)),
       engine_(nullptr),
-      lastScope_(512),
+      lastScope_(),
       new_frame_(false),
       is_playing_(false),
       barkband_table_(),
       prev_color_index_(0),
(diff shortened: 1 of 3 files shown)


gamedevsam:
Exceptional.

areading:
Thank you kindly.

This should be ready to review and/or merging. The couple force-pushes were for clang-format to fix the linter check, and to fix a sanity warning under gcc 12 with certain optimizer settings. All build checks should pass cleanly after merging #7183  (or whatever ends up being appropriate to address the Hirsute image being removed).

PR #7225, opened 2022-10-26
Replace play and pause with the single play/pause action in desktop file

adem4ik:
The idea is from Strawberry - https://github.com/strawberrymusicplayer/strawberry/commit/18023d2e18823cb743d079e692dfebe3be5e15c3

Task bar menu before / after:

Translations were mostly picked from the Pause action, a few of them are from SMPlayer.


fitojb:
Thank you for taking care of the translations!

PR #7238, opened 2022-12-21
Fix build: add zlib to deps where it's used 

DarthGandalf:
Ref https://bugs.gentoo.org/887105

### DIFF
--- a/3rdparty/qtiocompressor/CMakeLists.txt
+++ b/3rdparty/qtiocompressor/CMakeLists.txt
@@ -15,4 +15,4 @@ ADD_LIBRARY(qtiocompressor STATIC
     ${IOCOMPRESSOR-SOURCES-MOC}
 )
 
-target_link_libraries(qtiocompressor Qt5::Core)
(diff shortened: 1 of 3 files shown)


hatstand:
If zlib is only required by Qt5Network then shouldn't this be pulled in via the various Qt5 cmake rules?

DarthGandalf:
https://github.com/clementine-player/Clementine/blob/9ef681b0e9e945c662905c1ffc5e179dd8e04d3c/3rdparty/qtiocompressor/qtiocompressor.cpp#L48 requires it too

hatstand:
> https://github.com/clementine-player/Clementine/blob/9ef681b0e9e945c662905c1ffc5e179dd8e04d3c/3rdparty/qtiocompressor/qtiocompressor.cpp#L48
> 
> requires it too

I think adding `z` to the `target_link_libraries()` in that directory's `CMakeLists.txt` would fix that. It's better than doing it globally as it should take care of the ordering too I think.

DarthGandalf:
Yes, I agree, that's cleaner. Done.

PR #7342, opened 2024-10-02
build: remove bad double negative when discovering 'git'

muggenhor:
This automatically works because [CMake's `if` statement](https://cmake.org/cmake/help/latest/command/if.html#constant) considers variables which' content ends with "-NOTFOUND" to be FALSE.

The previous version however was first tried as a variable which didn't exist. As a result after that it's interpreted as a string (not a variable anymore). So it skips the "-NOTFOUND" check and only does a "string not empty" check. Which, as this is a string constant, is always true.

Of course this only fixes the check. It doesn't address the fact that the remaining code below depends on `GIT_REV` having some content. So maybe instead [`find_program`](https://cmake.org/cmake/help/latest/command/find_program.html) should be called with the `REQUIRED` flag?


hatstand:
> Of course this only fixes the check. It doesn't address the fact that the remaining code below depends on GIT_REV having some content. So maybe instead [find_program](https://cmake.org/cmake/help/latest/command/find_program.html) should be called with the REQUIRED flag?

`git` isn't intended to be required here, so that builds from source tars work without it.

PR #7356, opened 2024-10-31
Make git revision detection optional

### LINKED ISSUE #7355: Tag 1.4.1 fails to build from source tarball
### Before posting

### System information

- Operating System: Slackware64 15.0
- Clementine version: 1.4.1 tag ( a809096)

Expected: Tarballs tagged as a release version and downloaded via github tags should build without issue.

Actual: CMake configure fails when using the tagged 1.4.1 "release" due to forcing detecting a git revision in cmake/Version.cmake, tracing the breakage to commmit 5e5b888.

I could not find a build flag that allowed me to bypass the git revision detection (but I'm not super strong with cmake).

Download the 1.4.1 tag source tarball
`wget --content-disposition https://github.com/clementine-player/Clementine/archive/refs/tags/1.4.1/Clementine-1.4.1.tar.gz`
Extract source
`tar xvf Clementine-1.4.1.tar.gz`
cd into source, create and cd to `build/` dir and run cmake
`cd Clementine-1.4.1; mkdir build; cd build; cmake ..`

Build will fail with
```
fatal: not a git repository (or any of the parent directories): .git
CMake Error at cmake/Version.cmake:136 (message):
  git describe failed with code 128:
Call Stack (most recent call first):
  CMakeLists.txt:10 (include)


CMake Error at cmake/Version.cmake:141 (string):
  string sub-command REGEX, mode REPLACE needs at least 6 arguments total to
  command.
Call Stack (most recent call first):
  CMakeLists.txt:10 (include)


CMake Error at cmake/Version.cmake:145 (message):
  Failed to parse git revision string ''
Call Stack (most recent call first):
  CMakeLists.txt:10 (include)


-- Configuring incomplete, errors occurred!
See also "/tmp/SBo/Clementine-1.4.1/build/CMakeFiles/CMakeOutput.log".
```

I have an incoming PR that provides a way to bypass this, however, it requires changing the INCLUDE_GIT_REVISION variable within that file to OFF, which I was doing via a sed command in my build script.

---

Also, unrelated to this issue, the 1.4.1 tagged tarball (and the master branch) still have cmake/Version.cmake with versioned with 1.4.0rc2.

bassmadrigal:
With the [tagged 1.4.1](https://github.com/clementine-player/Clementine/releases/tag/1.4.1) source tarball downloaded from github, cmake fails due to forcing detection of a git revision, which doesn't exist from the github generated tarball.

This fix allows one to toggle that using the INCLUDE_GIT_REVISION variable at the top of cmake/Version.cmake, which if set to OFF will skip over git revision detection, allowing the user to complete cmake.

(Currently that INCLUDE_GIT_REVISION variable would need to be changed for a tagged release or have the user change it via sed or similar.)

Potentially fixes clementine-player/Clementine#7355

### DIFF
--- a/cmake/Version.cmake
+++ b/cmake/Version.cmake
@@ -119,39 +119,41 @@ else(CLEMENTINE_VERSION_PRERELEASE)
 endif(CLEMENTINE_VERSION_PRERELEASE)
 
 # Add git revision
-if(FORCE_GIT_REVISION)
-  set(GIT_REV ${FORCE_GIT_REVISION})
-  set(GIT_INFO_RESULT 0)
-else(FORCE_GIT_REVISION)
-  find_program(GIT_EXECUTABLE git)
-  message(STATUS "Found git: ${GIT_EXECUTABLE}")
+if(INCLUDE_GIT_REVISION)
+  if(FORCE_GIT_REVISION)
+    set(GIT_REV ${FORCE_GIT_REVISION})
+    set(GIT_INFO_RESULT 0)
+  else(FORCE_GIT_REVISION)
+    find_program(GIT_EXECUTABLE git)
+    message(STATUS "Found git: ${GIT_EXECUTABLE}")
 
-  if(NOT GIT_EXECUTABLE-NOTFOUND)
-    execute_process(COMMAND ${GIT_EXECUTABLE} describe
-        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
-        RESULT_VARIABLE GIT_INFO_RESULT
-        OUTPUT_VARIABLE GIT_REV
-        OUTPUT_STRIP_TRAILING_WHITESPACE)
-    if(NOT ${GIT_INFO_RESULT} EQUAL 0)
-      message(SEND_ERROR "git describe failed with code ${GIT_INFO_RESULT}: ${GIT_REV}")
+    if(NOT GIT_EXECUTABLE-NOTFOUND)
+      execute_process(COMMAND ${GIT_EXECUTABLE} describe
+          WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+          RESULT_VARIABLE GIT_INFO_RESULT
+          OUTPUT_VARIABLE GIT_REV
+          OUTPUT_STRIP_TRAILING_WHITESPACE)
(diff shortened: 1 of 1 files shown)


cmake/Version.cmake
```diff
@@ -119,39 +119,41 @@ else(CLEMENTINE_VERSION_PRERELEASE)
...
-      message(SEND_ERROR "git describe failed with code ${GIT_INFO_RESULT}: ${GIT_REV}")
+if(INCLUDE_GIT_REVISION)
+  if(FORCE_GIT_REVISION)
+    set(GIT_REV ${FORCE_GIT_REVISION})
+    set(GIT_INFO_RESULT 0)
+  else(FORCE_GIT_REVISION)
+    find_program(GIT_EXECUTABLE git)
+    message(STATUS "Found git: ${GIT_EXECUTABLE}")
+
+    if(NOT GIT_EXECUTABLE-NOTFOUND)
```
muggenhor, line 130: This still suffers from the problem I was attempting to fix in #7342. I.e. it shouldn't check for a "NOTFOUND" variable. That variable doesn't exist so this is `NOT FALSE` i.e. a tautology (always true, making the condition redundant).

It should check for the *existence* of `GIT_EXECUTABLE` instead:
```suggestion
    if(GIT_EXECUTABLE)
```

See #7342's explanation for more details.

bassmadrigal: This may well still cause a problem with certain builds, but it wasn't something that would ever be checked with the release tarballs I'm focusing on. With release tarballs, no git information will be included in the tarball, so INCLUDE_GIT_REVISION should be changed to OFF in those tarballs, and, if my changes are included, would skip over any of the git checking, never running into the check you mentioned.

Since release tarballs generated by GitHub don't include the .git/ folder, they will always fail with the current detection methods.

I haven't looked into the git detection at all since I am the maintainer of this package for SlackBuilds.org and they don't support git cloning when building packages. So your change might still be needed, but it was not looked at in this PR. That line only changed due to indentation when adding the `if(INCLUDE_GIT_REVISION)` statement before everything else.

vincele: This is easily seen with the "Hide whitespace" view option.
The PR LGTM.

PR #7402, opened 2025-07-07
Bump CMake minimum versions to >3.30

S4enL0ng:
It helps working with CMake >4.0 and also is working on CMake 3.31.7.

Any Suggestions to fix some things are welcome.


DarthGandalf:
why 3.30 instead of 3.10?

Justinzobel:
Ping.

CMakeLists.txt
```diff
@@ -1,8 +1,6 @@
-cmake_minimum_required(VERSION 3.13.0)
+cmake_minimum_required(VERSION 3.30.0)
```
vincele: Why is this single one 3.30 ?

PR #7407, opened 2025-08-20
Update fr.po

goofy-mdn:
mainly typofix


hatstand:
Translations are handled on Transifex: https://explore.transifex.com/davidsansome/clementine/

Issue #7370, opened 2024-12-27
Filtering tracks in library by length should handle abbreviations like lastplayed


residuum:
### Before posting

Please follow the steps below and check the boxes with [x] once you did the step.

- [x] I checked the issue tracker for similar issues
- [x] I checked the [changelog](https://github.com/clementine-player/Clementine/blob/master/Changelog) if the issue is already resolved
- [ ] I tried the latest Clementine build from [here](https://builds.clementine-player.org/)

### System information

Please provide information about your system and the version of Clementine used.

- Operating System: Debian testing/unstable
- Clementine version: 1.4 rc2

### Expected behaviour / actual behaviour

When filtering tracks in the library, it is possible to search for tracks played within a specified time with the syntax `lastplayed:<1h30m`. It is also possible to filter tracks by length, but not with the abbreviations for minutes, hours, seconds etc. So one must search for nanoseconds now, e.g. `length:<60000000000` returns tracks shorter than 1 minute in the library, but `length:<1m` would be nice for that.

PR #7413, opened 2025-09-07
Filter tracks by length with tokens similar to lastplayed etc.

residuum:
Implements #7370

### DIFF
--- a/src/library/libraryquery.cpp
+++ b/src/library/libraryquery.cpp
@@ -22,6 +22,7 @@
 #include <QtDebug>
 
 #include "core/song.h"
+#include "core/timeconstants.h"
 
 QueryOptions::QueryOptions() : max_age_(-1), query_mode_(QueryMode_All) {}
 
@@ -51,9 +52,7 @@ const QMap<QString, Song::FileType> kFiletypeId = QMap<QString, Song::FileType>(
                                       {"stream", Song::Type_Stream},
                                       {"unknown", Song::Type_Unknown}});
 
-const int kNsecsPerSec = 1e9;
 const QString kAllowedChars = "smhd";
-const int kSecsPerDay = 60 * 60 * 24;
 
 LibraryQuery::LibraryQuery(const QueryOptions& options)
     : include_unavailable_(false), join_with_fts_(false), limit_(-1) {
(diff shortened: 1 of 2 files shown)


src/library/libraryquery.cpp
```diff
@@ -105,32 +105,16 @@ LibraryQuery::LibraryQuery(const QueryOptions& options)
             if (ok) {
               AddWhere(columntoken, doubleVal, op);
             }
+          } else if (columntoken == "length") {
+            double seconds = GetSecondsFromToken(val);
+            if (seconds > 0) {
+              double nanoseconds = seconds * 1000000000;
```
hatstand, line 111: `seconds * kNsecsPerSec`

src/library/libraryquery.cpp
```diff
@@ -181,6 +165,33 @@ LibraryQuery::LibraryQuery(const QueryOptions& options)
   }
 }
 
+int LibraryQuery::GetSecondsFromToken(QString val) {
+  int seconds = 0;
+  QString tmp = "";
+  QString allowedChars = "smhd";
```
hatstand, line 171: `static const QString kAllowedChars`

src/library/libraryquery.cpp
```diff
@@ -181,6 +165,33 @@ LibraryQuery::LibraryQuery(const QueryOptions& options)
...
+  for (QChar c : val) {
+    if (c.isDigit()) {
+      tmp.append(c);
+    } else if (allowedChars.contains(c)) {
+      bool ok;
```
hatstand, line 176: `bool ok = false;`

src/library/libraryquery.cpp
```diff
@@ -181,6 +165,33 @@ LibraryQuery::LibraryQuery(const QueryOptions& options)
   }
 }
 
+int LibraryQuery::GetSecondsFromToken(QString val) {
```
hatstand, line 168: `int LibraryQuery::GetSecondsFromToken(const QString& val) const`

src/library/libraryquery.cpp
```diff
@@ -181,6 +165,33 @@ LibraryQuery::LibraryQuery(const QueryOptions& options)
...
+      tmp = "";
+      if (ok) {
+        if (c == 's') {
+          seconds += intVal;
+        } else if (c == 'm') {
+          seconds += intVal * 60;
+        } else if (c == 'h') {
+          seconds += intVal * 60 * 60;
+        } else if (c == 'd') {
+          seconds += intVal * 60 * 60 * 24;
```
hatstand, line 187: `kSecsPerDay`

hatstand (commented):
Just some style fixes while you're there.

src/library/libraryquery.cpp
residuum: I am not well versed in C++, so I am unsure about the semantics, and if this is also needed in the corresponding .h file, but I have added it there as well.

PR #7431, opened 2026-01-09
Bump CMake minimum to 3.16

hatstand:
(Focal comes with 3.16.3)


yselkowitz:
Works with CMake 4.2 on Fedora.

